PR #6, opened 2021-07-22
Reformat code using google-java-format

pradh:
This PR includes no logic change.


shifucun (approved):
Nice! Is there a way to add this check to cloud build as presubmit check?

pradh:
Thanks.

> Nice! Is there a way to add this check to cloud build as presubmit check?

Yes, I have a TODO in the cloudbuild for exactly that...

PR #7, opened 2021-07-22
Add protoc generation and formatting to pre-submit

pradh:
Rely on maven to generate protoc and format code by using different plugins.  Also upgrade Java from 8 to 11.


shifucun (approved):

pradh:
Thanks for the review!

PR #8, opened 2021-07-23
Fix protoc generation and simplify a few maven things

pradh:
* The previous change had a bug in protoc generation in that it was not using the right variables and for the variables to take effect, we should use the "generate-sources" or "compile" target, rather than "com.github.os72:protoc-jar-maven-plugin:run".
* Include code formatting as part of the build and format checking as part of test.
* Remove unnecessary shell scripts


shifucun (approved):

pradh:
Thank you!

PR #9, opened 2021-07-23
Add a new Log message proto

pradh:
This should technically be a "diff base" on https://github.com/datacommonsorg/import/pull/8. The main thing to review here is Debug.proto.

It is meant as a common message for use in lint and eventually in the pipelines.


util/src/main/proto/Debug.proto
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    // One log entry. This could be a sample of messages for a particular
+    // counter.
+    message Entry {
+        optional Level level = 1;
+        optional Location location = 2;
+        // This must be user understandable.
+        optional string message = 3;
+        // A counter in CounterSet.
```
beets, line 56: nit: A counter key for CounterSet? (and likewise, field name could be counter_key since counter alone is ambiguous)

util/src/main/proto/Debug.proto
```diff
@@ -0,0 +1,62 @@
...
+      map<string, int64> counters = 1;
+    }
+
+    // One log entry. This could be a sample of messages for a particular
+    // counter.
+    message Entry {
+        optional Level level = 1;
+        optional Location location = 2;
+        // This must be user understandable.
+        optional string message = 3;
```
beets, line 55: maybe rename to user_message? there might be an additional technical / debug message we would want to add (like stack traces, etc)

beets (commented):
only reviewed debug.proto

util/src/main/proto/Debug.proto
pradh: Done.

util/src/main/proto/Debug.proto
pradh: Done.

pradh (commented):
Thanks for the review!

util/src/main/proto/Mcf.proto
```diff
@@ -31,9 +32,14 @@ message McfGraph {
...
 
-  // Map of a property name to its values.
+  // Information about a node.
   message PropertyValues {
+    // Map of a property name to its values.
     map<string, Values> pvs = 1;
+
+    // Information identifying the location of this node in the source.
+    // There can be multiple if PVs in this node are merged from different files.
+    repeated Log.Location location = 2;
```
shifucun, line 42: should Location be outside Log proto? It seems like a general thing.

util/src/main/proto/Mcf.proto
```diff
@@ -15,6 +15,7 @@
 syntax = "proto2";
 
 package org.datacommons.proto;
+import "Debug.proto";
```
beets, line 18: just noticed this - does the [class naming convention](https://google.github.io/styleguide/javaguide.html#s5.2.1-package-names) apply to protos as well? this feels very unnatural, and i think we should stick with [proto package convention](https://developers.google.com/protocol-buffers/docs/style#packages)

util/src/main/java/org/datacommons/proto/Debug.java
```diff
@@ -0,0 +1,4416 @@
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
```
beets, line 1: Should generated files get checked in as well?

beets (approved):
thanks for making the changes!

Is there a linter that you're using at the moment? I found a couple that could work:
- https://github.com/checkstyle/checkstyle
- https://github.com/diffplug/spotless/tree/main/plugin-maven

pradh:
Sorry, I merged by mistake... let me revert.

util/src/main/java/org/datacommons/proto/Debug.java
pradh: Its a TODO there to remove them soon, but they are checked-in now until the protoc workflow is figured out right... so that presubmit works and IDE doesn't show errors until protoc is run (Mixer I believe checks them in too).

pradh:
> Is there a linter that you're using at the moment? I found a couple that could work:
> 
> * https://github.com/checkstyle/checkstyle
> * https://github.com/diffplug/spotless/tree/main/plugin-maven

Yes, it is integrated with google-java-format (https://github.com/datacommonsorg/import#coding-guidelines).

util/src/main/proto/Mcf.proto
pradh: Good catch! I think it should be small case. I will change it in follow-on PR...

util/src/main/proto/Debug.proto
```diff
@@ -0,0 +1,62 @@
...
+    enum Level {
+        LEVEL_UNSPECIFIED = 0;
+        LEVEL_INFO = 1;
+        LEVEL_WARNING = 2;
+        LEVEL_ERROR = 3;
+        LEVEL_FATAL = 4;
+    }
+
+    message CounterSet {
+      map<string, int64> counters = 1;
```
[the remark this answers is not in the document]
pradh: Commented that it is a counter name.

util/src/main/proto/Mcf.proto
pradh: Location sounded like too general a name :). At the moment its only planned for use in the context of logging, debugging, msgs... do you have an alternate name suggestion?

PR #10, opened 2021-07-30
Add utility to parse TMCF + CSV into node MCF

pradh:
This is a port of the flume-based implementation found internally here: https://source.corp.google.com/piper///depot/google3/datacommons/import/util/ext_table_util.cc;l=48-400;bpv=0;bpt=1


util/src/main/java/org/datacommons/util/McfUtil.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.List;
+import java.util.Map;
+import org.datacommons.proto.Mcf;
+
+// A container class of MCF related utilities.
+public class McfUtil {
+  // Serializes an MCFGraph proto into a string.  Sorts output if |sort| is true.
+  public static String serializeMcfGraph(Mcf.McfGraph graph, boolean sort) {
+    StringBuilder result = new StringBuilder();
+    String sentinel = graph.getType() == Mcf.McfType.TEMPLATE_MCF ? "Template" : "Node";
```
shifucun, line 28: maybe can use name() function on enum

shifucun (approved):

pradh:
Thanks for the review!

PR #11, opened 2021-08-03
Add a skeleton of a command line tool in a "tool" maven project

pradh:
Looking for feedback on the CLI.  The current help message (which gets auto-generated via picocli lib) is below.

Changes:

* For command line args, uses https://picocli.info/ 
* Plumbs in MCF parser and TMCF/CSV parser
* Sets up packaging so that a single jar with all deps get generated
* Validated some sample MCF/TMCF/CSV files.
* Includes simplified interface for McfParser

Updated help. The jar is available for trial [here](https://github.com/pradh/import/raw/lint_init/bin/datacommons-import-tool-1.0-SNAPSHOT-jar-with-dependencies.jar).

1. Main command
```
$ dc-import 
Missing required subcommand
Usage: dc-import [-hV] [-o=<outputDir>] [COMMAND]
Tool for use in developing datasets for Data Commons.
  -h, --help      Show this help message and exit.
  -o, --outputDir=<outputDir>
                  Directory to write output files. Default is current working
                    directory.
  -V, --version   Print version information and exit.
Commands:
  lint    Run various checks on input MCF/TMCF/CSV files
  genmcf  Generate Instance MCF from TMCF/CSV files
```

2. lint subcommand
```
$ dc-import lint
Missing required parameter: '<files>'
Usage: dc-import lint [-d=<delimiter>] [-o=<outputDir>] <files>...
Run various checks on input MCF/TMCF/CSV files
      <files>...   List of input files. The file extensions are used to infer
                     the format. Valid extensions include .mcf for Instance
                     MCF, .tmcf for Template MCF, .csv for tabular text files
                     delimited by comma (overridden with -d), and .tsv for
                     tab-delimited tabular files.
  -d, --delimiter=<delimiter>
                   Delimiter of the input CSV files. Default is ',' for .csv
                     files and '\t' for .tsv files.
  -o, --outputDir=<outputDir>
                   Directory to write output files. Default is current working
                     directory.
```

3. genmcf subcommand
```
$ dc-import genmcf
Missing required parameter: '<files>'
Usage: dc-import genmcf [-d=<delimiter>] [-o=<outputDir>] <files>...
Generate Instance MCF from TMCF/CSV files
      <files>...   List of input files. The file extensions are used to infer
                     the format. Valid extensions include .tmcf for Template
                     MCF, .csv for tabular text files delimited by comma
                     (overridden with -d), and .tsv for tab-delimited tabular
                     files. Note that .mcf is not a valid input.
  -d, --delimiter=<delimiter>
                   Delimiter of the input CSV files. Default is ',' for .csv
                     files and '\t' for .tsv files.
  -o, --outputDir=<outputDir>
                   Directory to write output files. Default is current working
                     directory.
```


lint/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+        <groupId>org.datacommons</groupId>
+        <artifactId>datacommons-import</artifactId>
+        <version>1.0-SNAPSHOT</version>
+        <relativePath>..</relativePath>
+    </parent>
+
+    <groupId>org.datacommons</groupId>
+    <artifactId>datacommons-import-lint</artifactId>
+    <name>Data Commons - Import Lint Tool</name>
+    <url>http://www.datacommons.org</url>
```
shifucun, line 16: very nit: could just be https://datacommons.org (www will be redirected)

shifucun:
"genMcf" does not seem to be a function for "linter"? Should that be a separate binary, then command type only has "check"?

lint/src/main/java/org/datacommons/lint/Lint.java
```diff
@@ -0,0 +1,162 @@
...
+      description =
+          "The command to run: ${COMPLETION-CANDIDATES}.  'chk' runs "
+              + "syntax checks on the input files.  'genMcf' produces instance MCF output, "
+              + "and only makes sense when --format is 'tmcfCsv'. Default: 'chk'",
+      defaultValue = "chk",
+      required = true)
+  private CommandType cmdType;
+
+  @CommandLine.Option(
+      names = {"-d", "--delimiter"},
```
shifucun, line 65: also from a "linter" point of view, should the tool determine the delimiter rather than let user specify it?

shifucun:
Added some comments, mostly try to see if we can have non-dependent cmd options and let the tool to deduce as much as possible (any way it's a linter :) )

shifucun:
Looking at https://eslint.org/docs/user-guide/command-line-interface, user does not need to say if the input is javascript, typescript or react script. Similarly I wonder if the tool can tell apart the input files and determine the format. So user mostly will just give a bunch of files and that should be enough.

beets:
just strictly from usability / cmd line pov:
- would be nice if we had subcommands, each with different arguments
- maybe we can deduce format from file extension
- would be nice not to have order requirements for files
- it does look like it will do more than lint, we can rename to dc :) or dcimport - so dcimport lint, dcimport resolve, dcimport genmcf

lint/src/main/java/org/datacommons/lint/Lint.java
```diff
@@ -0,0 +1,162 @@
...
+        throw new CommandLine.ParameterException(
+            spec.commandLine(), "Command 'genMcf' is only applicable with --format 'tmcfCsv'");
+      }
+      return HandleNodeFormat();
+    }
+  }
+
+  private Integer HandleNodeFormat() throws IOException {
+    Mcf.McfType mcf_type =
+        (formatType == FormatType.mcf ? Mcf.McfType.INSTANCE_MCF : Mcf.McfType.TEMPLATE_MCF);
```
beets, line 113: if this is inferred from filetype, we can do a simple check to see if there are any column references in the tmcf files first.

lint/src/main/java/org/datacommons/lint/Lint.java
```diff
@@ -0,0 +1,162 @@
...
+    mcf,
+    tmcf,
+    tmcfCsv
+  };
+
+  @CommandLine.Option(
+      names = {"-f", "--format"},
+      description = "Format of input files: ${COMPLETION-CANDIDATES}",
+      required = true)
+  private FormatType formatType;
```
beets, line 47: would the library take care of nice-to-haves like case-insensitive matching?
another concern with this enum is they are all very close to each other in spelling, so it's easy to make a mistake with these format options.

beets (commented):
thanks for sharing this for feedback on the cmd args!

lint/pom.xml
pradh: Done.

lint/src/main/java/org/datacommons/lint/Lint.java
```diff
@@ -0,0 +1,162 @@
...
+
+  @CommandLine.Option(
+      names = {"-o", "--outputDir"},
+      description = "Directory to write output files. Default: /tmp",
+      defaultValue = "/tmp",
+      scope = CommandLine.ScopeType.INHERIT)
+  private File outputDir;
+
+  @CommandLine.Option(
+      names = {"-r", "--isResolved"},
```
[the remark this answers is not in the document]
pradh: I've simplified it for now to be unresolved.

But it is useful to know what the user intended to do, provide resolved MCF or unresolved MCF, so that we can check accordingly. For instance, if they intended to provide resolved MCF and we find a local reference that's an error. But if we find the same with unresolved MCF, that's fine.

lint/src/main/java/org/datacommons/lint/Lint.java
pradh: Changed to use better default for .csv vs. .tsv. We can parse/interpret it, maybe in follow-on. I do wonder about  something like though:

`This,Should|Be,What|Delimiter`

Or, we should at least restrict characters that can go into the header, or something.  Or, provide feedback to user that this is how we interpreted the delimiter.  I just feel this one option is simple enough to provide for starters?

pradh:
> Looking at https://eslint.org/docs/user-guide/command-line-interface, user does not need to say if the input is javascript, typescript or react script. Similarly I wonder if the tool can tell apart the input files and determine the format. So user mostly will just give a bunch of files and that should be enough.

Very good point!  I've tried to simplify that part.

lint/src/main/java/org/datacommons/lint/Lint.java
```diff
@@ -0,0 +1,162 @@
...
+      defaultValue = ",",
+      description =
+          "Delimiter of the input CSV file. Relevant only when --format is 'tmcfCsv'. "
+              + "Default: ',' (comma)")
+  private char delimiter;
+
+  @CommandLine.Option(
+      names = {"-o", "--outputDir"},
+      description = "Directory to write output files. Default: /tmp",
+      defaultValue = "/tmp",
```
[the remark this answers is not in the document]
pradh: I read in some CLI best-practice (can't find ref now) that using cwd could be non-intuitive, esp if people run commands from anywhere. But that could be a general thing, and for here, perhaps people will think about where they run this from. So, yeah, changed to ".".

lint/src/main/java/org/datacommons/lint/Lint.java
pradh: This is obsolete now.

lint/src/main/java/org/datacommons/lint/Lint.java
pradh: This is also obsolete now.

pradh:
Thanks for the feedback!  I think I've incorporated most.  PTAL.

tool/src/main/java/org/datacommons/tool/Tool.java
```diff
@@ -0,0 +1,231 @@
...
+// TODO: Add e2e tests once Debug.Log is fully plumbed in.
+@CommandLine.Command(
+    name = "dc-import",
+    mixinStandardHelpOptions = true,
+    version = "dc-import 0.1",
+    description = "Tool for use in developing datasets for Data Commons.",
+    subcommands = {Lint.class, GenMcf.class})
+class Tool {
+  @CommandLine.Option(
+      names = {"-o", "--outputDir"},
```
shifucun, line 31: nit: --output-dir? dash is used in most popular tools

tool/src/main/java/org/datacommons/tool/Tool.java
```diff
@@ -0,0 +1,231 @@
...
+      logger.info(
+          "Checked CSV {} ({} rows, {} nodes)",
+          csvFile.getName(),
+          numRowsProcessed,
+          numNodesProcessed);
+    }
+  }
+}
+
+@CommandLine.Command(name = "lint", description = "Run various checks on input MCF/TMCF/CSV files")
```
shifucun, line 84: Dose lint tool output files? The command help message indicate but it's not relevant?

tool/src/main/java/org/datacommons/tool/Tool.java
```diff
@@ -0,0 +1,231 @@
...
+      names = {"-o", "--outputDir"},
+      description = "Directory to write output files. Default is current working directory.",
+      scope = CommandLine.ScopeType.INHERIT)
+  public File outputDir;
+
+  public static void main(String... args) {
+    System.exit(new CommandLine(new Tool()).execute(args));
+  }
+
+  public static void HandleNodeFormat(
```
beets, line 40: can we rename Tool to something that more clearly denotes that this is where `main` is? Main or Runner?

handleNodeFormat and handleTabFormat should also go in a separate class so that Lint and GenMcf can go in separate files. We should try to maintain one class per file. should we also give them more specific names? parseNodeFormat or parseTableFormat?

tool/src/main/java/org/datacommons/tool/Tool.java
pradh: Upcoming PRs will change it to output a report.

tool/src/main/java/org/datacommons/tool/Tool.java
pradh: Done.

lint/src/main/java/org/datacommons/lint/Lint.java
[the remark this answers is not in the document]
shifucun: I see that makes sense. "resolved" might be a concept that user gets confused (and they may give wrong option), hopefully we will have a way to hide this in the future..

lint/src/main/java/org/datacommons/lint/Lint.java
shifucun: I see, a default delimiter helps.

shifucun (approved):
Thanks for the update. Nice to see the tool in shape!

util/src/main/java/org/datacommons/util/McfParser.java
```diff
@@ -36,21 +35,21 @@
...
+      throws IOException {
+    McfParser parser = init(type, isResolved);
+    parser.lines =
+        Files.lines(FileSystems.getDefault().getPath(fileName), StandardCharsets.UTF_8).iterator();
     return parser;
   }
 
   // Parse a string with instance nodes in MCF format into the McfGraph proto.
-  public static McfGraph parseInstanceMcfString(String mcf_string, boolean isResolved) {
+  public static McfGraph parseInstanceMcfString(String mcf_string, boolean isResolved)
```
beets, line 51: javastyle :) mcfString or mcfStr
```suggestion
  public static McfGraph parseInstanceMcfString(String mcfString, boolean isResolved)
```

util/src/main/java/org/datacommons/util/McfParser.java
```diff
@@ -61,7 +60,7 @@ public static McfGraph parseInstanceMcfFile(String file_name, boolean isResolved
   }
 
   // Parse a string with template nodes in MCF format into the McfGraph proto.
-  public static McfGraph parseTemplateMcfString(String mcf_string) {
+  public static McfGraph parseTemplateMcfString(String mcf_string) throws IOException {
```
beets, line 63: ditto

beets (approved):
thanks for splitting the commands!

tool/src/main/java/org/datacommons/tool/Tool.java
pradh: 1. Renamed Tool to Main
2. Split the two subcommands into their own file
3. Moved the common helpers into a Processor container class.

RE: giving helper funcs specific names, leaving it as a somewhat general name (processFoo), because I was thinking that these functions will drive the work (like syntax check, stat check, resolution, etc.) for every node/sub-graph parsed in a streaming way (so people can pass in a very large input and we can report partial/incremental results).

pradh: And, wow, IDE makes refactoring a breeze :)

util/src/main/java/org/datacommons/util/McfParser.java
pradh: Done

util/src/main/java/org/datacommons/util/McfParser.java
pradh: Done

pradh:
Thanks for the reviews!

tool/src/main/java/org/datacommons/tool/Tool.java
pradh: I will merge this for ease of follow on changes, and incorporate comments on the above changes in a follow-on PR.

PR #12, opened 2021-08-05
Implement a simple log wrapper, plumb it into the parsers, and produce a JSON report

pradh:
* Add a LogWrapper class to provide helpers for logging to the Debug.Log proto on a file
* Plumb the LogWrapper into MCFParser and TmcfCsvParser classes, replacing every `assert` with a logging call
* Finally, produce a JSON report that can then be viewed in a tabular way using something like jsongrid.com. In future, we can see about producing a self-contained html, maybe.

Example run on a real files:

```
$ dc-import lint Google_Health_COVID19_US*
19:57:06.397 [main] INFO  org.datacommons.tool.Lint - Input includes 1 MCF file(s), 1 TMCF file(s), 1 CSV file(s)
19:57:06.504 [main] INFO  org.datacommons.tool.Processor - Checked Google_Health_COVID19_US_StatisticalVariables.mcf with 12 nodes
19:57:30.730 [main] INFO  org.datacommons.tool.Processor - Processed 100000 rows of Google_Health_COVID19_US.csv.  0 fatal, 0 error(s), 867696 warning(s).
19:57:50.925 [main] INFO  org.datacommons.tool.Processor - Processed 200000 rows of Google_Health_COVID19_US.csv.  0 fatal, 0 error(s), 1765154 warning(s).
19:58:11.079 [main] INFO  org.datacommons.tool.Processor - Processed 300000 rows of Google_Health_COVID19_US.csv.  0 fatal, 0 error(s), 2663454 warning(s).
19:58:31.211 [main] INFO  org.datacommons.tool.Processor - Processed 400000 rows of Google_Health_COVID19_US.csv.  0 fatal, 0 error(s), 3561430 warning(s).
19:58:51.391 [main] INFO  org.datacommons.tool.Processor - Processed 500000 rows of Google_Health_COVID19_US.csv.  0 fatal, 0 error(s), 4459560 warning(s).
19:58:59.669 [main] INFO  org.datacommons.tool.Processor - Checked CSV Google_Health_COVID19_US.csv (540971 rows, 5950681 nodes)
19:58:59.669 [main] INFO  org.datacommons.util.McfParser - Failures: 0 fatal, 0 error(s), 4827975 warning(s).  Writing details to ./report.json
```

The report.json when viewed on jsongrid.com looks like this: https://screenshot.googleplex.com/8JHZBZxndJ3jrDT

### DIFF
--- a/tool/src/main/java/org/datacommons/tool/GenMcf.java
+++ b/tool/src/main/java/org/datacommons/tool/GenMcf.java
@@ -83,9 +83,10 @@ class GenMcf implements Callable<Integer> {
     BufferedWriter writer = new BufferedWriter(new FileWriter(outPath.toString()));
 
     LogWrapper logCtx = new LogWrapper(Debug.Log.newBuilder(), parent.outputDir.toPath());
+    Processor processor = new Processor(logCtx);
     Integer retVal = 0;
     try {
-      Processor.processTables(tmcfFiles.get(0), csvFiles, delimiter, writer, logCtx);
+      processor.processTables(tmcfFiles.get(0), csvFiles, delimiter, writer);
     } catch (DCTooManyFailuresException ex) {
       // Regardless of the failures, we will dump the logCtx and exit.
       retVal = -1;
--- a/tool/src/main/java/org/datacommons/tool/GenMcf.java
+++ b/tool/src/main/java/org/datacommons/tool/GenMcf.java
@@ -79,7 +79,7 @@ class GenMcf implements Callable<Integer> {
     }
 
     Path outPath = Paths.get(parent.outputDir.getPath(), "generated.mcf");
-    logger.info("Writing to {}", outPath.toAbsolutePath().normalize().toString());
+    logger.info("Writing generated MCF to {}", outPath.toAbsolutePath().normalize().toString());
     BufferedWriter writer = new BufferedWriter(new FileWriter(outPath.toString()));
 
     LogWrapper logCtx = new LogWrapper(Debug.Log.newBuilder(), parent.outputDir.toPath());
@@ -91,7 +91,7 @@ class GenMcf implements Callable<Integer> {
       retVal = -1;
     }
     writer.close();
-    logCtx.writeLog(false);
+    logCtx.persistLog(false);
     return retVal;
   }
 }
--- a/tool/src/main/java/org/datacommons/tool/GenMcf.java
+++ b/tool/src/main/java/org/datacommons/tool/GenMcf.java
@@ -81,7 +81,7 @@ class GenMcf implements Callable<Integer> {
     String directory = parent.outputDir == null ? "." : parent.outputDir.getPath();
     Debug.Log.Builder logCtx = Debug.Log.newBuilder();
     Path outPath = Paths.get(directory, "generated.mcf");
-    logger.info("Writing to {}", outPath.toString());
+    logger.info("Writing to {}", outPath.toAbsolutePath().toString());
     BufferedWriter writer = new BufferedWriter(new FileWriter(outPath.toString()));
     Integer retVal = 0;
     try {
--- a/tool/src/main/java/org/datacommons/tool/GenMcf.java
+++ b/tool/src/main/java/org/datacommons/tool/GenMcf.java
@@ -85,7 +85,7 @@ class GenMcf implements Callable<Integer> {
     Integer retVal = 0;
     try {
       Processor.processTables(tmcfFiles.get(0), csvFiles, delimiter, writer, logCtx);
-    } catch (TooManyFailuresException ex) {
+    } catch (DCTooManyFailuresException ex) {
       // Regardless of the failures, we will dump the logCtx and exit.
       retVal = -1;
     }
--- a/tool/src/main/java/org/datacommons/tool/GenMcf.java
+++ b/tool/src/main/java/org/datacommons/tool/GenMcf.java
@@ -78,11 +78,11 @@ class GenMcf implements Callable<Integer> {
       delimiter = nTsv > 0 ? '\t' : ',';
     }
 
-    String directory = parent.outputDir == null ? "." : parent.outputDir.getPath();
-    Debug.Log.Builder logCtx = Debug.Log.newBuilder();
-    Path outPath = Paths.get(directory, "generated.mcf");
-    logger.info("Writing to {}", outPath.toAbsolutePath().toString());
+    Path outPath = Paths.get(parent.outputDir.getPath(), "generated.mcf");
+    logger.info("Writing to {}", outPath.toAbsolutePath().normalize().toString());
     BufferedWriter writer = new BufferedWriter(new FileWriter(outPath.toString()));
+
+    LogWrapper logCtx = new LogWrapper(Debug.Log.newBuilder(), parent.outputDir.toPath());
     Integer retVal = 0;
     try {
       Processor.processTables(tmcfFiles.get(0), csvFiles, delimiter, writer, logCtx);
@@ -91,8 +91,7 @@ class GenMcf implements Callable<Integer> {
       retVal = -1;
     }
     writer.close();
-
-    LogWrapper.writeLog(logCtx, Paths.get(directory, "report.json"));
+    logCtx.writeLog(false);
     return retVal;
   }
 }
--- a/tool/src/main/java/org/datacommons/tool/GenMcf.java
+++ b/tool/src/main/java/org/datacommons/tool/GenMcf.java
@@ -1,5 +1,6 @@
 package org.datacommons.tool;
 
+import com.google.protobuf.InvalidProtocolBufferException;
 import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileWriter;
@@ -40,7 +41,7 @@ class GenMcf implements Callable<Integer> {
   @CommandLine.Spec CommandLine.Model.CommandSpec spec; // injected by picocli
 
   @Override
-  public Integer call() throws IOException {
+  public Integer call() throws IOException, InvalidProtocolBufferException {
     List<File> tmcfFiles = new ArrayList<>();
     List<File> csvFiles = new ArrayList<>();
     int nTsv = 0;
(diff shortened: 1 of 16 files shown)


tool/src/main/java/org/datacommons/tool/GenMcf.java
```diff
@@ -75,13 +77,22 @@ public Integer call() throws IOException {
...
+
+    String directory = parent.outputDir == null ? "." : parent.outputDir.getPath();
     Debug.Log.Builder logCtx = Debug.Log.newBuilder();
-    Path outFile =
-        Paths.get(parent.outputDir == null ? "." : parent.outputDir.getPath(), "generated.mcf");
-    logger.info("Writing to {}", outFile.toString());
-    BufferedWriter writer = new BufferedWriter(new FileWriter(outFile.toString()));
-    Processor.processTables(tmcfFiles.get(0), csvFiles, delimiter, writer, logCtx, logger);
+    Path outPath = Paths.get(directory, "generated.mcf");
+    logger.info("Writing to {}", outPath.toString());
```
shifucun, line 84: maybe log absolute path here?

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -4,23 +4,40 @@
...
-    int numNodesProcessed = 0;
+  private static final Logger logger = LogManager.getLogger(Processor.class);
+
+  public static void processNodes(Mcf.McfType type, File file, Debug.Log.Builder logCtx)
+      throws IOException, DCTooManyFailuresException {
+    long numNodesProcessed = 0;
     logger.debug("Checking {}", file.getName());
     // TODO: isResolved is more allowing, be stricter.
-    McfParser parser = McfParser.init(type, file.getPath(), false);
+    McfParser parser = McfParser.init(type, file.getPath(), false, logCtx);
```
shifucun, line 32: I wonder should always put ctx as the first argument (Golang does this..)

tool/src/main/java/org/datacommons/tool/Lint.java
```diff
@@ -74,16 +79,25 @@ public Integer call() throws IOException {
...
+          Processor.processNodes(Mcf.McfType.TEMPLATE_MCF, f, logCtx);
+        }
+      }
+    } catch (DCTooManyFailuresException ex) {
+      // Regardless of the failures, we will dump the logCtx and exit.
+      retVal = -1;
     }
-    return 0;
+
+    String directory = parent.outputDir == null ? "." : parent.outputDir.getPath();
```
beets, line 99: Can "." be the default value of outputDir instead of testing for it everywhere?

tool/src/main/java/org/datacommons/tool/Lint.java
```diff
@@ -74,16 +79,25 @@ public Integer call() throws IOException {
...
+    LogWrapper.writeLog(logCtx, Paths.get(directory, "report.json"));
```
beets, line 100: maybe make "report.json" a constant.

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -54,4 +74,12 @@ public static void processTables(
           numNodesProcessed);
     }
   }
+
+  private static void maybePrintStatus(
+      long count, String thing, String fileName, Debug.Log.Builder logCtx) {
+    if (count > 0 && count % 100000 == 0) {
+      logger.info(
+          "Processed {} {} of {}.  {}.", count, thing, fileName, LogWrapper.logSummary(logCtx));
```
beets, line 82: nit: maybe use count/100000 instead (or the accurate equivalent)? i was confused by this in the eia import, and didn't realize those were cumulative counts.

alternatively we can use some ansi magic to reprint over the previous line: https://stackoverflow.com/questions/45068783/java-how-to-print-output-to-previous-line-after-input/45069499

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -167,18 +185,44 @@ public void process(CSVRecord dataRow) {
...
           if (referenceNode.startsWith(Vocabulary.DCID_PREFIX)) {
             String dcid = referenceNode.substring(Vocabulary.DCID_PREFIX.length());
             if (dcid.isEmpty()) {
-              incrementCounter("NumEmptyDCIDValues", logCtx);
+              addLog(
+                  Debug.Log.Level.LEVEL_WARNING,
+                  "CSV_EmptyDcidReferences",
+                  "Empty dcid reference in property "
+                      + currentProp
+                      + " in TMCF entity "
```
beets, line 224:
```suggestion
                      + " from TMCF entity "
```

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -191,24 +235,58 @@ public void process(CSVRecord dataRow) {
...
+          }
+          if (term.type != McfParser.SchemaTerm.Type.COLUMN) {
+            addLog(
+                Debug.Log.Level.LEVEL_ERROR,
+                "CSV_UnexpectedNonColumn",
+                "Unexpected non-column "
+                    + typedValue.getValue()
+                    + " for property "
+                    + currentProp
+                    + " in TMCF entity "
```
beets, line 250:
```suggestion
                    + " from TMCF entity "
```

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -191,24 +235,58 @@ public void process(CSVRecord dataRow) {
...
-          assert columnIndex < dataRow.size() : "Unexpected inconsistent row " + dataRow.toString();
+          McfParser.SchemaTerm term = McfParser.parseSchemaTerm(typedValue.getValue(), errCb);
+          if (term == null) {
+            continue;
+          }
+          if (term.type != McfParser.SchemaTerm.Type.COLUMN) {
+            addLog(
+                Debug.Log.Level.LEVEL_ERROR,
+                "CSV_UnexpectedNonColumn",
+                "Unexpected non-column "
```
beets, line 246: i'm not sure if this message is completely clear (what does non-column mean?). we might want to err on being more verbose with our errors.

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -167,18 +185,44 @@ public void process(CSVRecord dataRow) {
...
+          String referenceNode = toNodeName(typedValue.getValue(), errCb);
           Mcf.McfGraph.TypedValue.Builder newTypedValue = Mcf.McfGraph.TypedValue.newBuilder();
           if (referenceNode.startsWith(Vocabulary.DCID_PREFIX)) {
             String dcid = referenceNode.substring(Vocabulary.DCID_PREFIX.length());
             if (dcid.isEmpty()) {
-              incrementCounter("NumEmptyDCIDValues", logCtx);
+              addLog(
+                  Debug.Log.Level.LEVEL_WARNING,
+                  "CSV_EmptyDcidReferences",
+                  "Empty dcid reference in property "
```
beets, line 222: maybe mention the (required) missing dcid prefix?

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -191,24 +235,58 @@ public void process(CSVRecord dataRow) {
...
+                    + currentProp
+                    + " in TMCF entity "
+                    + templateEntity);
+            continue;
+          }
+          if (!cleanedColumnMap.containsKey(term.value)) {
+            addLog(
+                Debug.Log.Level.LEVEL_ERROR,
+                "CSV_MissingTmcfColumn",
+                "Missing column " + term.value);
```
beets, line 258: clarify that the error is in the tmcf? and will we be able to specify which tmcf file was used?

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -191,24 +235,58 @@ public void process(CSVRecord dataRow) {
...
+            continue;
+          }
+          int columnIndex = cleanedColumnMap.get(term.value);
+          if (columnIndex >= dataRow.size()) {
+            addLog(
+                Debug.Log.Level.LEVEL_WARNING,
+                "CSV_UnexpectedRow",
+                "Unexpected short row '"
```
beets, line 266: clarify what "short row" means

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
pradh: This is an edge case where value (reference) is missing... aka, there was `dcid:`

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
pradh: Done

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
pradh: Tried to reword.

Did not really focus on the error messages in this PR, just trying to plumb the logCtx everywhere... but yeah, we need to do a pass on every call that adds logEntry and ensure it is verbose enough...

pradh: One more idea is, since we have a distinct counter name for these errors, we can put a lot more verbose details in a central doc page, like:

https://docs.datacommons.org/errors#<counter_name>  and link to it.

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
pradh: used "of"

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
pradh: clarified.  for *every* addLog() call, the file name and line name gets added internally.

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
pradh: Done.

pradh (commented):
Thanks for the review!

tool/src/main/java/org/datacommons/tool/Processor.java
beets: these could also be non-static classes, and carry the log context as class state.

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
beets: oh did not get that from the message... maybe "entity dcid:{node} referenced in TMCF entity {entity} was not found"

tool/src/main/java/org/datacommons/tool/Processor.java
beets: yeah, ansi magic might not work with a logger and clobber additional metadata that gets printed out. would be cool on just stdout.

periodic dumping sounds good. i still think that having non-cumulative counts (i.e. showing instead how many lines were processed in the last n seconds) would be less confusing. but that's easily updated.

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
beets: oh yes, that would be cool. but i like the idea of revisiting all the error messages prior to launch

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
pradh: Clarified more. (This is the case where {node} is empty, rather than not found.)

tool/src/main/java/org/datacommons/tool/Processor.java
pradh: yes, good point, I was making the change in follow-on commits... folded it into here.

tool/src/main/java/org/datacommons/tool/Processor.java
pradh: Ah, missed that, changed to non-cumulative.

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
pradh: Agreed. In fact, might be good for another pair of eyes to do a pass on the strings passed into all addLog() calls....

beets (approved):
thanks for all these updates. really shaping up!

pradh:
Thanks for the detailed review!

PR #13, opened 2021-08-11
Complex Value parsing and a few more utilities

pradh:
* Add support for complex value parsing, effectively porting [this g3 API](http://google3/datacommons/util/kb_parser.h?l=85-100&rcl=389774321)
* Set file/line in every generated MCF node in McfParser and TmcfCsvParser (this will be used by downstream checkers)
* Add a couple of utils to McfUtil for use in upcoming checker
* Move some utils from McfParser to McfUtil
* Refactor some common test functions in TestUtil


util/src/main/java/org/datacommons/util/ComplexValueParser.java
```diff
@@ -0,0 +1,335 @@
+package org.datacommons.util;
+
+import static org.datacommons.proto.Mcf.ValueType.*;
+import static org.datacommons.util.McfParser.splitAndStripWithQuoteEscape;
```
beets, line 4: i think we should avoid imports of individual methods - it makes it harder to figure out, while reading, where the definition is situated. since this is in the same package, can this be used with McfParser.splitAndStripWithQuoteEscape(...) ?

util/src/main/java/org/datacommons/util/ComplexValueParser.java
```diff
@@ -0,0 +1,335 @@
...
+              + "node "
+              + mainNodeId,
+          mainNode.getLocationsList());
+      return false;
+    }
+    if (fields.size() != 2 && fields.size() != 3) {
+      logCtx.addEntry(
+          Debug.Log.Level.LEVEL_ERROR,
+          "MCF_MalformedComplexValueParts",
+          "Complex value must have 2 or 3 components but '"
```
beets, line 95: should we also add an example?

util/src/main/java/org/datacommons/util/ComplexValueParser.java
```diff
@@ -0,0 +1,335 @@
...
+        return false;
+      }
+      dcid = unit + fields.get(valueIdx);
+      name = unit + " " + fields.get(valueIdx);
+    } else {
+      assert fields.size() == 3;
+      if (unit.toLowerCase().equals(Vocabulary.LAT_AND_LONG.toLowerCase())) {
+        isLatLng = true;
+        if (!parseLatLng(fields.get(startIdx), fields.get(endIdx))) {
+          return false;
```
beets, line 166: maybe add a note here that the errors are generated in the follow on parseLatLng function?

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -179,8 +180,16 @@ public void process(CSVRecord dataRow) {
...
           nodeBuilder.putPvs(currentProp, values);
         }
-        // TODO: Add node sanity check
-        instanceMcf.putNodes(currentNode, nodeBuilder.build());
+        Debug.Log.Location.Builder loc = nodeBuilder.addLocationsBuilder();
+        loc.setFile(logCtx.getLocationFile());
+        loc.setLineNumber(csvParser.getCurrentLineNumber());
+
+        Mcf.McfGraph.PropertyValues newNode = nodeBuilder.build();
+        /*
```
beets, line 188: remove?

util/src/test/java/org/datacommons/util/ComplexValueParserTest.java
```diff
@@ -0,0 +1,166 @@
...
+    assertTrue(log.getCounterSet().containsCounters("MCF_QuantityMalformedValue"));
+    assertTrue(log.getEntries(0).getUserMessage().contains("must be a number"));
+
+    log = toComplexValueFailure("[dcs:Years - -]");
+    assertTrue(log.getCounterSet().containsCounters("MCF_QuantityRangeMalformedValues"));
+    assertTrue(log.getEntries(0).getUserMessage().contains("must be a number"));
+
+    log = toComplexValueFailure("[a b Years]");
+    assertTrue(log.getCounterSet().containsCounters("MCF_QuantityRangeMalformedValues"));
+    assertTrue(log.getEntries(0).getUserMessage().contains("must be a number"));
```
beets, line 98: maybe also add a test here for "[ years 11 2 ]" (are spaces before/after the square bracket allowed)?

beets (approved):
thanks for adding all the tests as well!

util/src/main/java/org/datacommons/util/ComplexValueParser.java
pradh: I see.  The advanced IDE technology auto-added all the imports, and I wasn't paying attention as long as things compiled.  I've fixed this one and others in tests manually.

util/src/main/java/org/datacommons/util/ComplexValueParser.java
```diff
@@ -0,0 +1,335 @@
...
+    dcid = new String();
+    name = new String();
+  }
+
+  public String getDcid() {
+    return dcid;
+  }
+
+  public boolean parse() {
+    if (!complexValue.startsWith("[") || !complexValue.endsWith("]")) {
```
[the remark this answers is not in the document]
pradh: Yes, this gets called using the value in McfGraph.TypedValue.value, which is already trimmed. I've added a comment.

util/src/main/java/org/datacommons/util/ComplexValueParser.java
pradh: Done.

util/src/main/java/org/datacommons/util/ComplexValueParser.java
pradh: Done.

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
pradh: Done

util/src/test/java/org/datacommons/util/ComplexValueParserTest.java
pradh: Good point. Yes, spaces before/after are okay. I've added this example in the success testcase above.

pradh (commented):
Thanks for the review!

PR #14, opened 2021-08-12
Implement MCF checking and transformations

pradh:
This PR largely implements logic in g3's [mcf_schema_util.cc](https://source.corp.google.com/piper///depot/google3/datacommons/import/util/mcf_schema_util.cc).

With this PR, modulo some minor transformation TODOs, the tool should have parity with cpp code for parsing MCF/TMCF/CSV and checking them.

So, for imports where the place DCIDs are mapped (because no resolution support yet), if the tool reports no errors/warnings, then the import should be structurally sane (or its a tool bug).

### DIFF
--- a/tool/src/main/java/org/datacommons/tool/Processor.java
+++ b/tool/src/main/java/org/datacommons/tool/Processor.java
@@ -55,7 +55,6 @@ public class Processor {
     logger.debug("TMCF " + tmcfFile.getName());
     for (File csvFile : csvFiles) {
       logger.debug("Checking CSV " + csvFile.getPath());
-      logCtx.setLocationFile(csvFile.getName());
       TmcfCsvParser parser =
           TmcfCsvParser.init(tmcfFile.getPath(), csvFile.getPath(), delimiter, logCtx);
       Mcf.McfGraph g;
--- a/tool/src/main/java/org/datacommons/tool/Processor.java
+++ b/tool/src/main/java/org/datacommons/tool/Processor.java
@@ -1,9 +1,5 @@
 package org.datacommons.tool;
 
-import java.io.BufferedWriter;
-import java.io.File;
-import java.io.IOException;
-import java.util.List;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.datacommons.proto.Mcf;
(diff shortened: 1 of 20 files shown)


util/src/main/java/org/datacommons/util/McfChecker.java
```diff
@@ -0,0 +1,498 @@
...
+  private final Set<String> PROP_REFS_IN_PROP =
+      Set.of(Vocabulary.NAME, Vocabulary.LABEL, Vocabulary.DCID, Vocabulary.SUB_PROPERTY_OF);
+  private Mcf.McfGraph graph;
+  private LogWrapper logCtx;
+  private Set<String> columns; // Relevant only when graph.type() == TEMPLATE_MCF
+  boolean foundFailure = false;
+
+  // Argument |graph| may be Instance or Template MCF.
+  public McfChecker(Mcf.McfGraph graph, LogWrapper logCtx) {
+    this.graph = graph;
```
shifucun, line 32: I found this.graph is not really used in the class. I wonder if all the check methods here could be static and just take the node or graph as argument, so the client can just do the check without constructing a checker object?

McfChecker.check(g, logCtx)

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -52,12 +55,17 @@ public void processTables(
...
-      logCtx.setLocationFile(csvFile.getName());
       TmcfCsvParser parser =
           TmcfCsvParser.init(tmcfFile.getPath(), csvFile.getPath(), delimiter, logCtx);
       Mcf.McfGraph g;
       long numNodesProcessed = 0, numRowsProcessed = 0;
       while ((g = parser.parseNextRow()) != null) {
+        McfMutator mutator = new McfMutator(g.toBuilder(), logCtx);
+        g = mutator.apply();
+
+        McfChecker checker = new McfChecker(g, logCtx);
```
shifucun, line 66: echo to the other comment, here feels more nature to do 

McfChecker.check(g, logCtx)

pradh: Done

util/src/main/java/org/datacommons/util/McfChecker.java
pradh: Good idea!

pradh:
Thanks for the review! PTAL.

shifucun (approved):

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -37,6 +34,11 @@ public void processNodes(Mcf.McfType type, File file)
     McfParser parser = McfParser.init(type, file.getPath(), false, logCtx);
     Mcf.McfGraph n;
     while ((n = parser.parseNextNode()) != null) {
+      n = McfMutator.apply(n.toBuilder(), logCtx);
```
shifucun, line 37: should this be McfMutator.mutate()?

pradh: Done.

PR #19, opened 2021-08-13
Bug fix: deduplicate single node content

pradh:
A user error resulted in duplicate typeOf values. And this then causes duplicate error messages, like below, because we were not deduplicating within a single node.

image


shifucun (approved):

pradh:
Thanks!

PR #20, opened 2021-08-13
 Update project, pom and README

shifucun:


tool/.project
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pradh, line 1: To confirm, this is needed for eclipse?  IntelliJ works off of pom.xml

pradh (approved):

tool/.project
shifucun: Removed them

PR #21, opened 2021-08-16
log property name in counter for empty CSV cell values

chejennifer:


util/src/main/java/org/datacommons/util/McfParser.java
```diff
@@ -526,7 +555,8 @@ public static SchemaTerm parseSchemaTerm(String value, BiConsumer<String, String
     } catch (IOException e) {
       if (errCb != null) {
         errCb.accept(
-            "StrSplit_ParserFailure", "Parsing failed on '" + orig + "' (" + arg.context + ")");
```
pradh, line 529: Maybe here too we can add property?

util/src/main/java/org/datacommons/util/McfParser.java
```diff
@@ -504,15 +528,20 @@ public static SchemaTerm parseSchemaTerm(String value, BiConsumer<String, String
...
+          }
+          errCb.accept(
+              "StrSplit_EmptyToken" + prop_suffix, "Empty value found (" + argContextString + ")");
         }
         return splits;
       }
       if (records.size() != 1) {
         if (errCb != null) {
           errCb.accept(
               "StrSplit_MultiToken",
```
pradh, line 543: Maybe here too we can add property?

util/src/main/java/org/datacommons/util/McfParser.java
```diff
@@ -504,15 +528,20 @@ public static SchemaTerm parseSchemaTerm(String value, BiConsumer<String, String
       List<CSVRecord> records = parser.getRecords();
       if (records.isEmpty()) {
         if (errCb != null) {
-          errCb.accept("StrSplit_EmptyToken", "Empty value found (" + arg.context + ")");
+          String prop_suffix = "";
+          if (argContext != null) {
+            prop_suffix = "_" + argContext.prop;
+          }
+          errCb.accept(
+              "StrSplit_EmptyToken" + prop_suffix, "Empty value found (" + argContextString + ")");
```
pradh, line 536: In an existing test, maybe you could add a check like https://github.com/datacommonsorg/import/blob/master/util/src/test/java/org/datacommons/util/TmcfCsvParserTest.java#L46-L47 ?

pradh (commented):
Thanks for the change!

util/src/main/java/org/datacommons/util/McfParser.java
chejennifer: done

util/src/main/java/org/datacommons/util/McfParser.java
chejennifer: done

util/src/main/java/org/datacommons/util/McfParser.java
```diff
@@ -484,12 +484,36 @@ public static SchemaTerm parseSchemaTerm(String value, BiConsumer<String, String
     public char delimiter = ',';
     public boolean includeEmpty = false;
     public boolean stripEnclosingQuotes = true;
-    public String context;
+    public SplitAndStripArgContext context;
   }
+
+  public static final class SplitAndStripArgContext {
```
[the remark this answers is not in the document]
chejennifer: will do the different approach in a separate pr

util/src/main/java/org/datacommons/util/McfParser.java
chejennifer: done

pradh (approved):

PR #22, opened 2021-08-16
Add barebone of server using Spring framework

shifucun:
This is to set up the sever and prepare for adding more APIs.


README.md
```diff
@@ -69,6 +69,20 @@ To build binary: ```mvn package```
 - which will produce  `tool/target/datacommons-import-tool-0.1-alpha.1-jar-with-dependencies.jar`
 - and you can run it with ```java -jar <path-to-datacommons-import-tool-jar-with-dependencies.jar>```
 
+### Run Server
```
pradh, line 72: Maybe indicate that this is an experimental server for private DC?

pradh (approved):

README.md
shifucun: Done

shifucun:
Thanks for review!

PR #23, opened 2021-08-16
Add general stat counters

chejennifer:
- added counters for NumRowSuccesses, NumPVSuccesses, NumNodeSuccesses
- should persistLog even if there are no warnings or errors to show the new counters

### DIFF
--- a/util/src/main/java/org/datacommons/util/LogWrapper.java
+++ b/util/src/main/java/org/datacommons/util/LogWrapper.java
@@ -106,10 +106,6 @@ public class LogWrapper {
   }
 
   public void persistLog(boolean silent) throws InvalidProtocolBufferException, IOException {
-    if (log.getLevelSummaryMap().isEmpty()) {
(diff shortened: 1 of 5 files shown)


pradh (approved):
Thanks!

util/src/main/java/org/datacommons/util/LogWrapper.java
```diff
@@ -106,10 +106,6 @@ public void provideStatus(long count, String thing)
   }
 
   public void persistLog(boolean silent) throws InvalidProtocolBufferException, IOException {
-    if (log.getLevelSummaryMap().isEmpty()) {
-      if (!silent) logger.info("Found no warnings or errors!");
```
pradh, line 110: nit: might be useful to log this message `if (log.getLevelSummaryMap().isEmpty() && !silent)`  ?

pradh: Nm. summaryString() should be good enough.

PR #24, opened 2021-08-17
Update maven config for server package

shifucun:


util/pom.xml
```diff
@@ -73,18 +73,6 @@
 
     <build>
         <plugins>
-            <plugin>
```
shifucun, line 76: removed these since there is no change from parent

pradh: I remember I had some issue with things not working unless I included it, and so I added it...

pradh (approved):
Just to sanity check, if you do mvn package, then run lint on an example MCF/TMCF/CSV does it run okay?

pradh:
Nice!

shifucun:
> Just to sanity check, if you do mvn package, then run lint on an example MCF/TMCF/CSV does it run okay?

Tested locally, and it worked fine.

shifucun:
> Nice!

Thanks for review!

util/pom.xml
pradh: Can we add this back?

With this change it seems the following don't seem to be included when I run `mvn compile`
```
[INFO] --- protoc-jar-maven-plugin:3.11.4:run (default) @ datacommons-import-util ---

[INFO] --- fmt-maven-plugin:2.11:format (default) @ datacommons-import-util ---
```

PR #26, opened 2021-08-18
Bail based on distinct error counters, rather than messages

pradh:


beets (approved):

pradh:
Thanks for the review!

PR #28, opened 2021-08-18
[Maven] Get rid of <pluginManagement>, rely on <plugins> inheritence

pradh:
A given plugin is specified exactly once in parent pom if it has a common configuration, and the child poms inherit it.

Tested:
1. Made an import order change in .java file, and confirmed formatting plugin fixes it.
2. Made a minor proto change, and confirmed that protoc regenerates the .java file
3. Ran the import-tool and confirmed it works fine
4. Ran the server tool, and confirmed it returned `Hello World`


shifucun (approved):
Thanks for the update!

PR #32, opened 2021-08-20
Add End to End tests

chejennifer:
- Added end to end tests that invoke GenMcf and Lint
- Fixed a few small bugs along the way
   -  output log when there is a fatal tmcf failure during genmcf
   -  handle ConcurrentModificationException thrown by McfMutator.mutate
   -  log Mutator_MissingTypeOf counter when typeOf is missing
   -  handle some NullPointerExceptions
   -  Small message fixes

### DIFF
--- a/tool/src/main/java/org/datacommons/tool/Processor.java
+++ b/tool/src/main/java/org/datacommons/tool/Processor.java
@@ -70,6 +70,8 @@ public class Processor {
       logger.debug("Checking CSV " + csvFile.getPath());
       TmcfCsvParser parser =
           TmcfCsvParser.init(tmcfFile.getPath(), csvFile.getPath(), delimiter, logCtx);
+      // If there were too many failures when initializing the parser, parser will be null and we
+      // don't want to continue processing.
       if (logCtx.loggedTooManyFailures()) {
         throw new DCTooManyFailuresException("processTables encountered too many failures");
       }
--- a/tool/src/main/java/org/datacommons/tool/Processor.java
+++ b/tool/src/main/java/org/datacommons/tool/Processor.java
@@ -70,6 +70,9 @@ public class Processor {
       logger.debug("Checking CSV " + csvFile.getPath());
       TmcfCsvParser parser =
           TmcfCsvParser.init(tmcfFile.getPath(), csvFile.getPath(), delimiter, logCtx);
+      if (logCtx.loggedTooManyFailures()) {
+        throw new DCTooManyFailuresException("processTables encountered too many failures");
+      }
       Mcf.McfGraph g;
       long numNodesProcessed = 0, numRowsProcessed = 0;
       while ((g = parser.parseNextRow()) != null) {
--- a/tool/src/test/java/org/datacommons/tool/GenMcfTest.java
+++ b/tool/src/test/java/org/datacommons/tool/GenMcfTest.java
@@ -16,14 +16,11 @@ package org.datacommons.tool;
 
 import static org.junit.Assert.assertTrue;
 
-import java.io.File;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Iterator;
-import org.apache.commons.io.FileUtils;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -39,14 +36,11 @@ public class GenMcfTest {
     String csv = resourceFile("Csv1.csv");
     CommandLine cmd = new CommandLine(app);
     cmd.execute("genmcf", tmcf, csv, "--output-dir=" + testFolder.getRoot().getPath());
-    File actualReportFile =
-        new File(Paths.get(testFolder.getRoot().getPath(), "report.json").toString());
-    File expectedReportFile =
-        new File(this.getClass().getResource("GenMcfTest_FatalTmcfReport.json").getPath());
-    File actualGeneratedMcfFile =
-        new File(Paths.get(testFolder.getRoot().getPath(), "generated.mcf").toString());
-    TestUtil.assertReportFilesAreSimilar(expectedReportFile, actualReportFile);
-    assertTrue(FileUtils.readFileToString(actualGeneratedMcfFile, StandardCharsets.UTF_8).isEmpty());
+    String actualReportString = TestUtil.getStringFromTestFile(testFolder, "report.json");
+    String expectedReportString =
+        TestUtil.getStringFromResource(this.getClass(), "GenMcfTest_FatalTmcfReport.json");
+    TestUtil.assertReportFilesAreSimilar(expectedReportString, actualReportString);
+    assertTrue(TestUtil.getStringFromTestFile(testFolder, "generated.mcf").isEmpty());
   }
 
   @Test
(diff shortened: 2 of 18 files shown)


tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -70,6 +70,9 @@ public void processTables(
       logger.debug("Checking CSV " + csvFile.getPath());
       TmcfCsvParser parser =
           TmcfCsvParser.init(tmcfFile.getPath(), csvFile.getPath(), delimiter, logCtx);
+      if (logCtx.loggedTooManyFailures()) {
```
pradh, line 73: nit: comment why this is needed?

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+public class GenMcfTest {
+
+  @Rule public TemporaryFolder testFolder = new TemporaryFolder();
+
+  @Test
+  public void GenMcfTest_FatalTmcf() throws IOException {
+    Main app = new Main();
+    String tmcf = resourceFile("TmcfWithErrors.tmcf");
+    String csv = resourceFile("Csv1.csv");
```
pradh, line 40: Should put the resource files for this in a per-test directory with input/ and output/?  Then there can be one @Test function that runs through every directory, invokes the command with files in input/ and checks against the files in output/.

This PR is already big, we can consider that in follow-up

tool/src/test/resources/org/datacommons/tool/Csv1.csv
```diff
@@ -0,0 +1,12 @@
+Year,Place_Id,Place_Name,Place_Type,LatLong,SV1,SV2
```
pradh, line 1: Nice test cases!

util/src/main/java/org/datacommons/util/McfMutator.java
```diff
@@ -32,18 +32,18 @@
 
   public static Mcf.McfGraph mutate(Mcf.McfGraph.Builder graph, LogWrapper logCtx) {
     McfMutator m = new McfMutator();
-    m.graph = graph;
+    m.graph = Mcf.McfGraph.newBuilder();
```
pradh, line 35: I don't follow this change, what was the problem?

I thought it slightly better that we set everything there is in graph first, otherwise we might need to copy over all non-nodes fields (right now type https://github.com/datacommonsorg/import/blob/master/util/src/main/proto/Mcf.proto#L53 but it can grow).

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -75,6 +75,7 @@ public static TmcfCsvParser init(
         McfChecker.check(
             tmcfCsvParser.tmcf, tmcfCsvParser.csvParser.getHeaderMap().keySet(), logCtx);
     if (!success) {
+      tmcfCsvParser.logCtx.setLocationFile(tmcfFile);
```
pradh, line 78: nit: add a comment why we set

pradh (approved):
Thanks for the thorough test-cases and the fixes!

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
```diff
@@ -0,0 +1,95 @@
...
+    CommandLine cmd = new CommandLine(app);
+    cmd.execute("genmcf", tmcf, csv, "--output-dir=" + testFolder.getRoot().getPath());
+    File gotReportFile =
+        new File(Paths.get(testFolder.getRoot().getPath(), "report.json").toString());
+    File gotGeneratedMcfFile =
+        new File(Paths.get(testFolder.getRoot().getPath(), "generated.mcf").toString());
+    File wantReportFile =
+        new File(this.getClass().getResource("GenMcfTest_FatalTmcfReport.json").getPath());
+    assertEquals(
+        FileUtils.readFileToString(wantReportFile, StandardCharsets.UTF_8),
```
pradh, line 50: Is it worth wrapping these as TestUtil.getStringFromTestFile("report.json") and TestUtil.getStringFromResource("GenMcfTest_foo.json") ?

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
```diff
@@ -0,0 +1,95 @@
...
+        FileUtils.readFileToString(gotReportFile, StandardCharsets.UTF_8));
+  }
+
+  private String resourceFile(String resource) {
+    return this.getClass().getResource(resource).getPath();
+  }
+
+  private boolean areSimilarGeneratedMcf(Path expectedFilePath, Path actualFilePath)
```
pradh, line 79: Explain why this is needed?

pradh (approved):

util/src/main/java/org/datacommons/util/McfMutator.java
chejennifer: ConcurrentModificationException got thrown when updating the nodesMap while iterating over it. I think this is because the nodesMap is an UnmodifiableMap: https://screenshot.googleplex.com/b24GVJLrkLPNsjp

pradh: Oh ok. Then maybe we can store the node names in a list and loop over those?  Like in http://google3/datacommons/import/util/mcf_schema_util.cc?l=433-438&rcl=377096245

tool/src/main/java/org/datacommons/tool/Processor.java
chejennifer: done

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
chejennifer: sounds good, will do it in a follow-up

tool/src/test/resources/org/datacommons/tool/Csv1.csv
chejennifer: thanks!

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
chejennifer: done

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
chejennifer: done

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
chejennifer: done

util/src/main/java/org/datacommons/util/McfMutator.java
chejennifer: done

pradh (approved):
Thanks for making the changes!  This is looking great!

util/src/main/java/org/datacommons/util/McfMutator.java
```diff
@@ -32,10 +33,12 @@
 
   public static Mcf.McfGraph mutate(Mcf.McfGraph.Builder graph, LogWrapper logCtx) {
     McfMutator m = new McfMutator();
-    m.graph = Mcf.McfGraph.newBuilder();
+    m.graph = graph;
     m.logCtx = logCtx;
-    for (String nodeId : graph.getNodesMap().keySet()) {
-      m.graph.putNodes(nodeId, m.mutateNode(nodeId, graph.getNodesOrThrow(nodeId).toBuilder()));
+    List<String> nodeIdList = new ArrayList<>();
```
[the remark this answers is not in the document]
chejennifer: done

PR #34, opened 2021-08-20
Abstract raw files read as a util function

shifucun:
This is going to be used in server module as well


tool/src/main/java/org/datacommons/tool/Lint.java
```diff
@@ -54,56 +54,23 @@
...
-        mcfFiles.size(),
-        tmcfFiles.size(),
-        csvFiles.size());
-    // Various checks
-    if (nTsv > 0 && nTsv != csvFiles.size()) {
-      throw new CommandLine.ParameterException(
-          spec.commandLine(), "Please do not mix .tsv and .csv files");
-    }
+
+    FileGroup fg = TmcfCsvReader.read(files, spec, logger);
```
pradh, line 58: The class TmcfCsvReader gives the impression that it is actually reading the files, but it is just enumerating them. Moreover, this exists to just build a FileGroup, so how about making it a static function inside FileGroup?  

FileGroup.Build(...)

pradh (commented):
Nice refactor.

tool/src/main/java/org/datacommons/tool/Lint.java
shifucun: Yea, that's better

pradh (approved):
Thanks.

PR #35, opened 2021-08-20
Implement existence checks by calling DC API (needs to be enabled by -e)

pradh:
This checks that references to schema-like things exist in the KG or the local graph (i.e., other provided MCFs).

Notably, for performance reasons, for now it does not validate references to place entities (`observationAbout` value).

Existence check results are cached locally to avoid future calls.  Batching would be nice, but that would be a more significant refactor to the call-sites (e.g., we need to save the per-reference context required for logging), so we can consider in future.

Testing: there's a unit test for ExistenceChecker and another to confirm that individual nodes are existence-checked in McfChecker. Plan to add an e2e once the LintTest is generalized.

TODO: get some real testing and enable `-e` by default...

### DIFF
--- a/util/src/main/java/org/datacommons/util/ExistenceChecker.java
+++ b/util/src/main/java/org/datacommons/util/ExistenceChecker.java
@@ -28,7 +28,7 @@ public class ExistenceChecker {
   private HttpClient httpClient;
 
   // This is a combination of local KG data and prior cached checks.
-  // Node is just the DCID. Triple is "s,p,o" and the property just includes CACHED_PROPERTIES.
+  // Node is just the DCID. Triple is "s,p,o" and the property just includes SCHEMA_PROPERTIES.
   private Set<String> existingNodesOrTriples; // Existence cache
   private Set<String> missingNodesOrTriples; // Absence cache
   private LogWrapper logCtx;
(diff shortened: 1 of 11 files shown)


util/src/main/java/org/datacommons/util/ExistenceChecker.java
```diff
@@ -0,0 +1,158 @@
...
+
+    var dataJson = callDc(sub, pred);
+    logCtx.incrementCounterBy("Existence_NumDcCalls", 1);
+    if (dataJson == null) {
+      // If the DCID is malformed Mixer can return failure.
+      return false;
+    }
+    if (dataJson.entrySet().size() != 1) {
+      throw new IOException(
+          "Invalid payload content from Staging DC API endpoint for: '"
```
beets, line 102:
```suggestion
          "Invalid results payload from Staging DC API endpoint for: '"
```

beets (approved):
thanks, sorry for the late review!

chejennifer (approved):
Sorry for the late review, looks good!

util/src/main/java/org/datacommons/util/ExistenceChecker.java
pradh: Done

PR #38, opened 2021-08-24
Generalize E2E Tests

chejennifer:
- generalized E2E tests so that resources are in a per-test directory with input/ and output/.
- fixed some bugs along the way:
    - Couldn't run lint without having a tmcf file
    - The e2e tests weren't comparing the report files properly


tool/src/test/java/org/datacommons/tool/LintTest.java
```diff
@@ -1,6 +1,10 @@
 package org.datacommons.tool;
 
+import java.io.File;
```
pradh, line 3: Add a short comment at the top on adding a new test-case?

pradh (approved):
This is neat!  So now we can just keep adding new test-cases with input/ and output/ and zero code change!

PR #47, opened 2021-08-27
Deploy config and script for AppEngine and CloudRun

shifucun:
Findings from the experiments:

- Our code structure is not friendly to AppEngine, which expects a specific layout: https://cloud.google.com/appengine/docs/standard/java11/configuring-your-app-with-app-yaml. As we have multiple binary and only deploy one of it to AppEngine, we need to build the jar file and deploy it directly (instead of letting AppEngine to bundle it). The drawback is the local Java version has to match the AppEngine Runtime.

- AppEngine standard supports Java8 and Java11. The maximum resources is from instance class B8 (2G of mem, 4.8G of cpu). The memory is apparently too limited.

- AppEngine Flex supports Java8 only! It can support custom memory and cpu limit but not aligned with our Java choice. This also poses too strong restriction for local environment.

- Cloud Run can support custom memory and cpu as well. The deployment process is fast. However when I read from GCS, there is SSL error. The compute engine service account has editor permission so it's a bit mysterious. When using Cloud Run, need to build docker image.

- Going to test GKE as well in follow up PR. Since we need docker image from Cloud Run anyway, using GKE will not be too much complicated given our experiences.

Overall, AppEngine has too strong restriction of code structure, run time and resources, does not seem a good option. Cloud Run has some unknown issue (could be from lack of experience). 

I borrowed existing project "datcom-mixer-autopush" for deployment (hard coded in the config). When we have a better understanding of the platform, should parameterize the script.


pradh (approved):
Thanks for the detailed notes! (Might be useful to add this to your Private DC doc too)

It feels like if we figure out Cloud Run issues, it might be promising?

I forget if we discussed this, but should we consider decoupling the two projects a bit, the util vs. server, so we are not constrained by directory requirements.  It seems one can host a maven repo privately or even on github (https://gist.github.com/fernandezpablo85/03cf8b0cd2e7d8527063).

shifucun:
> Thanks for the detailed notes! (Might be useful to add this to your Private DC doc too)
> 
> It feels like if we figure out Cloud Run issues, it might be promising?
> 
> I forget if we discussed this, but should we consider decoupling the two projects a bit, the util vs. server, so we are not constrained by directory requirements. It seems one can host a maven repo privately or even on github (https://gist.github.com/fernandezpablo85/03cf8b0cd2e7d8527063).

Yea, separating out the server will indeed help in many aspects. I will get into it once the API work is done. Agree with using cloud run. Need to buffer sometime to dig into the issues.

PR #55, opened 2021-08-27
Enable existence checks by default

pradh:
Ran this against all schema MCF, feels reasonably usable, though I'm sure there are uncovered bugs. At this point it might be more useful than without it.


chejennifer (approved):

pradh:
Thanks!

PR #60, opened 2021-09-03
DCID Generator Library

pradh:
Port the [prod DCID generation library](http://google3/datacommons/util/dcid.h) to Java.

Notably missing is support for "small hashed IDs", used for place types, Source, Curator and Provenance, because [Java doesn't seem to have](https://guava.dev/releases/21.0/api/docs/com/google/common/hash/Hashing.html) farmhash32 that prod uses. If we deviate from prod IDs that's worse than not supporting it.

I don't think this should affect most imports, but we'll see...

### DIFF
--- a/util/src/main/java/org/datacommons/util/DcidGenerator.java
+++ b/util/src/main/java/org/datacommons/util/DcidGenerator.java
@@ -1,0 +1,369 @@
+package org.datacommons.util;
+
+import com.google.common.base.Charsets;
+import com.google.common.hash.Hashing;
+import java.nio.file.Path;
+import java.util.*;
+import org.datacommons.proto.Debug;
+import org.datacommons.proto.Mcf;
+
+// Helper functions to generate dcids for various class types.
+// Unless a node passes checks, its dcid cannot be generated. The functions in this class assume
+// those checks have already happened. So they do not report anything in LogWrapper. If a dcid
+// was not successfully generated, then a prior check will say why that is. This is to avoid
+// double counting failures.
+//
+// TODO: Have a way to assert that node whose dcid cannot be generated has an ERROR entry in
+//  LogWrapper.
+// TODO: Support hashed short-IDs for places. farmhash32 doesn't seem to exist in Java.
+public class DcidGenerator {
+  public static boolean TEST_MODE = false;
+
+  private static final String DC_NAMESPACE = "dc/";
+  private static final String OBS_NAMESPACE = "o/";
+  private static final String SVOBS_NAMESPACE = "o/";
+  private static final String POP_NAMESPACE = "p/";
+
+  // This is an int->char map of 32 chars.
+  private static final char[] DCID_BASE32_MAP = {
+    '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'b', 'c', 'd', 'f', 'g', 'h', 'j', 'k', 'l',
+    'm', 'n', 'p', 'q', 'r', 's', 't', 'v', 'w', 'x', 'y', 'z', 'e'
+  };
+  // Max length of a base 32 encoded version of a 64-bit string: ceil(64 / 5)
+  private static final int MAX_LONG_ID_LEN = 13;
+  // Max length of a base 32 encoded version of a 32-bit string: ceil(32 / 5)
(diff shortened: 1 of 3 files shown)


util/src/main/java/org/datacommons/util/DcidGenerator.java
```diff
@@ -0,0 +1,369 @@
...
+  private static final Set<String> FIXED_STAT_VAR_PROPS =
+      Set.of(
+          Vocabulary.TYPE_OF,
+          Vocabulary.POPULATION_TYPE,
+          Vocabulary.MEASURED_PROP,
+          Vocabulary.STAT_TYPE,
+          Vocabulary.MEASUREMENT_QUALIFIER,
+          Vocabulary.MEASUREMENT_DENOMINATOR);
+
+  static final int LAST_REQUIRED_STAT_VAR_OBS_PROP_INDEX = 3; // until value
```
beets, line 73: i wonder if this should be defined after the list, and then the value derived with code (so that it's more robust to changes). something like below, but it's a bit verbose
```java
LAST_REQUIRED_STAT_VAR_OBS_PROP_INDEX = ORDERED_STAT_VAR_OBS_KEY_PROPS.indexOf(Vocabulary.GENERIC_VALUE)
```

It could also be the comment though? I needed to look twice to understand "until value"

util/src/main/java/org/datacommons/util/DcidGenerator.java
```diff
@@ -0,0 +1,369 @@
...
+          Vocabulary.STD_DEVIATION_VALUE,
+          Vocabulary.MEASUREMENT_DENOMINATOR,
+          Vocabulary.MEASUREMENT_QUALIFIER,
+          Vocabulary.SCALING_FACTOR);
+
+  private static final LogWrapper dummyLogCtx =
+      new LogWrapper(Debug.Log.newBuilder(), Path.of("/tmp/report.html"));
+
+  public static class Result {
+    public String dcid = new String();
```
beets, line 119: maybe also add a comment here that the result could be an empty dcid if there are errors

beets (approved):
i suppose the "java way" of handling the errors in the library would be to throw an error, rather than an empty result. you could also attach error messages there. it'll change how this library gets used, so just putting this out there as food for thought

util/src/main/java/org/datacommons/util/DcidGenerator.java
```diff
@@ -0,0 +1,369 @@
...
+          }
+          pvMap.put(prop, tv.getValue());
+        }
+      }
+    }
+
+    List<String> ordered_cprops = new ArrayList<>();
+    for (var pv : pvMap.entrySet()) {
+      var prop = pv.getKey();
+      var val = pv.getValue();
```
chejennifer, line 267: nit: i don't think val gets used

chejennifer (approved):
Nice!

util/src/main/java/org/datacommons/util/DcidGenerator.java
pradh: Done.

util/src/main/java/org/datacommons/util/DcidGenerator.java
pradh: Thanks, removed.

util/src/main/java/org/datacommons/util/DcidGenerator.java
pradh: That is a neat idea!  Need to remember you can do such things with constants in java :)

pradh (commented):
Thanks for the review!

(I had a git-fail when I merged my local repo to master, somehow messed up commit history, and wasted several minutes of my life. Git is so confusing!)

pradh:
> i suppose the "java way" of handling the errors in the library would be to throw an error, rather than an empty result. you could also attach error messages there. it'll change how this library gets used, so just putting this out there as food for thought

SG.  So our current error-handling sort of deviates from typical "java way" in that we throw an error only if its something systemic, rather than anything the user can cause.

We could do our usual "update LogWrapper", but I was worried about double-counting errors.  But yeah, like you suggest, we could return a string error message in Result... I've added a TODO.

PR #61, opened 2021-09-03
Batch existence check requests to DC

pradh:
Previously, we relied only on caching prior checks. But each DC call from India takes ~250ms, so ~5K calls (e.g., to validate [all SVs](https://github.com/datacommonsorg/schema/tree/main/stat_vars)), which takes 5m from my laptop (and 2m from cloud-top), takes 20m from India.

With this change, we first collect checks (along with a callback) to batch and invoke DC calls when we have collected enough. The same SV validation now takes ~8s and makes 53 calls (see screenshot).

Additional changes:
1. Add a check that all stat-vars include a DCID.  We could in theory have StatVars without DCID and we compute it, but in practice we don't.  This tripped @ajaits.
2. Fix minor regression introduced [here](https://github.com/datacommonsorg/import/pull/34/files#diff-368ed99815e72875840d6857e431260a508c7f7fa7f6a375d45d2b12365eb99dL104-L107).

Screen Shot 2021-09-03 at 9 04 36 AM

### DIFF
--- a/server/src/main/java/org/datacommons/server/ServerCommand.java
+++ b/server/src/main/java/org/datacommons/server/ServerCommand.java
@@ -14,11 +14,7 @@
 
 package org.datacommons.server;
 
-import com.google.cloud.storage.Blob;
-import com.google.cloud.storage.BlobId;
-import com.google.cloud.storage.Storage;
-import com.google.cloud.storage.StorageException;
-import com.google.cloud.storage.StorageOptions;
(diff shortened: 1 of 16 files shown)


beets (approved):
very cool!

chejennifer (approved):
This is awesome!

pradh:
Thanks for the reviews!

PR #64, opened 2021-09-10
Add library to resolve an in-memory MCF graph

pradh:
This will be used to resolve either a row of CSV (in TMCF mode), or a full input MCF.  This is an approximate port of logic [here](https://source.corp.google.com/piper///depot/google3/datacommons/import/util/internal_resolve.h).

Once https://github.com/datacommonsorg/import/pull/62 is submitted, this library will use the external ID resolver to resolve places.


chejennifer (approved):
looks good!

pradh:
Thanks for the review!

PR #69, opened 2021-09-15
Integrate resolution into dc-import tool (not enabled by default)

pradh:
+ Unify the genmcf and lint flows inside Processor.process(). The only difference between them now is whether MCF output gets generated or not.
+ New with genmcf: when resolution is enabled, for MCF inputs, we generate output resolved MCFs.
+ Plumb the ExternalIdMapper into McfResolver and use it for resolution.
+ The main flow is in Processor.process(). If resolution is enabled: first, loop over the nodes and tables to build all external ID mappings (do so with a dummy LogWrapper to avoid double counting errors); next, do the actual checking, resolution and writing.
+ Finally, for nodes that failed resolution, set the actual error reason as comment next to every node in the generated MCF file.

(Sorry about the large PR! In hindsight, I should have done the Processor refactor to unify genmcf and lint flows as a separate first PR. Please let me know if you prefer that.)


tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -142,4 +229,79 @@ public void checkAllNodes() throws IOException, InterruptedException, DCTooManyF
...
+      }
+    }
+    idResolver.drainRemoteCalls();
+  }
+
+  private void writeGraph(OutputFileType type, Mcf.McfGraph graph) throws IOException {
+    var writer = writers.getOrDefault(type, null);
+    if (writer == null) {
+      var fileString = outputFiles.get(type).toString();
+      logger.info("Writing to file-type " + type.name() + " in " + fileString);
```
chejennifer, line 286: nit: Maybe `"Writing to file " + fileString + " of type " + type.name()`

chejennifer (approved):
This is great!

tool/src/main/java/org/datacommons/tool/Processor.java
pradh: Done

util/src/main/java/org/datacommons/util/McfResolver.java
```diff
@@ -16,70 +17,103 @@
...
     while (true) {
-      if (firstRound || localRefReplacement.numUpdated > 0) {
+      if (round == 0 || dcidAssignment.numUpdated > 0) {
         // First round, or a new DCID got assigned, so we might have a local-ref to replace.
         localRefReplacement = replaceLocalRefs();
+        if (verbose) {
+          logger.info(
+              "LocalRef Replacement Round "
+                  + (round + 1)
+                  + " :: replaced "
```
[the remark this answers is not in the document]
pradh: Done

pradh (commented):
Thanks for the review!

Issue #67, opened 2021-09-14
Warn on duplicate nodes in the MCF


ajaits:
Generate a warning when there are duplicate nodes in the MCFs with the same set of properties.

Nodes with the same dcid, but differences in PVs could be reported as errors.

pradh:
There are legitimate cases where a node's DCID might have PVs coming from different files (e.g., geos), or could be repeated for nodes generated from CSV (e.g., [tmcf](https://github.com/datacommonsorg/data/blob/master/scripts/covid19_india/medical_tests_in_data/COVID19_tests_india.tmcf), [csv](https://github.com/datacommonsorg/data/blob/master/scripts/covid19_india/medical_tests_in_data/test_data/test2_expected_data.csv)).

And we have a "merging" functionality that deduplicates PVs of the same DCID.

Is this bug specifically about StatVars?  For that I think we could have a targeted check like this:
1. When StatVar nodes have the same content (aka generatd DCID), cannot have diverging assigned DCID
2. A StatVar DCID cannot be assigned to StatVar nodes with different content.

Would that work?

ajaits:
On Thu, 16 Sept 2021 at 07:32, Prashanth R ***@***.***> wrote:

> There are legitimate cases where a node's DCID might have PVs coming from
> different files (e.g., geos), or could be repeated for nodes generated from
> CSV (e.g., tmcf
> <https://github.com/datacommonsorg/data/blob/master/scripts/covid19_india/medical_tests_in_data/COVID19_tests_india.tmcf>,
> csv
> <https://github.com/datacommonsorg/data/blob/master/scripts/covid19_india/medical_tests_in_data/test_data/test2_expected_data.csv>
> ).
>
> And we have a "merging" functionality that deduplicates PVs of the same
> DCID.
>
> Is this bug specifically about StatVars?
>
Yes, this was in reference to a statvar mcf for a single import. Normally
we don't expect an import to generate duplicate StatVars. In this case some
statvars became dups as they were missing other distinguishing properties.
This is likely specific to imports. We could add a utility class that the
import scripts could use to emit statvar mcf with optional checks for dups.


> For that I think we could have a targeted check like this:
>
>    1. When StatVar nodes have the same content (aka generatd DCID),
>    cannot have diverging assigned DCID
>    2. A StatVar DCID cannot be assigned to StatVar nodes with different
>    content.
>
> Yes, both these checks would be useful.

I think part of this is covered in dc-import lint as I got an error when
[...]

PR #71, opened 2021-09-17
New Checker: detect duplicates across StatVar nodes in an import

pradh:
Related issue: https://github.com/datacommonsorg/import/issues/67


util/src/main/java/org/datacommons/util/McfChecker.java
```diff
@@ -67,36 +71,40 @@ public static boolean checkNode(
...
+      StatVarState svState,
       LogWrapper logCtx) {
     this.graph = graph;
     this.columns = columns;
     this.logCtx = logCtx;
     this.existenceChecker = existenceChecker;
+    this.svState = svState;
   }
 
   // Returns true if there was an sanity error found.
```
chejennifer, line 97: nit: looking at the logic, it looks like this method returns true if no sanity error is found

util/src/main/java/org/datacommons/util/StatVarState.java
```diff
@@ -0,0 +1,67 @@
...
+    if (generated.dcid.isEmpty()) {
+      // This is due to malformed SV node, which should have been handled in the checker.
+      return false;
+    }
+
+    var existingGeneratedDcid = curatedToGenerated.getOrDefault(curatedDcid, null);
+    if (existingGeneratedDcid != null && !existingGeneratedDcid.equals(generated.dcid)) {
+      logCtx.addEntry(
+          Debug.Log.Level.LEVEL_ERROR,
+          "Checker_SameDcidForDifferentStatVars",
```
chejennifer, line 38: the prefix "Checker_" seems a little ambiguous as the counter key. Maybe this could also take on the "MCF_" prefix or the "Sanity_" prefix. Same for next log entry.

util/src/main/java/org/datacommons/util/StatVarState.java
```diff
@@ -0,0 +1,67 @@
...
+    var existingCuratedDcid = generatedToCurated.getOrDefault(generated.dcid, null);
+    if (existingCuratedDcid != null && !existingCuratedDcid.equals(curatedDcid)) {
+      logCtx.addEntry(
+          Debug.Log.Level.LEVEL_ERROR,
+          "Checker_DifferentDcidsForSameStatVar",
+          "Found different curated IDs for same StatVar :: dcid1: '"
+              + existingCuratedDcid
+              + "', dcid2: '"
+              + curatedDcid
+              + "', one node: '"
```
chejennifer, line 58: nit: maybe "one node" -> "node"

chejennifer (approved):

chejennifer (approved):
Nice new checker!

util/src/main/java/org/datacommons/util/McfChecker.java
pradh: good catch!

util/src/main/java/org/datacommons/util/StatVarState.java
pradh: agreed... went with Sanity_ since this gets called in the context of the McfChecker which uses that.

util/src/main/java/org/datacommons/util/StatVarState.java
pradh: Done

pradh (commented):
Thanks for the review!

PR #72, opened 2021-09-17
Improve existence-check batching and resolution usability

pradh:
1. Use POST requests and increase existence-check batching by 10x (largely for India latencies)
2. Add more log messages for resolution flow
3. Add a LOCAL resolution mode where we don't need two passes, or recon calls
4. Fix some double-counting bugs in McfChecker lib


beets (approved):
subtle!

pradh:
Thanks for the review!

PR #73, opened 2021-09-17
Integrate stat checker into dc-import tool

chejennifer:
- For both genmcf and lint, if stat checks is enabled, we will validate the time series
- Counters for stat checks are added to main log counters
- stat checks are not enabled by default


tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -99,6 +105,10 @@ public static Integer process(Args args) throws IOException {
         processor.processNodes(Mcf.McfType.TEMPLATE_MCF);
       }
 
+      if (args.doStatChecks) {
+        processor.checkStats(graphAfterResolution);
```
pradh, line 109: I wonder if its slightly more readable to fold the checkStats() logic in within the if (args.doResolution) above? So what happens in which case (e..g, the long comment in checkStats()) is obvious.  Feel free to disagree :)

```
if (args.doResolution) {
  ...

  graph = processor.resolveNodes();
  if (statChecker != null) {
    ... extract from graph ...
  }
} else {
  if (statChecker != null) {
     ... extract from nodesForVariousChecks ...
  }
}

...

// We've all along been adding stats to statChecker, do the final check.
if (statChecker != null) {
  statChecker.check();
}
```

tool/src/test/resources/org/datacommons/tool/genmcf/resolution/output/report.json
```diff
@@ -25,7 +25,8 @@
     "LEVEL_WARNING": {
       "counters": {
         "Existence_MissingReference_typeOf": "1",
-        "Existence_MissingReference_Property": "2"
+        "Existence_MissingReference_Property": "2",
+        "StatsCheck_1_Sigma": "1"
```
pradh, line 29: Should we skip 1 Sigma? (have only 3)

tool/src/test/resources/org/datacommons/tool/genmcf/resolution/output/report.json
```diff
@@ -117,5 +118,99 @@
...
+  "statsCheckSummary": [{
+    "placeDcid": "wikidataId/Q1437",
+    "statVarDcid": "CumulativeCount_MedicalTest_ConditionCOVID_19_Positive",
+    "measurementMethod": "",
+    "observationPeriod": "",
+    "scalingFactor": "",
+    "unit": "",
+    "seriesLargestPercentDiff": {
+      "percentDifference": 0.0
+    }
```
pradh, line 131: This seems spammy... should we skip when percentDiff is 0?

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -329,12 +341,17 @@ protected static void checkDates(
...
+      inconsistentDateCounter.setCounterKey(inconsistentDateCounterKey);
       // When there are multiple date granularity, the problem points will be those that are of a
       // different date granularity than the most common one.
       dateLenList.sort((d1, d2) -> dateLen.get(d2).size() - dateLen.get(d1).size());
       for (int i = 1; i < dateLenList.size(); i++) {
         dateLen.get(dateLenList.get(i)).forEach(inconsistentDateCounter::addProblemPoints);
+        logCtx.incrementCounterBy(
+            Level.LEVEL_WARNING.name(),
+            inconsistentDateCounterKey,
+            dateLen.get(dateLenList.get(i)).size());
```
pradh, line 354: Wondering why we don't just update this counter once for the series? Add comments?

pradh (commented):
Very nice!

tool/src/main/java/org/datacommons/tool/Main.java
```diff
@@ -58,6 +59,21 @@
...
+      defaultValue = "false",
+      scope = CommandLine.ScopeType.INHERIT,
+      description =
+          "Checks integrity of time series by checking for holes, variance in values, etc.")
+  public boolean doStatChecks;
+
+  @CommandLine.Option(
+      names = {"-p", "--sample-places"},
+      scope = CommandLine.ScopeType.INHERIT,
+      description = "List of place dcids to run stats check on.")
```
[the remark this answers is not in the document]
chejennifer: I don't think the dependency needs to be asserted because if user enters sample places without enabling statchecks, it doesn't do anything, the sample places will just go unused. I added a log warning that alerts users that their sample places will go unused, do you think that's good enough? Also updated the description of the sample places option

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -253,6 +274,7 @@ private void resolveCommon(
         writeGraph(failureFile, failed);
       }
     }
+    return McfUtil.mergeGraphs(List.of(resolver.resolvedGraph(), resolver.failedGraph()));
```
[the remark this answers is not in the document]
chejennifer: Good call, updated to only return resolved graph.

tool/src/main/java/org/datacommons/tool/Processor.java
chejennifer: good suggestion, I feel like it makes things clearer. Updated

tool/src/test/resources/org/datacommons/tool/genmcf/resolution/output/report.json
chejennifer: Actually, yeah, looking at prod it looks like 1 and 2 sigma get computed but the dashboard only displays 3 sigma. Updated

tool/src/test/resources/org/datacommons/tool/genmcf/resolution/output/report.json
chejennifer: done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: I thought it would make sense for the counter number to match the number of problem points for this counter. added comment.

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -105,8 +110,9 @@ public static Integer process(Args args) throws IOException {
         processor.processNodes(Mcf.McfType.TEMPLATE_MCF);
       }
 
+      // We've been adding stats to statChecker all along, now do the actual check.
       if (args.doStatChecks) {
```
pradh, line 114: nit: be consistent with whether statChecker check happens outside (here) or inside (addStats) ?

tool/src/main/java/org/datacommons/tool/Main.java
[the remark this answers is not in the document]
pradh: sg

tool/src/main/java/org/datacommons/tool/Processor.java
pradh: also, perhaps checkStats() can just inlined

tool/src/test/resources/org/datacommons/tool/genmcf/resolution/output/report.json
```diff
@@ -125,50 +124,28 @@
...
+    "statVarDcid": "CumulativeCount_MedicalTest_ConditionCOVID_19_Positive",
     "measurementMethod": "",
     "observationPeriod": "",
     "scalingFactor": "",
-    "unit": "",
-    "seriesLargestPercentDiff": {
-      "percentDifference": 0.0
-    }
+    "unit": ""
   }, {
```
pradh, line 135: Can we drop the entry if there are no errors?

util/src/main/java/org/datacommons/util/StatChecker.java
pradh: My feeling is it is more useful to see the number of series' where a problem occurs, than the points in a single series?

pradh (approved):
Thanks for the changes!

tool/src/test/resources/org/datacommons/tool/genmcf/resolution/output/report.json
chejennifer: done

tool/src/main/java/org/datacommons/tool/Processor.java
chejennifer: done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: that makes sense, made the change

pradh (approved):
Thanks for the changes!

tool/src/test/resources/org/datacommons/tool/genmcf/resolution/output/report.json
```diff
@@ -14,11 +14,11 @@
...
-        "Resolution_OrphanLocalReference_parent": "1",
-        "Resolution_DcidAssignmentFailure_Song": "1",
-        "Resolution_DcidAssignmentFailure_StatisticalVariable": "1",
-        "Resolution_DcidAssignmentFailure_MusicAlbum": "1",
-        "Resolution_ReferenceToFailedNode_variableMeasured": "1"
+        "Resolution_OrphanLocalReference_parent": "2",
+        "Resolution_DcidAssignmentFailure_Song": "2",
+        "Resolution_DcidAssignmentFailure_StatisticalVariable": "2",
+        "Resolution_DcidAssignmentFailure_MusicAlbum": "2",
+        "Resolution_ReferenceToFailedNode_variableMeasured": "2"
```
pradh, line 21: Oh wondering why this is increased

chejennifer: I also noticed this, looking into it right now.

chejennifer: Merging from master had caused an extra round of resolution. Fixed now

pradh:
Ooh nice catch

PR #74, opened 2021-09-20
Add percent fluctuation counters to StatChecker

chejennifer:
- fold the percent fluctuation information into the StatValidationEntries and add those counters to the main log counters

### DIFF
--- a/util/src/main/java/org/datacommons/util/StatChecker.java
+++ b/util/src/main/java/org/datacommons/util/StatChecker.java
@@ -28,7 +28,6 @@ import org.datacommons.proto.Debug.DataPoint;
 import org.datacommons.proto.Debug.DataPoint.DataValue;
 import org.datacommons.proto.Debug.Log.Level;
 import org.datacommons.proto.Debug.StatValidationResult;
-import org.datacommons.proto.Debug.StatValidationResult.PercentDifference;
 import org.datacommons.proto.Debug.StatValidationResult.StatValidationEntry;
 import org.datacommons.proto.Mcf.McfGraph;
 import org.datacommons.proto.Mcf.ValueType;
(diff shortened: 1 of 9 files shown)


util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -272,21 +274,39 @@ protected static void checkPercentFluctuations(
...
+            percent100Counter.addProblemPoints(dp);
+            percent100Counter.setPercentDifference(currDelta);
+          }
+        } else if (Math.abs(currDelta) > 50) {
+          if (!percent50Counter.hasPercentDifference()
+              || Math.abs(currDelta) > Math.abs(percent50Counter.getPercentDifference())) {
+            percent50Counter.addProblemPoints(baseDataPoint);
+            percent50Counter.addProblemPoints(dp);
+            percent50Counter.setPercentDifference(currDelta);
+          }
```
pradh, line 298: Just wondering, why not compute the maxDelta like before, and include a single counter for the series? If there is a point with >500%, just an exemplar pair of points should be sufficient (given this will likely be used to spot-check the outliers)

chejennifer: Oh, I thought we wanted counters for each of the percent fluctuations >50, >100, >500? Or do we just want the max percent fluctuation?

pradh: Just max percent fluctuation. One series should map to a single such counter, so by looking at the counters users can tell how many series have 5x, vs. 0.5x fluctuations... (we might fine-tune the 50, 100, 500 buckets in future)

chejennifer: ah, ok, makes sense, updated

pradh (approved):
Thanks for the change!

PR #77, opened 2021-09-22
Profile performance and improve LogWrapper locking

pradh:
From profiling with JProfiler, it turns out that the LogWrapper locks were heavily contended (profile attached at the end).  So mainly restructure it to allow for more parallelism.

Previously, past 5 threads, throughput would drop, now the performance scales with threads (measured till 8). Comparing the throughput (measured as "Million nodes per minute") with https://github.com/datacommonsorg/import/pull/75, 

Mode                                                                                         |   `--num-threads=5`|  `--num-threads=8`
----------------------------------------------------------|----------------------|----------------------
`genmcf --existence-checks=False --resolution=LOCAL` |      4.6 ->  5.4 M/m      |   8   M/m
`lint --existence-checks=False`                                            |      7.5 ->  8.3 M/m      | 13.6 M/m

Looking at CPU profiles for the first genmcf with resolution case (screenshots below), here is the breakdown:
+ 25% in CSV parsing (TmcfCsvParser)
+ 30% in two node checks pre-/post-mutation (McfChecker)
+ 5% mutation (McfMutator)
+ 28% in resolution (McfResolver)
+ 8% in serializing/writing MCFs

Other (low-hanging) optimizations in the PR:
1. Switch UID which internally uses SecureRandom which is not MT-friendly to use ThreadLocalRandom ([explanation](https://issues.apache.org/jira/browse/PHOENIX-3070?attachmentSortBy=fileName)); these showed up in lock profiles too
2. Instead of getMapOrThrow(), return null and check (exceptions shows up as 10% under DcidGenerator.forStatVarObs)
3. Turn some constant members in McfChecker into statics (~1% spent there)

**CPU Profiles**
Screen Shot 2021-09-21 at 11 58 27 PM

Screen Shot 2021-09-21 at 11 40 20 PM
 
**Original Monitor Profile**
Screen Shot 2021-09-21 at 8 04 11 PM


util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -213,8 +211,6 @@ public void process(CSVRecord dataRow) throws IOException, InterruptedException
         boolean success =
             McfChecker.checkNode(Mcf.McfType.INSTANCE_MCF, currentNodeId, newNode, logCtx);
         if (success) {
-          logCtx.incrementInfoCounterBy("NumNodeSuccesses", 1);
-          logCtx.incrementInfoCounterBy("NumPVSuccesses", newNode.getPvsCount());
```
chejennifer, line 217: Curious why we're dropping this counter

util/src/main/java/org/datacommons/util/LogWrapper.java
```diff
@@ -49,10 +51,13 @@
...
-  private Instant lastStatusAt;
+  // Update with lock, but read without object lock.
+  private volatile Instant lastStatusAt;
+  // Updated and read with object lock.
   private long countAtLastStatus = 0;
-  private long currentCount = 0;
-  private final Set<String> countersWithErrors = new HashSet<>();
+  private AtomicLong currentCount = new AtomicLong();
+  // A copy of the counters that are updated into the Log before persisting.
+  private final List<ConcurrentHashMap<String, Long>> counterMaps = new ArrayList<>();
```
chejennifer, line 60: nit: add comment that position in list corresponds to number for Debug.Log.Level of the counterMap (that's what I inferred from the rest of this file, but not sure if I'm right)

chejennifer (approved):
Nice performance improvements!

util/src/main/java/org/datacommons/util/LogWrapper.java
pradh: Done

pradh (commented):
Thanks!

PR #79, opened 2021-09-23
Optimize extractSeriesFromGraph method in StatChecker

chejennifer:
- remove synchronized from extractSeriesFromGraph which is a method called for every node

### DIFF
--- a/util/src/main/java/org/datacommons/util/StatChecker.java
+++ b/util/src/main/java/org/datacommons/util/StatChecker.java
@@ -57,7 +57,7 @@ public class StatChecker {
   // scaling factor, and unit of the stat var observations of the series summary.
   private final Map<String, SeriesSummary> seriesSummaryMap;
   // key is place namespace + length of place dcid, value is set of place dcids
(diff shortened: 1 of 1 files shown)


util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -57,7 +57,7 @@
   // scaling factor, and unit of the stat var observations of the series summary.
   private final Map<String, SeriesSummary> seriesSummaryMap;
```
pradh, line 58: Maybe its worth summarizing the concurrency handling of this class (spanning seriesSummaryMap, samplePlaces and svObsValues data structures)?  The stuff we discussed on chat...

pradh (approved):
Thanks for the change!

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: done

chejennifer:
Thanks for the review!

PR #80, opened 2021-09-23
Various Updates to StatChecker

chejennifer:
- For DataHole counter, show list of all dates in the series
- For percentDifference values, multiply by 100 and truncate to 2 digit decimal
- Remove StatsCheck_MaxPercentFluctuationGreaterThan50 counter
- Limit number of StatValidationResult entries to covering 10 of each counter

Additional fix that is not related to StatChecker:
- Generated table mcf node files should have mcf extension instead of csv extension


util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -126,8 +140,19 @@ public synchronized void check() {
...
+        // entries for.
+        boolean shouldAddEntry = false;
+        for (StatValidationEntry entry : resBuilder.getValidationCountersList()) {
+          String counterKey = entry.getCounterKey();
+          if (countersRemaining.containsKey(counterKey)) {
+            shouldAddEntry = true;
+            countersRemaining.compute(counterKey, (k, v) -> v != null ? v - 1 : 0);
+            if (countersRemaining.get(counterKey) < 1) countersRemaining.remove(counterKey);
+          }
+          if (shouldAddEntry) logCtx.addStatsCheckSummaryEntry(resBuilder.build());
```
pradh, line 154: Should this be outside the for loop?

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -390,8 +415,12 @@ protected static void checkDates(
           StatValidationEntry.Builder dataHoleCounter = StatValidationEntry.newBuilder();
           String dataHoleCounterKey = "StatsCheck_Data_Holes";
           dataHoleCounter.setCounterKey(dataHoleCounterKey);
+          List<String> dateList = new ArrayList<>();
+          for (DataPoint dp : timeSeries) {
+            dateList.add(dp.getDate());
+          }
           dataHoleCounter.setAdditionalDetails(
-              "Data hole found between the dates: " + prev.toString() + " and " + dt.toString());
+              "Possible data hole found. Dates in this series: " + String.join(",", dateList));
```
pradh, line 423: ", " for readability?

tool/src/main/java/org/datacommons/tool/WriterPair.java
```diff
@@ -49,7 +49,7 @@ public void close() throws IOException {
   private BufferedWriter newWriter(Args.OutputFileType type) throws IOException {
     String filePath = args.outputFiles.get(type).toString();
     if (csvFile != null) {
-      String fileSuffix = FilenameUtils.removeExtension(csvFile.getName()) + ".csv";
+      String fileSuffix = FilenameUtils.removeExtension(csvFile.getName()) + ".mcf";
```
pradh, line 52: When I made this change I think I also broke genmcf test (https://github.com/datacommonsorg/import/blob/master/tool/src/test/java/org/datacommons/tool/GenMcfTest.java#L70-L76). I'll fix that after this goes in...

pradh (approved):
Thanks for the fixes!

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: good catch, done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: done

tool/src/main/java/org/datacommons/tool/WriterPair.java
chejennifer: ok, sounds good

chejennifer:
Thanks for the review!

PR #81, opened 2021-09-24
Fix genmcf test and enable stat-checks and local-resolution by default

pradh:
Also, fix flakiness, root-caused by @chejennifer, in `statchecks` e2e test.


chejennifer (approved):
Thanks for the changes!

PR #84, opened 2021-09-29
Summary Report Generation

chejennifer:
Generate summary report as an html file (can preview html files using: https://htmlpreview.github.io/). This is currently not enabled by default.

Generation of charts will come in later PR, this is just a primitive version of the report with the summary tables.

Examples:
* [statchecks test report](https://htmlpreview.github.io/?https://raw.githubusercontent.com/datacommonsorg/import/f80d9ee3cfd355ed86c834ceb2b2fec702d82c41/tool/src/test/resources/org/datacommons/tool/genmcf/statchecks/output/summary_report.html)
* [allfiletypes test report](https://htmlpreview.github.io/?https://raw.githubusercontent.com/datacommonsorg/import/f80d9ee3cfd355ed86c834ceb2b2fec702d82c41/tool/src/test/resources/org/datacommons/tool/lint/allfiletypes/output/summary_report.html)


beets (approved):
the report looks neat!

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
```diff
@@ -85,6 +89,7 @@ public void GenMcfTest() throws IOException {
       }
       argsList.add("--resolution=FULL");
       argsList.add("--output-dir=" + Paths.get(testFolder.getRoot().getPath(), testName));
+      argsList.add("--summary-report");
```
pradh, line 92: Unlike absl, for these flags I think we need to specify true or false, otherwise it seemed to toggle the default...

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -0,0 +1,109 @@
...
+import org.datacommons.proto.Debug.DataPoint.DataValue;
+import org.datacommons.proto.Debug.StatValidationResult;
+import org.datacommons.proto.Mcf.McfGraph;
+import org.datacommons.util.SummaryReportGenerator.StatVarSummary;
+
+// An object to hold all the series information for all the statistical variables for a place.
+public class PlaceSeriesSummary {
+  public static class SeriesSummary {
+    StatValidationResult.Builder validationResult;
+    TreeMap<String, DataPoint> timeSeries;
```
pradh, line 22: nit: comment why TreeMap (presumably to keep keys sorted?)

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private final Map<String, Map<Long, SeriesSummary>> svSeriesSummaryMap = new HashMap<>();
```
pradh, line 25: nit: document keys?

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -70,7 +67,8 @@
   private final LogWrapper logCtx;
   // key is a string made up of place dcid, stat var dcid, measurement method, observation period,
   // scaling factor, and unit of the stat var observations of the series summary.
```
pradh, line 69: nit: comment needs update?

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -471,4 +451,17 @@ private boolean checkSvObsValueInconsistency(McfGraph.PropertyValues node) {
       return true;
     }
   }
+
+  private synchronized void extractStatVarInfoFromNode(McfGraph.PropertyValues node) {
+    String svDcid = McfUtil.getPropVal(node, Vocabulary.VARIABLE_MEASURED);
```
pradh, line 456: From what I recall, this bubbles up in the profile, and we're looking up every prop thrice now:
1. here
2. checkSvObsInGraph
3. extractSeriesFromNode

The first two happen for every svobs too. Perhaps we could extract into a struct and pass it around? If you agree, maybe add TODO for future PR.

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -98,12 +97,21 @@ public StatChecker(LogWrapper logCtx, Set<String> samplePlaces, boolean verbose)
...
-      if (isSvObWithNumberValue(node.getValue()) && shouldExtractSeriesInfo(node.getValue())) {
-        extractSeriesInfoFromNode(node.getValue());
+      if (isSvObWithNumberValue(node.getValue())) {
+        // We will extract basic stat var information from every StatVarObservation nodes
+        extractStatVarInfoFromNode(node.getValue());
+        // We will only extract series information from StatVarObservation nodes about sample places
+        if (shouldExtractSeriesInfo(node.getValue())) {
+          String placeDcid = McfUtil.getPropVal(node.getValue(), Vocabulary.OBSERVATION_ABOUT);
+          PlaceSeriesSummary placeSeriesSummary =
+              placeSeriesSummaryMap.getOrDefault(placeDcid, new PlaceSeriesSummary());
```
pradh, line 111: I think this will allocate a new PlaceSeriesSummary() every time (given the heap pressure with large imports, this can matter a lot). Can we do: https://github.com/datacommonsorg/import/blob/7529566dcd5491ab376b3fadb193dd9c46d35db1/util/src/main/java/org/datacommons/util/StatChecker.java#L183

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -471,4 +451,17 @@ private boolean checkSvObsValueInconsistency(McfGraph.PropertyValues node) {
       return true;
     }
   }
+
+  private synchronized void extractStatVarInfoFromNode(McfGraph.PropertyValues node) {
+    String svDcid = McfUtil.getPropVal(node, Vocabulary.VARIABLE_MEASURED);
+    if (svDcid.isEmpty()) return;
+    StatVarSummary svMap = svSummaryMap.getOrDefault(svDcid, new StatVarSummary());
```
pradh, line 458: ditto... don't alloc every time?

util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import org.datacommons.proto.Debug;
+
+public class SummaryReportGenerator {
+
+  // An object to save the information about a stat var. This contains all the necessary getters to
+  // access the information in this object from SummaryReport.ftl
+  public static final class StatVarSummary {
```
pradh, line 25: It is somewhat strange that parts of this is updated in the fast-path and parts of it used just for rendering. I wonder if its worth separating them... maybe a TODO for this PR?

util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+    int numObservations = 0;
+    Set<String> places = new HashSet<>();
+    Set<String> mMethods = new HashSet<>();
+    Set<String> units = new HashSet<>();
+    Set<String> scalingFactors = new HashSet<>();
+    // This set will not be populated when StatVarSummary is generated by PlaceSeriesSummary
+    TreeSet<String> dates = new TreeSet<>();
```
pradh, line 32: Given this is updated in the fast-path, if we only need the ordering at rendering time, perhaps use Set<> and sort? (https://stackoverflow.com/a/4464394)

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -98,12 +97,21 @@ public StatChecker(LogWrapper logCtx, Set<String> samplePlaces, boolean verbose)
...
-  public void extractSeriesInfoFromGraph(McfGraph graph) {
+  // Given a graph, extract stat var info and time series info (about the chosen sample places) from
+  // the statVarObservation nodes.
+  public void extractStatsFromGraph(McfGraph graph) {
     for (Map.Entry<String, McfGraph.PropertyValues> node : graph.getNodesMap().entrySet()) {
-      if (isSvObWithNumberValue(node.getValue()) && shouldExtractSeriesInfo(node.getValue())) {
-        extractSeriesInfoFromNode(node.getValue());
+      if (isSvObWithNumberValue(node.getValue())) {
+        // We will extract basic stat var information from every StatVarObservation nodes
+        extractStatVarInfoFromNode(node.getValue());
```
pradh, line 106: We now do synchronized work for *every* SVObs. I think this will cause more contention. Probably add a TODO to look into optimizing this?

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -98,12 +97,21 @@ public StatChecker(LogWrapper logCtx, Set<String> samplePlaces, boolean verbose)
...
+        // We will only extract series information from StatVarObservation nodes about sample places
+        if (shouldExtractSeriesInfo(node.getValue())) {
+          String placeDcid = McfUtil.getPropVal(node.getValue(), Vocabulary.OBSERVATION_ABOUT);
+          PlaceSeriesSummary placeSeriesSummary =
+              placeSeriesSummaryMap.getOrDefault(placeDcid, new PlaceSeriesSummary());
+          placeSeriesSummary.extractSeriesFromNode(node.getValue());
+          placeSeriesSummaryMap.put(placeDcid, placeSeriesSummary);
```
pradh, line 113: Does this need synchronized access? Given the change to do some work in every call, I just wonder if we should make this function itself synchronized now.

pradh (commented):
Report looks great!!

Should we add the error logs to the report too (perhaps at the very end)?  For instance, its somewhat of a pain to read those from the JSON, and we're having to copy/paste to jsongrid to view.  Can happen in future PRs...

Could you test this change with many CSVs in multi-threaded mode (-n=4) before merging?

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
chejennifer: done

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
chejennifer: done

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
chejennifer: done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: good catch, done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
chejennifer: The thing is that the parts that are updated in the fast-path are saved to be read later for rendering .. but I can still add a TODO to look into this class to figure out a better way to handle everything

util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
chejennifer: done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: Actually, yes, I'll make this function itself synchronized and add a todo to look into optimizing this

chejennifer:
Also made a change of adding TEST_mode to SummaryReportGenerator to fix flaky tests

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
```diff
@@ -113,6 +115,7 @@ public void GenMcfTest() throws IOException {
             TestUtil.assertReportFilesAreSimilar(
                 expect, TestUtil.readStringFromPath(expected), TestUtil.readStringFromPath(actual));
           } else if (f.equals("summary_report.html")) {
+            // TestUtil.assertReportHtmlAreSimilar(expected, actual);
```
pradh, line 118: nit: remove ?

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -17,15 +17,21 @@
...
+
   public static class SeriesSummary {
     StatValidationResult.Builder validationResult;
+    // Key is date of each datapoint. Use treemap here to keep the dates sorted.
     TreeMap<String, DataPoint> timeSeries;
   }
 
+  public static boolean TEST_mode = false;
+  // Key in svSeriesSummaryMap is stat var dcid & key in the Map<Long, SeriesSummary> is a hash
+  // constructed using place dcid, stat var dcid, measurement method, observation period,
```
pradh, line 29: nit: do we need stat-var dcid again?

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -40,8 +46,8 @@ public void extractSeriesFromNode(McfGraph.PropertyValues node) {
     Hasher hasher = Hashing.farmHashFingerprint64().newHasher();
     hasher.putString(vres.toString(), StandardCharsets.UTF_8);
     Long hash = hasher.hash().asLong();
-    Map<Long, SeriesSummary> seriesSummaryMap =
-        svSeriesSummaryMap.getOrDefault(vres.getStatVarDcid(), new HashMap<>());
+    svSeriesSummaryMap.computeIfAbsent(vres.getStatVarDcid(), k -> new HashMap<>());
```
pradh, line 49: Doesn't computeIfAbsent return the value? That will save another map lookup.

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -104,6 +110,6 @@ public void extractSeriesFromNode(McfGraph.PropertyValues node) {
           });
       statVarSummaryMap.put(svSeriesSummary.getKey(), summary);
     }
-    return statVarSummaryMap;
+    return TEST_mode ? new TreeMap<>(statVarSummaryMap) : statVarSummaryMap;
```
pradh, line 113: comment?

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -99,18 +98,18 @@ public StatChecker(LogWrapper logCtx, Set<String> samplePlaces, boolean verbose)
...
         extractStatVarInfoFromNode(node.getValue());
         // We will only extract series information from StatVarObservation nodes about sample places
         if (shouldExtractSeriesInfo(node.getValue())) {
           String placeDcid = McfUtil.getPropVal(node.getValue(), Vocabulary.OBSERVATION_ABOUT);
-          PlaceSeriesSummary placeSeriesSummary =
-              placeSeriesSummaryMap.getOrDefault(placeDcid, new PlaceSeriesSummary());
-          placeSeriesSummary.extractSeriesFromNode(node.getValue());
-          placeSeriesSummaryMap.put(placeDcid, placeSeriesSummary);
+          placeSeriesSummaryMap.computeIfAbsent(placeDcid, k -> new PlaceSeriesSummary());
+          placeSeriesSummaryMap.get(placeDcid).extractSeriesFromNode(node.getValue());
```
pradh, line 112: ditto

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -453,15 +452,17 @@ private boolean checkSvObsValueInconsistency(McfGraph.PropertyValues node) {
   }
 
   private synchronized void extractStatVarInfoFromNode(McfGraph.PropertyValues node) {
+    // TODO (chejennifer): extract prop value into a struct and pass around instead of looking it up
+    // in multiple places
     String svDcid = McfUtil.getPropVal(node, Vocabulary.VARIABLE_MEASURED);
     if (svDcid.isEmpty()) return;
-    StatVarSummary svMap = svSummaryMap.getOrDefault(svDcid, new StatVarSummary());
+    svSummaryMap.computeIfAbsent(svDcid, k -> new StatVarSummary());
```
pradh, line 459: here too

util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
```diff
@@ -82,8 +85,13 @@ public static void generateReportSummary(
     Template template = cfg.getTemplate("SummaryReport.ftl");
     HashMap<String, Object> data = new HashMap<>();
     data.put("levelSummary", log.getLevelSummaryMap());
-    data.put("svSummaryMap", svSummaryMap);
     data.put("commandArgs", log.getCommandArgs());
+    if (TEST_mode) {
+      svSummaryMap = new TreeMap<>(svSummaryMap);
+      placeSeriesSummaryMap = new TreeMap<>(placeSeriesSummaryMap);
```
pradh, line 91: comment?

(is this for ordering tests?)

pradh (approved):
Thanks for the changes!

tool/src/test/java/org/datacommons/tool/GenMcfTest.java
chejennifer: done

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
chejennifer: we don't need the stat-var dcid again, but the string used to make the hash is generated using the StatValidationResult which contains the stat-var dcid. Since including the stat var dcid doesn't require any additional memory (the hash is just a Long created from a string), I don't think having the stat var dcid will cause any problems

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
chejennifer: done

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
chejennifer: done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/StatChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
chejennifer: done

chejennifer:
Thanks for the reviews!

PR #94, opened 2021-12-07
Add new properties for statvars

ajaits:
Add the following properties for statvars:
- utteranceTemplate: non constraint property with text values describing the statvarobs
- geo properties for India data sets


pradh (approved):
Thanks!

PR #95, opened 2021-12-09
Update html file comparison in e2e tests

chejennifer:
- When generating charts for the summary report, depending on the machine, the dimensions may vary slightly. In the e2e tests, instead of comparing the actual vs expected summary_report.html files just as strings, compare the elements on the page and their text content. This tests the important parts of the html file

### DIFF
--- a/tool/src/test/java/org/datacommons/tool/LintTest.java
+++ b/tool/src/test/java/org/datacommons/tool/LintTest.java
@@ -1,7 +1,6 @@
 package org.datacommons.tool;
 
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
-import static org.junit.Assert.assertEquals;
 
 import com.google.common.truth.Expect;
 import java.io.File;
@@ -71,8 +70,8 @@ public class LintTest {
           if (f.equals("report.json")) {
             TestUtil.assertReportFilesAreSimilar(
                 expect, TestUtil.readStringFromPath(expected), TestUtil.readStringFromPath(actual));
-          } else {
-            assertEquals(
+          } else if (f.equals("summary_report.html")) {
+            TestUtil.assertHtmlFilesAreSimilar(
                 TestUtil.readStringFromPath(expected), TestUtil.readStringFromPath(actual));
           }
         }
--- a/tool/src/test/java/org/datacommons/tool/TestUtil.java
+++ b/tool/src/test/java/org/datacommons/tool/TestUtil.java
@@ -35,6 +35,8 @@ import org.junit.rules.TemporaryFolder;
 
 // Common set of utils used in e2e tests.
 public class TestUtil {
+  private static String EMPTY_TEXT_HTML_ELEMENT_STRING = "<ELEMENT WITHOUT TEXT>";
+
   public static void assertReportFilesAreSimilar(Expect expect, String expected, String actual)
       throws IOException {
     Debug.Log expectedLog = reportToProto(expected).build();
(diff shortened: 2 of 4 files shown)


tool/src/test/java/org/datacommons/tool/LintTest.java
```diff
@@ -71,8 +70,8 @@ public void LintTest() throws IOException {
           if (f.equals("report.json")) {
             TestUtil.assertReportFilesAreSimilar(
                 expect, TestUtil.readStringFromPath(expected), TestUtil.readStringFromPath(actual));
-          } else {
-            assertEquals(
+          } else if (f.equals("summary_report.html")) {
+            TestUtil.assertHtmlFilesAreSimilar(
                 TestUtil.readStringFromPath(expected), TestUtil.readStringFromPath(actual));
           }
```
pradh, line 76: To confirm, there's no else case here?

tool/src/test/java/org/datacommons/tool/TestUtil.java
```diff
@@ -68,4 +74,28 @@ public static String readStringFromPath(Path filePath) throws IOException {
...
+      Element expectedElement = expectedElements.poll();
+      Element actualElement = actualElements.poll();
+      expectedElements.addAll(expectedElement.children());
+      actualElements.addAll(actualElement.children());
+      String expectedElementText = expectedElement.ownText();
+      String actualElementText = actualElement.ownText();
+      assertEquals(
+          "Actual HTML file has an element that differs in text content from what is expected",
+          expectedElementText,
+          actualElementText);
```
pradh, line 94: Is it worth using expect so we don't stop on the first diff?

tool/src/test/java/org/datacommons/tool/TestUtil.java
```diff
@@ -68,4 +74,28 @@ public static String readStringFromPath(Path filePath) throws IOException {
...
+    }
+    assertTrue(
```
pradh, line 96: Print the elements to aid debugging?

tool/src/test/java/org/datacommons/tool/TestUtil.java
```diff
@@ -68,4 +74,28 @@ public static String readStringFromPath(Path filePath) throws IOException {
...
+    Parser htmlParser = Parser.htmlParser();
+    Document expectedHtml = htmlParser.parseInput(expected, "expected");
+    Document actualHtml = htmlParser.parseInput(actual, "actual");
+    LinkedList<Element> expectedElements = new LinkedList<>(expectedHtml.children());
+    LinkedList<Element> actualElements = new LinkedList<>(actualHtml.children());
+    while (!expectedElements.isEmpty() && !actualElements.isEmpty()) {
+      Element expectedElement = expectedElements.poll();
+      Element actualElement = actualElements.poll();
+      expectedElements.addAll(expectedElement.children());
+      actualElements.addAll(actualElement.children());
```
pradh, line 88: I wonder about append all the actual and expected ownTexts into a single long string and comparing them?  Would the output of that be more readable than doing piecemeal (e.g., first element is an addition, there are extra elements, etc)?

tool/src/test/java/org/datacommons/tool/TestUtil.java
pradh: (Moot if we do a single full text assert)

tool/src/test/java/org/datacommons/tool/TestUtil.java
pradh: (Moot if we do a single full text assert)

tool/src/test/java/org/datacommons/tool/LintTest.java
chejennifer: yes, no else case here

tool/src/test/java/org/datacommons/tool/TestUtil.java
chejennifer: I like that idea, updated!

pradh: Thanks. Maybe we can append *all* expectedElements and actualElements and then compare them?

Then we can get rid of the two assertTrue() statements below?

(Feel free to submit :) )

pradh (approved):
Thanks for the change!

PR #97, opened 2021-12-15
Upgrade log4j version to 2.16.0

ajaits:
2.16.0 has a more complete fix and is recommended.


pradh (approved):
I don't know why the presubmit is not running, but could you confirm this locally with an `mvn package` (which runs all tests)

ajaits:
Yes, `mvn package` and `mvn test`, both run without errors.

PR #99, opened 2022-01-07
fix runtime error for non positive range

chejennifer:
- fix bug where if there is only one series value and the value is zero, `java.lang.IllegalArgumentException: A positive range length is required: Range[0.0,0.0]` gets thrown.


pradh2:
I can't review this since I'm unable to log into my account. Could you please get this reviewed by cyin, ajaits?

ajaits (approved):

PR #102, opened 2022-01-20
update documentation about flags

chejennifer:
- add an instruction in the README for getting information about flags that can be set
- update what the default value is for some flags


pradh (approved):

sharadshriram (approved):
Looks good! Thanks for the update :)

PR #104, opened 2022-02-08
Removing the requirement for StatVars to have a populationType.

jehangiramjad:
If a StatVar does not have a populationType, the tool will now generate a warning instead of an error. Note that if a StatVar has a 'PopulationType' field instead of 'populationType', the tool will still generate an error.


pradh (approved):
Nice!

PR #105, opened 2022-02-09
Pass HTTP 1.1 version for DC calls

pradh:
This fixes the presubmit failures [here](https://pantheon.corp.google.com/cloud-build/builds/3f58eb85-10f5-4d06-8334-a4d6954ccd4d;step=0?project=879489846695).

Possible reason:  There seems to have been a regression in Java 11 HttpClient, wherein, if the server supports only HTTP 1.1 (as our servers do), the default Java HttpClient connection (which uses HTTP2) throws this error.  See details in these bugs: [1](https://bugs.openjdk.java.net/browse/JDK-8218546), [2](https://bugs.openjdk.java.net/browse/JDK-8218623)

I suspect the Docker image that CloudBuild uses somehow uses a minor version of Java that includes the regression bug, but not the subsequent fix.  Handle this by setting the HTTP version that the servers support.

Bonus:  upgrade recon API endpoint to staging

### DIFF
--- a/util/src/main/java/org/datacommons/util/ExistenceChecker.java
+++ b/util/src/main/java/org/datacommons/util/ExistenceChecker.java
@@ -24,10 +24,8 @@ import org.datacommons.proto.Mcf;
 // This class is thread-safe.
 public class ExistenceChecker {
   private static final Logger logger = LogManager.getLogger(ExistenceChecker.class);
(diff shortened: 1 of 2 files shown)


chejennifer (approved):
thanks for the fix!

PR #107, opened 2022-02-23
Support schemaless SVs with init-cap mprop

pradh:
This is so that we can have schema-less SVs that have mprops that refer to the SV itself (which may start with an upper case letter).


chejennifer (approved):
thanks for the update!

pradh:
Thanks!

PR #114, opened 2022-05-27
Fix test errors

chejennifer:
- fix error when running LintTest which was caused by passing an empty dcid to the /node/property-values call. Do this by adding checks that dcid is not empty
- fix bug of infinite loop of independent RPC call issued -> RPC call fails -> same independent RPC call issued

### DIFF
--- a/util/src/main/java/org/datacommons/util/ExistenceChecker.java
+++ b/util/src/main/java/org/datacommons/util/ExistenceChecker.java
@@ -73,6 +73,7 @@ public class ExistenceChecker {
     if (checkLocal(node, Vocabulary.TYPE_OF, "", logCb)) {
       return;
     }
+    // assert !node.isEmpty();
     batchRemoteCall(node, Vocabulary.TYPE_OF, "", logCb);
   }
 
@@ -88,6 +89,7 @@ public class ExistenceChecker {
     if (checkLocal(sub, pred, obj, logCb)) {
       return;
     }
+    // assert !sub.isEmpty();
     batchRemoteCall(sub, pred, obj, logCb);
   }
 
@@ -203,6 +205,19 @@ public class ExistenceChecker {
     if (dataJson == null) {
       if (verbose) {
         logger.info("DC call failed for - " + Strings.join(subs, ',') + ", " + pred);
+      }
+      // If this was an independent RPC call, don't want to re-issue the call.
+      if (subs.size() == 1) {
+        var sub = subs.get(0);
+        if (!subMap.containsKey(sub)) {
+          return;
+        }
+        var objMap = subMap.get(sub);
+        for (var cbs : objMap.values()) {
+          totalPendingCallCount -= cbs.size();
+        }
+        subMap.remove(subs.get(0));
+        return;
       }
       // Important: If the dcid is malformed, Mixer can return failure. Also, if the URI is too
       // long, then too this happens. So issue independent RPCs now. If this happens often enough,
--- a/util/src/main/java/org/datacommons/util/ExistenceChecker.java
+++ b/util/src/main/java/org/datacommons/util/ExistenceChecker.java
@@ -73,7 +73,7 @@ public class ExistenceChecker {
     if (checkLocal(node, Vocabulary.TYPE_OF, "", logCb)) {
       return;
     }
-    // assert !node.isEmpty();
+    assert !node.isEmpty();
     batchRemoteCall(node, Vocabulary.TYPE_OF, "", logCb);
   }
 
@@ -89,7 +89,7 @@ public class ExistenceChecker {
     if (checkLocal(sub, pred, obj, logCb)) {
       return;
     }
-    // assert !sub.isEmpty();
+    assert !sub.isEmpty();
     batchRemoteCall(sub, pred, obj, logCb);
   }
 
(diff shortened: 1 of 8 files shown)


util/src/main/java/org/datacommons/util/ExistenceChecker.java
```diff
@@ -73,6 +73,7 @@ public synchronized void submitNodeCheck(String node, LogCb logCb)
     if (checkLocal(node, Vocabulary.TYPE_OF, "", logCb)) {
       return;
     }
+    // assert !node.isEmpty();
```
pradh, line 76: nit: why commented?

util/src/main/java/org/datacommons/util/McfChecker.java
```diff
@@ -438,7 +440,9 @@ private Set<String> checkCommon(String nodeId, Mcf.McfGraph.PropertyValues node)
                 .setDetail(LogCb.PREF_KEY, prop)
                 .setDetail(LogCb.NODE_KEY, nodeId)
                 .setCounterSuffix(Vocabulary.PROPERTY_TYPE);
-        existenceChecker.submitNodeCheck(prop, logCb);
+        if (!prop.isEmpty()) {
```
pradh, line 443: nit: can do:
`if (existenceChecker != null && !prop.isEmpty()) {` ?

util/src/main/java/org/datacommons/util/McfChecker.java
```diff
@@ -218,7 +218,9 @@ private void checkStatVar(String nodeId, Mcf.McfGraph.PropertyValues node)
               .setDetail(LogCb.OBJ_KEY, popType)
               .setDetail(LogCb.NODE_KEY, nodeId)
               .setCounterSuffix(Vocabulary.DOMAIN_INCLUDES);
-      existenceChecker.submitTripleCheck(mProp, Vocabulary.DOMAIN_INCLUDES, popType, logCb);
+      if (!mProp.isEmpty()) {
```
pradh, line 221: nit: `if (existenceChecker != null && !prop.isEmpty()) {` instead?

util/src/main/java/org/datacommons/util/ExistenceChecker.java
```diff
@@ -204,6 +206,19 @@ private void performDcCall(
       if (verbose) {
         logger.info("DC call failed for - " + Strings.join(subs, ',') + ", " + pred);
       }
+      // If this was an independent RPC call, don't want to re-issue the call.
+      if (subs.size() == 1) {
+        var sub = subs.get(0);
+        if (!subMap.containsKey(sub)) {
+          return;
```
pradh, line 213: This should not happen, right?  should we assert?

util/src/main/java/org/datacommons/util/ExistenceChecker.java
```diff
@@ -204,6 +206,19 @@ private void performDcCall(
...
+        }
+        var objMap = subMap.get(sub);
+        for (var cbs : objMap.values()) {
+          totalPendingCallCount -= cbs.size();
+        }
```
pradh, line 218: should we loop over the cbs and do: `cb.logError("Existence_FailedDCCall", "Failed to ...");` or something?

that way we bubble up the error instead of silently dropping.

util/src/main/java/org/datacommons/util/McfChecker.java
```diff
@@ -486,7 +490,9 @@ private Set<String> checkCommon(String nodeId, Mcf.McfGraph.PropertyValues node)
             if (prop.equals(Vocabulary.MEASUREMENT_METHOD)) {
               value = value.replace("dcAggregate/", "");
             }
-            existenceChecker.submitNodeCheck(value, logCb);
+            if (!value.isEmpty()) {
```
pradh, line 493: nit: `if (... existenceChecker != null && !tv.getValue().isEmpty()) {` instead?

pradh (approved):
Thank you for the quick fix!

Please feel free to submit this after addressing the comments.... esp the one about updating counter on failure.

util/src/main/java/org/datacommons/util/ExistenceChecker.java
chejennifer: forgot to remove the "//", fixed

util/src/main/java/org/datacommons/util/McfChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/McfChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/ExistenceChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/ExistenceChecker.java
chejennifer: done

util/src/main/java/org/datacommons/util/McfChecker.java
chejennifer: done

PR #122, opened 2022-06-08
Add sorting to StatVars table

enjoythecode:
- use DataTables on the StatVars table with only the "ordering" option
  enabled
- creates two new dependencies:
  1. [jQuery](https://jquery.org/); [MIT License](https://jquery.org/license/)
  2. [DataTables](https://datatables.net/); [MIT License](https://datatables.net/license/mit)
- regenerated test goldens because this is a `SummaryReport.ftl` template change


util/src/main/resources/SummaryReport.ftl
```diff
@@ -189,4 +195,15 @@
       </div>
     </#if>
   </body>
+  <script>
+    $(document).ready(function () {
+      $('#statvars-table').DataTable({
+        paging: false,
+        searching: false,
+        info: false,
+        order: [] // don't apply initial ordering (which is turned on by default)
```
chejennifer, line 204: curious why we're adding this new DataTable if ordering is not being applied? What is the goal of using DataTable?

enjoythecode: the addition of DataTable allows users to click on columns to sort the table according to their interests.

without specifying an empty `order` configuration, DataTables will default to sorting by the first column. This alphabetically sorts the table based on StatVar names, which does not make much sense.

this line removes that default sorting, and instead displays the statvars in the order they are present in the source HTML

if we want a default sort order, we can specify that in this line 204 that you've highlighted

enjoythecode:
unfortunately, [the HTML preview tool](https://htmlpreview.github.io/) I have been using does not work with external script tags (which breaks jQuery).

Instead, I created a code sandbox of an example report from this so that you can preview from the browser without pulling the changes: https://codesandbox.io/s/green-smoke-ghjtv5?file=/index.html

util/src/main/resources/SummaryReport.ftl
chejennifer: ohh, I see, I couldn't see it when I opened the html in the github html previewer so I couldn't tell what the change was but the code sandbox was useful!

util/src/main/resources/SummaryReport.ftl
```diff
@@ -189,4 +195,15 @@
       </div>
     </#if>
   </body>
+  <script>
+    $(document).ready(function () {
+      $('#statvars-table').DataTable({
```
chejennifer, line 200: do we want to do this for the series summary table for each place as well?

util/src/main/resources/SummaryReport.ftl
```diff
@@ -82,7 +88,7 @@
     <#if svSummaryMap?has_content>
       <div>
         <h2>StatVarObservations by StatVar</h2>
-        <table width="95%">
+        <table id="statvars-table" width="95%">
```
chejennifer, line 91: since the new DataTable has borders, maybe we can get rid of the borders for this table in the css

util/src/main/resources/SummaryReport.ftl
enjoythecode: I do not think that will be useful, since the values there are not really meant for comparison across series, but instead a description of the time-series within itself.

util/src/main/resources/SummaryReport.ftl
enjoythecode: done. [code sandbox](https://codesandbox.io/s/green-smoke-ghjtv5?file=/index.html) updated with the new HTML.

note that there are minor changes in the styling of DataTables that are 1) negligible, 2) requires non-trivial to match the rest of the document. I'm opting to not waste effort on this.

util/src/main/resources/SummaryReport.ftl
pradh: I actually think thats a good idea.  We have imports with a large number of SVs for a place, and for the same reasons that we may want to sort the main table (by SV name, unit, #obs, etc.) we might want to sort this too.

chejennifer (approved):
Nice changes!

PR #126, opened 2022-06-17
Show sample place names instead of DCIDs where possible 

enjoythecode:
- makes requests to the Data Commons API to get the place name
- displays this in both the Table of Contents and the actual `<details>` tag for the sample place in `summary_report.html`
- names fall back to the DCID when not available (which is the current behavior without this PR)
- DCIDs are still used in anchor links due to stability

Live `summary_report.html` preview: https://wentyb.csb.app/ 
(from test `genmcf/statchecks`)

Keep in mind while reviewing test outputs that [`geoId/07`](https://datacommons.org/browser/geoId/07) does not exist, however, it is used in test inputs as it follows [`geoId/06`](https://datacommons.org/browser/geoId/06) which is California. Because `geoId/07` does not exist, the name for that falls back to the DCID. The next real geoID is [`geoId/08`](https://datacommons.org/browser/geoId/08) (Colorado).


tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -116,14 +122,41 @@ public static Integer process(Args args) throws IOException, TemplateException {
     return retVal;
   }
 
+  private void fetchNamesforSamplePlaces(Boolean verbose) throws IOException, InterruptedException {
```
pradh, line 125: I wonder about moving this to ApiHelper as an API that given a list of DCIDs plus a property ("name" here), returns a map of DCID->value?

That way, we don't have such lower-level logic in Processor, and also move the "fetchNamesForSamplePlaces" callsite into StatsChecker?

pradh (commented):
Nice job!

tool/src/main/java/org/datacommons/tool/Processor.java
enjoythecode: I moved the the logic to `StatChecker`. I support the idea of moving the logic of parsing the API results to a Map in ApiHelper, however, I believe that is out of the scope of this PR. Once this is merged, I can work on that (along with moving other DC API calls to ApiHelper) in a separate PR.

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -154,12 +126,12 @@ private Processor(Args args) {
 
     // we initialize an httpClient regardless of args.doExistenceChecks
     // because other features might still make API calls
-    httpClient = HttpClient.newHttpClient();
+    this.httpClient = HttpClient.newHttpClient();
     if (args.doExistenceChecks) {
       existenceChecker = new ExistenceChecker(httpClient, args.verbose, logCtx);
```
pradh, line 131: this.httpClient (also below)

pradh (approved):

PR #127, opened 2022-06-22
Replace getPvsOrThrow with getPvsOrDefault in StringUtil.getPropVal for speed gain

enjoythecode:
# Background

I was running a profiler on the import tool for another reason, when I saw distinct "red peaks" where the code was spending a lot of time handling exceptions. I attributed these to three functions where exceptions are used to control program flow, as annotated in the image below.
image

I also found [this article](https://www.baeldung.com/java-exceptions-performance) detailing how using exceptions for program flow  can cause 10x (or more) slowdowns in Java code. So, I investigated if replacing exceptions in these code areas would be possible.

# Investigation

For `StringUtil.getPropVal`, the current method was accessing proto-generated Java code, so it was simple enough for me to find a way to circumvent throwing errors. (explained below, in "changes")

For `StringUtil.getValidISO8601DatePattern()`, I was unable to find a quick replacement for the use of exceptions. I found [this StackOverflow answer](https://stackoverflow.com/a/32315946) advertising Time4J as a library that can validate date strings without throwing exceptions, however, this involves a new dependency, so I erred on not delving into it in my first pass
- (reviewers: do you think a new dependency is worth the speed-up?)

For `StringUtil.IsNumber()`, I've looked into [`org.apache.commons.lang3.math.NumberUtils.isParsable`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html#isParsable-java.lang.String-), however, this does not meet our need of confirming scientific notation strings as valid. The other alternative, [`org.apache.commons.lang3.math.NumberUtils.isCreatable`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html#isCreatable-java.lang.String-) correctly returns `true` for scientific notation, however, it also validates hexadecimal numbers as correct, which goes against the current expected behavior, as coded in [`StringUtilTest.java:162`](https://github.com/datacommonsorg/import/blob/6e424820c22d406b4ebe53f3bf63b4eca00f2000/util/src/test/java/org/datacommons/util/StringUtilTest.java#L162).
- (reviewers: would it be worth considering the option where we allow hexadecimals as valid for the speed-gain?)

Therefore, I only implement changes for `StringUtil.getPropVal` in this PR.

# Changes to `StringUtil.getPropVal`
- getPropVal is a commonly used function where node.getPvsOrThrow was
  used along with a try/catch block for program flow.
- in place of getPvsOrThrow, use getPvsOrDefault (with a default value
  of null) to catch the case there the property does not exist in the
  node.
- in place of the try/catch block, use a simple not-null check (vals !=
  null)

# Confirming Speed-up
I ran `genmcf` on the FEMA NRI dataset (2 csv files; 1 with 10k rows, and another with 3.1k rows) on my local machine three times each with the code before and after the change.

Run times, in seconds:

Before these changes:
`71.656, 65.684, 69.291` (mean: `68.877`)

After these changes:
`61.420, 59.611, 61.859` (mean: `60.963`)

Approx. speed-up : `(68.877 - 60.963) / 68.877` = `11.4%`


pradh (approved):
Very nice analysis!

Regarding the speedup of the other two things:
1) For date, do you have a sense for how maintained/reliable/used Time4J is?  If it is generally used commonly, worth switching
2) For number, I think we probably don't want to change behavior for the speed up.

enjoythecode:
re: Time4J. It is not the _de facto_ time library. The _de facto_, mainstream libraries are built into the JDK (and that we use), however, their pattern of throwing exceptions (rather than allowing the user to query for the error somehow) is not performant for our use case where we expect a lot of exceptions.

General info: Much of the performance hit happens when the current call stack is traced upon creation of the Exception object. [source](https://stackoverflow.com/a/28075127)

Two other related ideas for speed improvements:
1. Caching the results of **failed** checks where we depend on exceptions. This 1) does not use memory for valid input (which is already fast) but 2) speeds up repeated checks of the same incorrect data.
2. For date formatting, create a map of input string length and possible patterns. For example, if the input is of the form `yyyy-MM-dd'T'HH:mm:ss.SSSXXX`, there is no need to try and fail to check against `yyyy`, `yyyy-MM`, (each of which raises an exception) etc... There are 11 patterns in `StringUtil.java`, and an extra 2 in regex for CPP parity.

pradh:
> re: Time4J. It is not the _de facto_ time library. The _de facto_, mainstream libraries are built into the JDK (and that we use), however, their pattern of throwing exceptions (rather than allowing the user to query for the error somehow) is not performant for our use case where we expect a lot of exceptions.
> 
> General info: Much of the performance hit happens when the current call stack is traced upon creation of the Exception object. [source](https://stackoverflow.com/a/28075127)
> 
> Two other related ideas for speed improvements:
> 
> 1. Caching the results of **failed** checks where we depend on exceptions. This 1) does not use memory for valid input (which is already fast) but 2) speeds up repeated checks of the same incorrect data.
> 2. For date formatting, create a map of input string length and possible patterns. For example, if the input is of the form `yyyy-MM-dd'T'HH:mm:ss.SSSXXX`, there is no need to try and fail to check against `yyyy`, `yyyy-MM`, (each of which raises an exception) etc... There are 11 patterns in `StringUtil.java`, and an extra 2 in regex for CPP parity.

I like the idea (2) of optimizing the checks we do, and continue to use the de facto library.

PR #129, opened 2022-06-22
Differ sample place generation grouping behavior based on namespace

enjoythecode:
# Old Behavior
For sample place generation, all place IDs are keyed by a unique key consisting of their namespace (the part before `/`, for example, `geoId` for `geoId/06`) and the length of the key.

This is a heuristic to try and get sample places of different types.

# The Issue
This heuristic works well for namespaces where the length of the key carries information about the place type, however, leads to noise and/or too few samples in cases where it doesn't.

# New Behavior
The namespaces `geoId` and `nuts` have meaningful ID lengths. For these namespaces, preserve old behavior.

For all other namespaces, do not bucket the place IDs into length buckets. Instead, have a single bucket with expanded capacity (in this PR, `25`, compared to `5` per length in the old behavior) per namespace.

# Other Changes
A test golden changed because the places included moves to single bucket in this change, which has a higher max size, so more places are included in the `summary_report.html` for the `genmcf/successtmcf` test

Extracted the logic for attempting to add the sample place to the `samplePlaces` map out of `shouldExtractSeriesInfo`, and into a new function `attemptAddingPlaceToSamplePlaces`. `shouldExtractSeriesInfo` was both checking whether series info should be extracted, and it was attempting to add the place to sample places key.

All `samplePlaces` keys are now computed in a new function `getSamplePlacesKey`, including when `shouldGenerateSamplePlaces` is `false`. This ensures consistent keying into `samplePlaces`, and abstracts away the details of how keys are calculated from other functions.


util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -40,7 +40,8 @@
 // b. making methods that access the map synchronized because when methods are not on a hot path and
 //    it is not ok for access to the map to be race-y (seriesSummaryMap).
 public class StatChecker {
-  private static final int MAX_NUM_PLACE_PER_TYPE = 5;
+  private static final int LENGTH_BASED_SAMPLE_PLACE_BUCKET_MAX_NUM_PLACE_PER_BUCKET = 5;
```
pradh, line 43: use a smaller name?
- MAX_PLACES_FOR_TYPE_INFERRED_NS
- MAX_PLACES_FOR_TYPELESS_NS

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -192,25 +194,80 @@ private boolean isSvObWithNumberValue(McfGraph.PropertyValues node) {
...
+    return namespace;
+  }
+
+  // given a place dcid, returns whether we want to bucket it by its length
+  private boolean isPlaceBucketedByLength(String placeDcid) {
+    String namespace = getNamespaceOfPlaceName(placeDcid);
+    return PLACE_NAMESPACES_BUCKETED_BY_LENGTH.contains(namespace);
+  }
+
+  // given a place dcid, returns the key in the samplePlaces map
```
pradh, line 243: nit: Given... (use full english casing)

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -192,25 +194,80 @@ private boolean isSvObWithNumberValue(McfGraph.PropertyValues node) {
   private boolean shouldExtractSeriesInfo(McfGraph.PropertyValues node) {
```
pradh, line 194: I wonder if we should have some targeted unit tests for this functionality.

For example, change the interface of shouldExtractSeriesInfo() to take a place DCID, make it public and exercise it in unit tests...

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -192,25 +194,80 @@ private boolean isSvObWithNumberValue(McfGraph.PropertyValues node) {
...
     String placeDcid = McfUtil.getPropVal(node, Vocabulary.OBSERVATION_ABOUT);
     if (placeDcid.isEmpty()) return false;
+
+    // if shouldGenerateSamplePlaces is true, and the bucket that the placekey
+    // belongs has empty space, the sample place will be added to the
+    // samplePlaces map
+    attemptAddingPlaceToSamplePlaces(placeDcid);
```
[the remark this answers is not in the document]
enjoythecode: I am on the fence about this, but preferred to break it up because `shouldExtractSeriesInfo` implies read-only, but `attemptAddingPlaceToSamplePlaces` is potentially write  (modifies samplePlaces field of the class). Maybe we should move `attempt...` out of `should...` to avoid having a state-modifying behavior inside a seemingly read-only function?

util/src/main/java/org/datacommons/util/StatChecker.java
enjoythecode: added in 04686d9

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -193,37 +193,40 @@ private boolean isSvObWithNumberValue(McfGraph.PropertyValues node) {
   // Only extract series information for a statVarObservation node that is about a sample place.
   private boolean shouldExtractSeriesInfo(McfGraph.PropertyValues node) {
     String placeDcid = McfUtil.getPropVal(node, Vocabulary.OBSERVATION_ABOUT);
```
pradh, line 195: This private function has one callsite, get the placeDcid there and remove this?

util/src/test/java/org/datacommons/util/StatCheckerTest.java
```diff
@@ -388,6 +388,48 @@ public void testFuncCheckSigma() {
     assertEquals(1, TestUtil.getCounter(logCtx.getLog(), "StatsCheck_3_Sigma"));
   }
 
+  @Test
+  public void testShouldExtractSeriesInfo() {
```
pradh, line 392: nice tests

util/src/main/java/org/datacommons/util/StatChecker.java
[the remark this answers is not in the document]
pradh: The long name of the function doing some pretty specific logic called from one place from a not-so-long function is a "smell test" that the function should be inlined.

For e.g., the long block of comment at the callsite describing the logic should instead just be inlined code.

pradh (approved):

util/src/test/java/org/datacommons/util/StatCheckerTest.java
enjoythecode: thanks.

util/src/main/java/org/datacommons/util/StatChecker.java
enjoythecode: done.

PR #131, opened 2022-06-23
Add existence checks for `observationAbout` references

enjoythecode:
Checks that the ID referenced in the `observationAbout` field of SVObs nodes exist, using the Data Commons API. This check is behind a new flag (`--oa`, `--observation-about`), which defaults to `true`.

### Performance
I performed tests on the performance effect of this change on my local machine. In a dataset with 1300 rows ([FEMA NRI](https://github.com/datacommonsorg/data/pull/699)) where each row is a unique place, the run times were as follows;

**`--oa=false`**
`64.144`, `66.921`, `68.490` -- Avg: `66.518`

**`--oa=true`**
`74.970`, `75.152`, `77.691` -- Avg: `75.937`

**Difference**
`(75.937 - 66.518) / 66.518` = `~14.1%` slower. 

Note that `ExistenceChecker` caches results locally, so when the ratio of unique place references to rows is lower, the slowdown should be less in magnitude. In other words, the results of this test likely represent an approximate upper-bound on the performance impact of these changes.


docs/usage.md
```diff
@@ -121,6 +121,12 @@ Specifies the directory to write output files.
 
 Default is `dc_generated/` within current working directory.
 
+### `-oa`, `--observation-about`
```
pradh, line 124: wonder if we should make it `-ep --existence-checks-place`

docs/usage.md
```diff
@@ -121,6 +121,12 @@ Specifies the directory to write output files.
...
+
+Specificies whether the observationAbout value of StatVarObservation nodes should be checked against the Data Commons API for existence.
```
pradh, line 126: Specifies whether to perform existence checks for places found in StatVarObservation nodes.

util/src/main/java/org/datacommons/util/McfChecker.java
```diff
@@ -53,15 +53,28 @@ public class McfChecker {
...
   private StatVarState svState;
+  boolean shouldCheckObservationAbout; // Should check for observationAbout existence on SVObs
 
   // Argument |graph| may be Instance or Template MCF.
+  public static boolean check(
+      Mcf.McfGraph graph,
+      ExistenceChecker existenceChecker,
+      StatVarState svState,
+      LogWrapper logCtx,
+      boolean shouldCheckObservationAbout)
```
pradh, line 64: nit: keep LogWrapper as last arg?

util/src/main/java/org/datacommons/util/McfChecker.java
```diff
@@ -269,6 +285,17 @@ private void checkSVObs(String nodeId, Mcf.McfGraph.PropertyValues node)
         node,
         Vocabulary.STAT_VAR_OBSERVATION_TYPE,
         Vocabulary.GENERIC_VALUE);
+
+    if (shouldCheckObservationAbout) {
+      if (existenceChecker != null) {
```
pradh, line 290: nit: if (shouldCheckObservationAbout && existenceChecker != null) {
}

pradh (commented):
Nice!

docs/usage.md
pradh: They are by large locations / places.  I was just suggesting that so it's more readable. You can include a more accurate description of course.

pradh (approved):

enjoythecode:
> Is there a chance the chance of such errors increase from defaulting to true? Because the e2e tests now do more RPCs...

In principle this is correct. However;
1) this specific network failure was when installing a module from maven repos
2) recent changes did not increase # of RPCs more than ~20%, I believe.

tool/src/main/java/org/datacommons/tool/Main.java
```diff
@@ -96,6 +96,16 @@ class Main {
       description = "Generates a summary report in html format. Defaults to true.")
   public boolean generateSummaryReport;
 
+  @CommandLine.Option(
+      names = {"-ep", "--existence-checks-place"},
+      defaultValue = "true",
```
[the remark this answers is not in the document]
pradh: (I thought sent this comment out, but it was in my draft)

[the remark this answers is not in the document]
enjoythecode: TODO(snny)

PR #134, opened 2022-06-27
Wizdetect

jehangiramjad:


chejennifer (approved):
this makes sense to me for the CsvData structure

wizard/src/types.ts
```diff
@@ -52,7 +52,31 @@ export interface MappingVal {
...
+  // a sample. This is the structure that should be used for detection
+  // heuristics and other advanced processing.
+  // It is assumed that all columns present in the original csv data file will
+  // be represented in this structure. All values in the orderedColumnNames
+  // array should be present as keys of columnValuesSampled.
+  // Note that the length of all column-values need not be the same, e.g. due to
+  // the removal of duplicate values.
+  columnValuesSampled: Map<string, Array<string>>,
+
+  // rowsForDisplay is a mapping from the row index in the original csv file to
```
pradh, line 74: Wondering why this is not an Array<Array<string>>?  Is that because we only want to retain certain rows?

If the latter, given the key is row index, should it be int?

pradh (approved):
Thanks!

wizard/src/types.ts
jehangiramjad: Only because this may not be the full csv. This could just be the first N rows and last N rows to display, like what print() produces on a Pandas dataframe (see example one here: https://www.geeksforgeeks.org/how-to-print-an-entire-pandas-dataframe-in-python/)

jehangiramjad: But you're right that the keys should be numbers anyway.

pradh: Got it, might be useful to note that in the comment.  Any reason the key shouldn't be long/int?

jehangiramjad: Great point! Updated both the type and the comment

PR #136, opened 2022-06-28
[beast] lint and type updates

chejennifer:
- add linting and fix current lint errors
- update CsvData type

### DIFF
--- a/build/cloudbuild.npm.yaml
+++ b/build/cloudbuild.npm.yaml
@@ -1,0 +1,29 @@
+
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This cloud build is triggered when there is code change in the wizard folder.
+# Run npm tests and lint
+
+# Run Python tests, lint, etc.
+steps:
+  - id: npm_test
+    name: npm:3.7
+    entrypoint: /bin/sh
+    args:
+      - -c
+      - |
+        cd wizard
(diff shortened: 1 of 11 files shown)


jehangiramjad (approved):

shifucun (commented):
Maybe add a new cloudbuild file in https://github.com/datacommonsorg/import/tree/master/build to enforce the test and lint pass

chejennifer:
> Maybe add a new cloudbuild file in https://github.com/datacommonsorg/import/tree/master/build to enforce the test and lint pass

done

build/cloudbuild.npm.yaml
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This cloud build is triggered when there is code change in the wizard folder.
+# Run npm tests and lint
+
+# Run Python tests, lint, etc.
+steps:
+  - id: npm_test
+    name: npm:3.7
```
shifucun, line 22: should name be "node:16"?

You can test this by running `cloud-build-local --config=build/cloudbuild.npm.yaml --dryrun=false .`

wizard/src/types.ts
```diff
@@ -69,24 +70,29 @@ export interface CsvData {
...
   // the orderedColumnNames.
-  rowsForDisplay: Map<BigInt, Array<string>>,
-}
+  rowsForDisplay: Map<BigInt, Array<string>>;
+
+  // if csv input was a user uploaded file, the uploaded csv file.
+  rawCsvFile?: File;
 
+  // if csv input was a user entered url, the url to get the csv file.
+  rawCsvUrl?: string;
```
pradh, line 87: nit: one of these must be set?

pradh (approved):
Nice!  Thanks for adding this.

shifucun (approved):

wizard/src/types.ts
chejennifer: I don't think there's a way to enforce that one of these is set, but I can add a comment

pradh: Yup, comment is good.

build/cloudbuild.npm.yaml
chejennifer: you're right, fixed!

wizard/src/types.ts
chejennifer: done

PR #137, opened 2022-06-29
Refactoring place detection to clearly distinguish Place types and Place properties.

jehangiramjad:


wizard/src/utils/detect_place.ts
```diff
@@ -13,9 +13,67 @@
...
+  PlaceProperty,
+  PlaceType,
+  TypeProperty,
+} from "../types";
 import countriesJSON from "./country_mappings.json";
 
+const MIN_HIGH_CONF_DETECT = 0.9;
+
+// All supported Place types must be encoded below.
+const PLACE_TYPES = new Map<string, PlaceType>([
```
pradh, line 28: Document what the key is?  Also I don't follow why the key is different from typeName, though they seem to match 1:1.

pradh: Ditto with PLACE_PROPERTIES below

wizard/src/types.ts
```diff
@@ -91,8 +91,14 @@ export interface CsvData {
...
+export interface PlaceType {
+  typeName: string;
+  displayName: string;
+}
+
+// The Data Commons property of a place, e.g. longitude.
+export interface PlaceProperty {
   propertyName: string;
   displayName: string;
 }
```
pradh, line 104: Can we generalize PlaceType and PlaceProperty so that this can be:

```
Entity {
  dcName: string;
  displayName: string;
}

export interface TypeProperty {
  type: Entity;
  property?: Entity;
}
```

wizard/src/utils/detect_place.ts
```diff
@@ -28,52 +86,61 @@ export class PlaceDetector {
...
 
-  placeTypes: Map<string, string>;
+  // A set of all place types and their associated properties.
+  placeTypesAndProperties: Set<TypeProperty>;
 
-  static typeFormatMappings = new Map<string, Array<DetectedFormat>>([
-    ["Longitude", [{ propertyName: "longitude", displayName: "Longitude" }]],
-    ["Latitude", [{ propertyName: "latitude", displayName: "Latitude" }]],
+  // Mapping between Place types and supported properties associated with each
+  // type.
```
pradh, line 93: Maybe I'm missing something, but not sure why we need the key indirection, and why this cannot just be a map from type as key and list of properties as values.

pradh (commented):
I like the direction of this change!

wizard/src/utils/detect_place.ts
jehangiramjad: Removed the key from here (not using the map structure here) and instead creating in-memory maps during initialization using the dcNames as keys.

wizard/src/types.ts
jehangiramjad: Done. With a slightly minor edit where I am also using PlaceType and PlaceProperty as extensions of Entity

wizard/src/utils/detect_place.ts
jehangiramjad: As discussed offline, the keys here are to match the user defined column names.

wizard/src/types.ts
```diff
@@ -91,12 +91,18 @@ export interface CsvData {
...
-export interface DetectedFormat {
-  propertyName: string;
+// An abstraction for a Data Commons entity, e.g. a place type or property.
+interface Entity {
+  dcName: string;
   displayName: string;
 }
 
+// A Data Commons entity type, e.g. Country (which a type of Place).
+export type DCType = Entity;
```
chejennifer, line 101: wondering if its necessary to define these new types DCType and DCProperty ..

wizard/src/utils/detect_place.ts
```diff
@@ -28,52 +67,75 @@ export class PlaceDetector {
...
-      "GeoCoordinates",
-      [{ propertyName: "GeoCoordinates", displayName: "Geo Coordinates" }],
-    ],
+  // A set of all place types and their associated properties.
+  placeTypesAndProperties: Set<TypeProperty>;
+
+  // Mapping between Place types and supported properties associated with each
+  // type.
+  static keyTypePropertyMappings = new Map<string, Array<TPName>>([
+    ["longitude", [{ tName: "GeoCoordinates", pName: "longitude" }]],
```
chejennifer, line 81: curious what the difference between the key of the keyTypePropertyMappings map and the tName values is because they've both been described as place type?

chejennifer (commented):
nice, thanks for refactoring!

wizard/src/types.ts
jehangiramjad: I don't think they are necessary. But I feel it's more readable and obvious what they are. Otherwise, it is easy to mistakenly swap the Property entity (DCProperty) with the Type entity (DCType).

wizard/src/utils/detect_place.ts
jehangiramjad: Good question and I should update the comment above. The difference is that the keys in the keyTypePropertyMappings are the (approx) column names that we might anticipate a user would put in their csv. So, when detecting if a particular column name matches a valid location type, we can't rely on just matching the DC Names. The keys here are what we expect to match with a user's csv headers. If there is a match, then the values in keyTypePropertyMappings are the DC Type and DC Property we can infer.

chejennifer (approved):
thanks for the changes!

wizard/src/utils/detect_place.ts
```diff
@@ -28,52 +67,78 @@ export class PlaceDetector {
...
-    ],
+  // A set of all place types and their associated properties.
+  placeTypesAndProperties: Set<TypeProperty>;
+
+  // Mapping between Place types and supported properties associated with each
+  // type. The keys of keyTypePropertyMappings are matched against the column
+  // headers in the user csv files. If a column header matches a key in
+  // keyTypePropertyMappings then the associated value in
+  // keyTypePropertyMappings is the inferred location type and property.
+  static keyTypePropertyMappings = new Map<string, Array<TPName>>([
```
pradh, line 83: nit: would columnToTypePropertyMap clarify it further... (feel free to ignore/defer)

pradh (approved):
Thanks!

wizard/src/utils/detect_place.ts
jehangiramjad: Done!

PR #142, opened 2022-06-30
[beast] update cloudbuild linter

chejennifer:


shifucun (approved):
very nice change and investigation!

PR #144, opened 2022-07-01
Integrating country detection in to heuristics.

jehangiramjad:
Note: currently, all properties for country detection are treated equal and ties are broken by picking the first one processed. In a later PR, we will introduce some order between the properties.


wizard/src/utils/heuristics.ts
```diff
@@ -14,13 +14,89 @@
...
+import { PlaceDetector } from "./detect_place";
 
 /**
- * Given a csv, returns the predicted mapping
+ * Process all columns and return the one which best represents the detected
+ * Place along with its details. If no Place is detected, the return value is
+ * null.
  *
- * @param csv array of csv rows where each row is an array of strings
+ * @param columns is a mapping from column indices to (sampled) column values.
```
chejennifer, line 36: nit: either change this to "cols" or change the param to "columns"

wizard/src/utils/heuristics.ts
```diff
@@ -14,13 +14,89 @@
...
+      }
+    }
+  });
+
+  if (detectedCountries.size > 0) {
+    // Choose the first detected country columns.
+    const ind = 0;
+    return {
+      type: MappingType.COLUMN, // Place detection is only possible for columns.
+      column: columnOrder[ind],
```
chejennifer, line 69: it looks like `ind` is always 0 .. which means the chosen column is always going to be the first one in columnOrder .. I'm guessing this isn't what you want?

chejennifer (commented):
Nice!

wizard/src/utils/heuristics.ts
jehangiramjad: Duh! Yes, I had forgotten to update the code when switching from an array to a map :( sorry!

wizard/src/utils/heuristics.ts
jehangiramjad: done

ajaits (approved):

PR #146, opened 2022-07-05
Ordering the country detection heuristic

jehangiramjad:


wizard/src/utils/heuristics.ts
```diff
@@ -14,13 +14,115 @@
...
+  //  4. Country name.
+
+  const propDetected = new Map<string, number>();
+
+  detectedCountries.forEach((details: DetectedDetails, index: number) => {
+    const prop = details.detectedTypeProperty.dcProperty.dcid;
+    propDetected.set(prop, index);
+  });
+
+  if (propDetected.has("isoCode")) {
```
chejennifer, line 45: nit: can just have a global constant holding an array with the country types ordered from most preferred to least preferred. Then, here, we can just iterate through the array & return if propDetected has that country type?

chejennifer (commented):
awesome!

wizard/src/utils/heuristics.ts
```diff
@@ -14,13 +14,115 @@
...
+  DCType,
+  DetectedDetails,
+  MappedThing,
+  Mapping,
+  MappingType,
+  MappingVal,
+} from "../types";
+import { PlaceDetector } from "./detect_place";
+
+function countryOrder(detectedCountries: Map<number, DetectedDetails>): number {
```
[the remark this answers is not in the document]
jehangiramjad: done!

wizard/src/utils/heuristics.ts
jehangiramjad: I am not sure that will improve either the functionality or the readability? I wanted to keep all the country property preference order local to this one function so that there are no other global (or other) dependencies. Let me know if you feel strongly?

chejennifer: I don't feel strongly, I just think it'll make the code cleaner by not having so many else-ifs

chejennifer (approved):

wizard/src/utils/heuristics.ts
jehangiramjad: Ok I made the suggested change except not by having a global array. The order array is local to the function.

PR #148, opened 2022-07-06
Filter to only supported place types/properties

jehangiramjad:
Also reduces the country detection threshold to 70%


chejennifer (approved):
thanks for the update!

PR #152, opened 2022-07-07
Supporting State detection (US and India)

jehangiramjad:
This PR does the following:

1. Adds support for States (type and properties)
2. Adds the static state names, isocodes, fips codes json file
3. Detection of States
4. Ordering between State properties
5. Ordering between States and Countries (State given preference)
6. Updates and adds new unit tests.


wizard/src/utils/detect_place.ts
```diff
@@ -260,6 +298,53 @@ export class PlaceDetector {
...
+   */
+  detectStateHighConf(column: Array<string>): TypeProperty {
+    let numValid = 0;
+
+    const counters = new Map<string, number>();
+    counters["name"] = 0;
+    counters["isoCode"] = 0;
+    counters["fips52AlphaCode"] = 0;
+    for (const cVal of column) {
+      if (cVal == null) {
```
chejennifer, line 320: do we want to add cVal if it's an empty string?

chejennifer, line 320: do we want to include cVal if it's an empty string?

wizard/src/utils/detect_place.ts
```diff
@@ -260,6 +298,53 @@ export class PlaceDetector {
...
+        continue;
+      }
+      const v = toAlphaNumeric(cVal);
+      numValid++;
+
+      if (this.stateNames.has(v)) {
+        counters["name"]++;
+      } else if (this.stateISO.has(cVal.toLowerCase())) {
+        // Note: this matches lower case only.
```
chejennifer, line 329: does casing matter for stateNames and fips code?

chejennifer (commented):
Cool!

I have not reviewed the heuristics.ts file because I'm confused about all the commented out code.. is everything commented out by accident? or was the code supposed to be deleted?

wizard/src/utils/detect_place.ts
jehangiramjad: Replaced with the lodash isEmpty() function.

wizard/src/utils/detect_place.ts
jehangiramjad: Sorry, they all use the helper function which deletes non-Alpha Numeric chars but also converts to lowercase. I have renamed the function to toAlphaNumericAndLower() from toAlphaNumeric(). Hope this makes the contrast clearer.

jehangiramjad:
Thanks for the review Jennifer! Yes, that commented out code was refactored and I forgot to delete it :(

wizard/src/utils/heuristics.ts
```diff
@@ -27,41 +27,41 @@ import {
...
-  detectedCountries.forEach((details: DetectedDetails, index: number) => {
+  detectedPlaces.forEach((details: DetectedDetails, index: number) => {
     const prop = details.detectedTypeProperty.dcProperty.dcid;
     propDetected.set(prop, index);
   });
 
-  for (let i = 0; i < propOrder.length; i++) {
-    if (propDetected.has(propOrder[i])) {
-      return propDetected.get(propOrder[i]);
+  for (let i = 0; i < propertyOrder.length; i++) {
```
chejennifer, line 62: nit: can do `for (const prop of propertyOrder) {...}` to access each value in the propertyOrder array

wizard/src/utils/heuristics.ts
```diff
@@ -84,31 +84,58 @@ function detectPlace(
...
+      detectedIndex = index;
+      detectedPlaces = detectedStates;
     }
+  } else if (detectedCountries.size > 0) {
+    // Get the index of the detected property according to a preference order.
+    const index = placePropertyOrder(detectedCountries, COUNTRY_PROPERTY_ORDER);
+    if (index != null) {
+      detectedIndex = index;
+      detectedPlaces = detectedCountries;
+      // return {
```
chejennifer, line 121: delete these lines?

wizard/src/utils/heuristics.ts
```diff
@@ -27,41 +27,41 @@ import {
...
- *    1. ISO code
- *    2. Alpha Numberic 3 Letter Abbreviation
- *    3. Numeric code
- *    4. Country name.
- *
- * @param detectedCountries is a mapping from column indices to the DetectedDetails
+ * @param detectedPlaces is a mapping from column indices to the DetectedDetails
  *    objects which contain the specifics of the type and property detected.
  *
  * @returns the column index of the most preferred country column.
```
chejennifer, line 49: "...the most preferred place column"

wizard/src/utils/heuristics.ts
```diff
@@ -84,31 +84,58 @@ function detectPlace(
   columnOrder: Array<Column>,
   pDetector: PlaceDetector
 ): MappingVal {
-  // Currently, only countries can be detected as Places.
+  // Currently, only countries and states can be detected as Places.
+  // Preference is given to States over Countries.
```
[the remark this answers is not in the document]
jehangiramjad: Took out the second line.

wizard/src/utils/heuristics.ts
jehangiramjad: Done

wizard/src/utils/state_mappings.json
```diff
@@ -0,0 +1,446 @@
+[{
+  "id": "geoId/01",
+  "name": "Alabama",
+  "iso_code": "US-AL",
+  "fips52AlphaCode": "AL"
```
pradh, line 5: Catching up on PRs:

(1) Should we also resolve by geoId ? (aka FIPS) This is also very common.

(2) We today do not support resolution by fips52AlphaCode.  We might need to add it to:
- http://google3/datacommons/import/mcf_vocab.h;l=296;rcl=460356614
- https://github.com/datacommonsorg/import/blob/master/util/src/main/java/org/datacommons/util/Vocabulary.java#L268
- https://github.com/datacommonsorg/mixer/blob/master/internal/server/recon/resolve.go#L37
(unfortunately, these are 3 different binaries...)

jehangiramjad: (1). Do you mean these as FIPS codes: https://www.nrcs.usda.gov/wps/portal/nrcs/detail/?cid=nrcs143_013696 If so, I wonder if this might dupe the detection more often than not because I can imagine numbers 01-79 appearing the data values column too.

(2). Interesting. I also don't see just FIPS codes in there. Would you also suggest adding those? Alternatively, do we have another property which refers to the 2-letter state codes?

PR #153, opened 2022-07-07
[beast] add mapping section

chejennifer:
- add the components for the mapping section

screen-recording-_44_


wizard/src/components/mapping_column_options.tsx
```diff
@@ -0,0 +1,211 @@
...
+import { ColumnInfo } from "./mapping_section";
+
+interface MappingColumnOptionsProps {
+  column: ColumnInfo;
+  onColumnUpdated: (column: ColumnInfo) => void;
+  validPlaceTypeProperties: Record<string, Set<string>>;
+  placeDetector: PlaceDetector;
+}
+
+const MAPPED_THING_NAMES = {
```
jehangiramjad, line 36: Even if we aren't using this here, if we're defining a mapping from MappedThing to names, shouldn't we include all the MappedThing keys? Perhaps we could move this to some constants file anyway (but that we can do later)?

wizard/src/components/mapping_column_options.tsx
```diff
@@ -0,0 +1,211 @@
...
+  [MappedThing.UNIT]: "unit of measure",
+  [MappedThing.VALUE]: "observation value",
+};
+
+export function MappingColumnOptions(
```
jehangiramjad, line 41: why does this function name start with a capital letter? Also, earlier we have an interface name starting with "Mapping.." and now a function name which also has the same start. Not sure we want to do that? Also, would be great to have a function comment?

wizard/src/components/mapping_column_options.tsx
```diff
@@ -0,0 +1,211 @@
...
+            checked={_.isEmpty(props.column.type)}
+            onChange={() => updateColumn(null, null)}
+          />
+          Not mapped
+        </Label>
+      </FormGroup>
+    </div>
+  );
+
+  function updateColumn(
```
jehangiramjad, line 177: nit: this function is being used by the function above. Ideally, we'd have this show up above in the file?

wizard/src/components/mapping_column_options.tsx
```diff
@@ -0,0 +1,211 @@
...
+    mappingType: MappingType,
+    mappedThing: MappedThing,
+    columnPlaceType?: string,
+    columnPlaceProperty?: string
+  ): void {
+    const updatedColumn = _.cloneDeep(props.column);
```
jehangiramjad, line 183: Where is props being used from? It's not provided as a function arg and I don't see a global variable?

wizard/src/components/mapping_column_options.tsx
```diff
@@ -0,0 +1,211 @@
...
+    updatedColumn.type = mappingType;
+    updatedColumn.mappedThing = mappedThing;
+    if (mappingType === MappingType.COLUMN) {
+      updatedColumn.columnMappedThing = mappedThing;
+      if (!_.isEmpty(columnPlaceProperty)) {
```
jehangiramjad, line 188: This is fairly hairy conditional magic. It would be helpful to add some comments about what the logical structure is. Can instead of nested conditionals, perhaps we could have specific ones where you combine multiple conditions together so it is clear when each condition would be hit?

jehangiramjad (commented):
Neat!

wizard/src/components/mapping_column_options.tsx
chejennifer: This is how a React component is defined! And there is a top level comment in this file saying what this component does

wizard/src/components/mapping_column_options.tsx
chejennifer: this has been the pattern we've been using for React components in the website code, but I will do some research to see what best practice is

wizard/src/components/mapping_column_options.tsx
chejennifer: done

wizard/src/components/mapping_column_options.tsx
jehangiramjad: Ok.

wizard/src/components/mapping_column_options.tsx
jehangiramjad: Cool

wizard/src/components/mapping_column_options.tsx
jehangiramjad: I see

wizard/src/components/mapping_column_options.tsx
jehangiramjad: thanks

jehangiramjad (approved):
Thanks for making the update and for answering my questions! Approved from my end but let's make sure Bo also has a look :)

wizard/src/components/mapping_column_options.tsx
```diff
@@ -0,0 +1,209 @@
...
+  const validPlaceProperties = Array.from(
+    props.validPlaceTypeProperties[props.column.columnPlaceType.dcid]
+  );
+  return (
+    <div className="column-options">
+      <div className="column-options-header">
+        The values in the column <b>{props.column.column.header}</b> are:
+      </div>
+      <FormGroup radio="true" className="column-options-input-section">
+        <Label radio="true" className="column-options-input column-type">
```
shifucun, line 54: Good to add a comment above each <Label> or <Input> to indicate what it is for, since all of them look very similar

wizard/src/components/mapping_section.tsx
```diff
@@ -0,0 +1,366 @@
...
+import { checkMappings } from "../utils/validation";
+import { MappingColumnOptions } from "./mapping_column_options";
+import { MappingConstantsSection } from "./mapping_constants_section";
+import { MappingSectionTable } from "./mapping_section_table";
+
+export interface ColumnInfo {
+  // the column of interest
+  column: Column;
+  // the mapped thing selected for column mapping type
+  columnMappedThing: MappedThing;
```
shifucun, line 45: why columnMappedThing and headerMappedThing are different?

wizard/src/components/mapping_section.tsx
```diff
@@ -0,0 +1,366 @@
...
+        <div className="selected-column-container">
+          <MappingColumnOptions
+            column={selectedColumnInfo}
+            onColumnUpdated={(columnInfo: ColumnInfo) =>
+              onColumnInfoUpdated(selectedColumnId, columnInfo)
+            }
+            validPlaceTypeProperties={validPlaceTypeProperties.current}
+            placeDetector={props.placeDetector}
+          />
+          <Button
```
shifucun, line 139: similar here, add comments on what these <Button> do?

wizard/src/components/mapping_section_table.tsx
```diff
@@ -0,0 +1,183 @@
...
+  // true if the last row is empty.
+  const lastRow = props.csvData.rowsForDisplay.get(
+    orderedDataRows[orderedDataRows.length - 1]
+  );
+  const hideLastRow =
+    _.isEmpty(lastRow) || (lastRow.length === 1 && _.isEmpty(lastRow[0]));
+  return (
+    <table id={TABLE_ID}>
+      <thead className="mapping-info-row">
+        <tr>
```
shifucun, line 84: I wonder if it worth to have a small component for each <tr> here to reduce nested level.

shifucun (approved):
Very structured PR! Just left some comments for readability.

wizard/src/components/mapping_section.tsx
chejennifer: The selected input for columnMappedThing and headerMappedThing could be different: https://screenshot.googleplex.com/3AgXskBkmBXmxsJ

wizard/src/components/mapping_section.tsx
chejennifer: done

wizard/src/components/mapping_section_table.tsx
chejennifer: done

chejennifer:
Thanks for the reviews!

PR #154, opened 2022-07-07
[beast] add preview section

chejennifer:
- add section to show example observations
- functionality for downloading import package as a zip file

Screen Shot 2022-07-07 at 10 21 59 AM


wizard/src/utils/file_generation.ts
```diff
@@ -36,16 +37,33 @@ export function generateSvMCF(csv: CsvData, mappings: Mapping): string {
   return "";
 }
 
+// convert Mapping object to a regular javascript object so it can be converted
+// into a JSON string
+function mappingToObject(mapping: Mapping): Record<string, MappingVal> {
+  if (_.isEmpty(mapping)) {
```
jehangiramjad, line 43: is this really needed? If mapping is empty, the for loop won't be hit, right? If so, then the empty check isn't needed?

wizard/src/utils/file_generation.ts
```diff
@@ -36,16 +37,33 @@ export function generateSvMCF(csv: CsvData, mappings: Mapping): string {
...
+    return {};
+  }
+  const obj = {};
```
jehangiramjad, line 46: nit: shouldn't this technically be "let obj = {};" ?

jehangiramjad (commented):
A couple of nitty comments. But I think you'll need to wait for Bo's review for more specific JSX insights.

wizard/src/utils/file_generation.ts
chejennifer: no, it will always be the same object, we're just adding entries to it. It's just like a map where the map object doesn't change, we're just setting key value pairs.

wizard/src/utils/file_generation.ts
chejennifer: ah yes, removed

wizard/src/utils/file_generation.ts
jehangiramjad: Ah I guess you can change/update members of a const object. Weird :)

shifucun (approved):
This is very readable and clear!

chejennifer:
Thanks for the reviews!

PR #155, opened 2022-07-07
Add header and footer to match dc.org

shifucun:
image


wizard/src/index.html
```diff
@@ -24,9 +24,108 @@
   <body>
     <div id="main">
       <header id="main-header">
-        Data Commons Import Wizard
+        <nav class="navbar navbar-dark navbar-expand-lg col" id="main-nav">
+          <div class="container-fluid">
+            <div class="navbar-brand">
+              <a href="http://datacommons.org">Data Commons</a>
```
chejennifer, line 30: I think it should be "https://datacommons.org" - here and everywhere else

chejennifer (commented):
Nice, thanks!

PR #156, opened 2022-07-07
Add favicon.ico to bundle

shifucun:


chejennifer (commented):
ah, that's what the copy plugin is used for. Awesome, thanks!

chejennifer (approved):

PR #158, opened 2022-07-08
Minor fixes to ensure WDI data columns are detected correctly.

jehangiramjad:
1. Lower the place detection threshold (this needed some fixes in tests because now two letter states and two letter countries can match more easily).
2. date parsing was ignoring negative offsets but that means all dates < 1970 were not being detected. removed that check.
3. We support ISO-8601 dates but don't want date strings to be less than 4 chars.


wizard/src/utils/detect_date.ts
```diff
@@ -18,8 +18,11 @@ import _ from "lodash";
 const MIN_HIGH_CONF_DETECT = 0.9;
 
 function detectDate(header: string): boolean {
+  // Date is detected if header is a valid ISO-8601 format.
+  // Note also: since we are only interested in detecting dates (and not time),
+  // if the header string length is <= 3, we consider that to be invalid.
   const d = Date.parse(header);
-  return !Number.isNaN(d) && d >= 0;
+  return !Number.isNaN(d) && header.length > 3;
```
chejennifer, line 25: maybe also add a test for this?

chejennifer (commented):
Thanks for the updates!

wizard/src/utils/detect_place.ts
```diff
@@ -25,7 +25,7 @@ import {
 import countriesJSON from "./country_mappings.json";
 import statesJSON from "./state_mappings.json";
 
-const MIN_HIGH_CONF_DETECT = 0.7;
+const MIN_HIGH_CONF_DETECT = 0.4;
```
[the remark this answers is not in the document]
jehangiramjad: yeah, it's because of those non-country abbreviations. we can discuss with the team but for now Guha said he'd be fine with a super low threshold to make sure we are unblocked on Monday.

wizard/src/utils/detect_date.ts
jehangiramjad: there is one in heuristics_test* which does this one (it is in countryDetectionOrder()). So i don't think we need one more.

chejennifer (approved):
thanks for the changes!

PR #163, opened 2022-07-12
Explicitly Disallow Not-a-Number StatVarObservation Values

enjoythecode:
### Current Behavior
Currently, Not-a-Number (NaN; text, reference, complex value, etc.) values in the `value` field of `StatVarObservation` nodes do not raise an error.

### Changes
- With this PR, these values will be logged at the `ERROR` level.
- Also adds a flag to allow NaN SVObs values; `--allow-non-numeric-svobs`, which defaults to `false`. If set to `true`, NaN values will NOT log an error.
- Also adds the ability to specify command-line arguments to integration tests in `GenMcfTest.java`. Specifying an `args.txt` file in the input folder will add each line of the file as an argument to the execution.
  - This is used with the integration tests `genmcf/measurementresult` and `genmcf/successmcf` using the flag added in this PR.

### Notes
To preserve unit-test behavior, McfChecker allows NaN values by default when called from outside `Processor.java`. We might look into this as we develop more functionality around NaN SVObs, specifically, `statType: measurementResult`, etc.


docs/counters.md
```diff
@@ -385,6 +385,13 @@ Found a property reference that does not start with a lower-case.
 
 **Suffix Description:** The property, and optionally, the type of the node separated with an underscore (`_`) from the property.
 
+### Sanity_SVObs_Value_NotANumber
+
+Found a SVObs whose `value` was not a number. If you are importing a dataset
+where this is expected (for example, `statType` is `measurementResult` and
+therefore the SVObs values are references), set `--allow-non-numeric-svobs=true`
```
pradh, line 392: nit: --allow-non-numeric-obs-values ?

util/src/main/proto/Debug.proto
```diff
@@ -84,6 +84,7 @@ message CommandArgs {
     optional bool stat_checks = 4;
     repeated string sample_places = 5;
     optional bool observation_about = 6;
+    optional bool allow_nan_svobs = 7; // allow not-a-number svobs values
```
pradh, line 87: nit: two letter space

util/src/main/java/org/datacommons/util/McfChecker.java
```diff
@@ -54,17 +54,25 @@ public class McfChecker {
...
       StatVarState svState,
       boolean shouldCheckObservationAbout,
+      boolean allowNanSVObs,
       LogWrapper logCtx)
       throws IOException, InterruptedException {
     return new McfChecker(
-            graph, null, existenceChecker, svState, shouldCheckObservationAbout, logCtx)
+            graph,
+            null,
+            existenceChecker,
```
pradh, line 71: not related to this PR:  It might be worth putting the args to a "struct" like class and passing that into various util functions here and elsewhere.  add a TODO?

util/src/test/java/org/datacommons/util/McfCheckerTest.java
```diff
@@ -271,7 +271,9 @@ public void checkSVObs() throws IOException, InterruptedException {
             + "value: \"DataSuppressed\"\n"
             + "observationAbout: dcid:geoId/SF\n"
             + "observationDate: \"2020\"\n";
-    assertTrue(success(mcf));
+    // This should only succeed when we allow not-a-number SVObs values
```
pradh, line 274: Should we have a focussed unit test in McfCheckerTest that exercises Sanity_SVObs_Value_NotANumber ?

pradh (commented):
Nicely done!

tool/src/test/resources/org/datacommons/tool/genmcf/measurementresult/output/summary_report.html
```diff
@@ -53,6 +53,8 @@
     </div>
     <h1>Summary Report</h1>
     <h3>Table of Contents</h3>
+    <button onclick="open_all_details()">Expand All</button>
+    <button onclick="close_all_details()">Collapse All</button>
```
pradh, line 57: puzzled by this change?

pradh (approved):
Thanks for the fixes!

enjoythecode:
The `genmcf/measurementresult` e2e test was added in this PR on the first commit ([72037fa](https://github.com/datacommonsorg/import/pull/163/commits/72037fa93359d9f26f725b228ca6e2734b247c20)) Merging with the main branch (which got the expand/collapse all button after these changes branched out) required to re-generate that golden to include those buttons.

tool/src/test/resources/org/datacommons/tool/genmcf/measurementresult/output/summary_report.html
enjoythecode: put the comment in the wrong place, see thread.

also note that this preview on the GitHub UI links to "changes from 7 commits" (from your last review), not the entirety of the PR.

PR #164, opened 2022-07-13
Detect FIPS codes for US States

jehangiramjad:
In this PR, we have the following updates:

1. Support for detecting FIPS codes for US States
2. FIPS codes for states can be mistaken for Numeric codes for countries. Added some logic to disambiguate the two when additional information may be available. Specifically, if the column header has "country" or "state" in it, then country/state detection can be biased in favor of country/state. 
3. Updated some existing unit tests and added new ones.


wizard/src/utils/detect_place.ts
```diff
@@ -213,13 +216,19 @@ export class PlaceDetector {
     this.stateNames = new Set<string>();
```
pradh, line 216: nit: is this function more general than preProcessUSStates?  Given isoCodes exist everywhere?

pradh (approved):
Thanks for the change!

PR #167, opened 2022-07-13
 Fix test `report.json` flaky order

### LINKED ISSUE #166: `report.json` from test has flaky order
### Problem
`report.json` generated from integration tests under `tool/src/test/resources/org/datacommons/tool/` have flaky order of some keys when running golden generation. A few of the keys tend to show up in swapped order. This causes 1) distractions to the developer who needs to stage the files differently, and 2) noise in diffs.

(note: golden generation command is `mvn -DgoldenFilesPrefix=$PWD/tool/src/test/resources/org/datacommons/tool test`)

Note that this flakiness does NOT fail the tests; `TestUtil.assertReportFilesAreSimilar` [ignores repeated field order](https://github.com/enjoythecode/import/blob/d182efadc4d685c3620817e50cd2972fe1b8aeeb/tool/src/test/java/org/datacommons/tool/TestUtil.java#L49)

### Example 
noisy diff committed to the repo:
https://github.com/datacommonsorg/import/commit/59736fab8f496f8debd0e007aa18c8ab13c13e79#diff-c68a8bd8e55421f20dbdaa33c1bcf138025b41d428832b4daf679ef52b145516L127-L136

### Desired Solution

It would be nice if TEST_MODE also applied to report.json such that the noise in the diffs were eliminated.

enjoythecode:
Closes #166 

### Changes
- LogWrapper.java calls sortLogEntries() when LogWrapper.TEST_MODE is
  true
  - sortLogEntries() sorts entries by building each entry, and then
    sorting using custom sort defined in SortByCounterKeyThenUserMessage
  - the Comparator function uses CounterKey as the first key, and if
    that is equal between to Log.Entry objects, uses UserMessage
  - Updated GenMcfTest.java and LintTest.java to correctly set
    LogWrapper.TEST_MODE to true
  - Updated LogWrapperTest.java to to set TEST_MODE to true. While doing
    this, refactored the common set up between two test functions in
    that file.
- Re-generated goldens with sorted counter keys
  - updated tool/**/report.json
  - updated LogWrapperTest_result.json
- remove unused print statement in LogWrapper.java

### Testing

Tested that this change addresses the flakiness by running `mvn clean && mvn golden` and observing no diffs detected by git.

Performed this 6 times back to back with no diffs.

### DIFF
--- a/tool/src/test/java/org/datacommons/tool/GenMcfTest.java
+++ b/tool/src/test/java/org/datacommons/tool/GenMcfTest.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import org.apache.commons.io.FilenameUtils;
+import org.datacommons.util.LogWrapper;
 import org.datacommons.util.SummaryReportGenerator;
 import org.datacommons.util.TmcfCsvParser;
 import org.junit.Rule;
@@ -61,6 +62,7 @@ public class GenMcfTest {
     // Set this so that the generated node IDs are deterministic
     TmcfCsvParser.TEST_mode = true;
     SummaryReportGenerator.TEST_mode = true;
+    LogWrapper.TEST_MODE = true;
 
     String goldenFilesPrefix = System.getProperty("goldenFilesPrefix");
     Main app = new Main();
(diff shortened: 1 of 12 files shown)


pradh (approved):
Nice change!

PR #168, opened 2022-07-14
Support NaN values in StatChecker

enjoythecode:
Builds on top of #163 -- once that PR is merged, this PR will be merged into that to clean up the diff.


util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -42,13 +45,34 @@ public String getValueString() {
       List<String> valueStrings = new ArrayList<String>();
 
       for (DataPoint dv : this.timeSeries.values()) {
-        valueStrings.add(Double.toString(dv.getValues(0).getValue()));
+        TypedValue tv = dv.getValues(0).getValue();
+        String valueString = tv.getValue();
+        // if the value is a number, parse it to Double and back to get the .0 suffix
```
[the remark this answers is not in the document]
enjoythecode: I wrote it in to minimize golden diff, very optional whether to keep it.

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -109,9 +133,14 @@ public synchronized void extractSeriesFromNode(McfGraph.PropertyValues node) {
     // Add the value of this StatVarObservation node to the timeseries of this node's SeriesSummary.
     String obsDate = McfUtil.getPropVal(node, Vocabulary.OBSERVATION_DATE);
     String value = McfUtil.getPropVal(node, Vocabulary.VALUE);
+    // TODO: WHAT TO PUT FOR isResolved?
```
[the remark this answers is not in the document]
enjoythecode: obsolete, since we are using the already parsed node TypedValue

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -109,9 +133,14 @@ public synchronized void extractSeriesFromNode(McfGraph.PropertyValues node) {
...
+    // TODO: WHAT TO PUT FOR logCb?
```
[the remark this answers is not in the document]
enjoythecode: obsolete, since we are using the already parsed node TypedValue

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -63,7 +87,7 @@ public String getTimeSeriesSVGChart() {
                 localDateTime.getDayOfMonth(),
                 localDateTime.getMonthValue(),
                 localDateTime.getYear()),
-            timeSeriesDataPoint.getValue().getValues(0).getValue());
+            Double.parseDouble(timeSeriesDataPoint.getValue().getValues(0).getValue().getValue()));
```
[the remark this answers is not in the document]
enjoythecode: done.

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -42,13 +45,34 @@ public String getValueString() {
...
+        valueStrings.add(valueString);
       }
       return String.join(" | ", valueStrings);
     }
 
+    public ValueType getValueType() {
+      // We can return the type of the first DataPoint
+      // Whether the types in this series is consistent is checked by
+      // StatChecker.checkSeriesTypeInconsistencies
+      return new ArrayList<>(this.getTimeSeries().values())
```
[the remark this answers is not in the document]
enjoythecode: We only create SeriesSummary when adding a node that does not belong to an existing SeriesSummary and immediately add it. I can add a check to reduce cognitive load and future proof it, however.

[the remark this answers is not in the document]
enjoythecode: Done.

tool/src/test/resources/org/datacommons/tool/genmcf/successtmcf/output/summary_report.html
```diff
@@ -817,6 +823,25 @@ <h2>
...
                     </tr>
+                    <tr>
+                      <td><a href="#places--CA-SK--SV1" name="places--CA-SK--SV1">SV1</a></td>
+                      <td>1</td>
+                      <td>2019</td>
+                      <td>65</td>
```
enjoythecode, line 830: this is an error. the header and source rows in the CSV is:
```
Year,Place_Id,Place_Name,Place_Type,LatLong,SV1,SV2
2019,CA-SK,Saskatchewan,Province,[LatLong 43N 116W],"[5 6]",2
```

that is, the value `"[5 6]"` is being rendered as a `65`.

TODO(snny) investigate this

enjoythecode: I have investigated this. The tool is understanding `"[5 6]"` as a complex value meaning of "`5` of the unit `6`".

A few issues here;
1. Is the test case intentional? Before this PR allowed NaN SVObs values, there was no corresponding output in `summary_report.html` for this row, so it is hard to understand the desired behavior.
  - There is, however, a counter in `report.json` for `6` not being a valid unit. [(link)](https://github.com/datacommonsorg/import/blob/1f3b72bac5ddd86d57d87c64f8790f4f4617e376/tool/src/test/resources/org/datacommons/tool/genmcf/successtmcf/output/report.json#L279-L287)
2. `ComplexValueParser.java` attempts to understand if the value or the unit is given first in a complex value of the format `[value1 value2]` by checking if the first digit of the first field is a digit [(link)](https://github.com/datacommonsorg/import/blob/1f3b72bac5ddd86d57d87c64f8790f4f4617e376/util/src/main/java/org/datacommons/util/ComplexValueParser.java#L118), however, this fails because this test case seemingly has a numeric unit.
3. Even with such a parsing, the resulting TypedValue is `type: RESOLVED_REF` and `value: 65`. The `type` seems correct, but the `value` is unexpected; shouldn't the unit come later?

Two further thoughts about better supporting NaN values in StatChecker;
1. We should display the "type" of a series (facet) to make it clear how the values were parsed.
2. How should we display complex values in `summary_report.html`? This might not be as important to cover if we implement (1) and do not expect real-world imports to have "numeric" units (even if they do; that error will be surfaced as a failed existence check for the unit)

@pradh could you please review and let me know your thoughts?

enjoythecode: > We should display the "type" of a series (facet) to make it clear how the values were parsed.

https://github.com/datacommonsorg/import/pull/174 is an implementation of this. It is in draft status until this PR is merged to master.

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -256,21 +271,46 @@ private int getSamplePlaceBucketSize(String placeDcid) {
...
+    for (DataPoint dp : timeSeries) {
+      for (DataValue val : dp.getValuesList()) {
+        ValueType type = val.getValue().getType();
+        if (firstType == null) {
+          firstType = type;
+        }
+        if (!firstType.equals(type)) {
+          inconsistentTypeCounter.addProblemPoints(dp);
+          logCtx.incrementWarningCounterBy(counterKey, 1);
+        }
```
pradh, line 291: We probably want a pair of inconsistent points.  How about we add the first point when "firstType == null" and when you find "firstType != type", also add that point, add to resBuilder and return....

tool/src/test/resources/org/datacommons/tool/genmcf/measurementresult/output/report.json
```diff
@@ -12,7 +12,7 @@
     "LEVEL_WARNING": {
       "counters": {
         "Existence_MissingTriple_domainIncludes": "1",
-        "API_EmptyDcCallResponse": "1"
+        "StatsCheck_Inconsistent_Values": "2"
```
pradh, line 15: Is this the input CSV? https://github.com/datacommonsorg/import/blob/master/tool/src/test/resources/org/datacommons/tool/genmcf/measurementresult/input/acre.csv

Wonder why the counter is 2?

Also, can we add more data points to trigger inconsistent value types in a series?

tool/src/test/resources/org/datacommons/tool/genmcf/successtmcf/output/summary_report.html
pradh: Good catch!

Yeah, the complex-value parsing could use some improvement.

So, in the default flag mode, where we don't allow string values for obs, does it error out?

I worry that the value-type is really implementation detail and adding a column for it is not very useful in *most* cases (e.g., unless its a NUMBER its a failure).  Perhaps we add a column only when something is a non-NUMBER? That is, if the measurement-result flag is enabled.

enjoythecode: > So, in the default flag mode, where we don't allow string values for obs, does it error out?

TODO(snny) to look into it

> Perhaps we add a column only when something is a non-NUMBER? That is, if the measurement-result flag is enabled.

Good idea. I will update #174 accordingly (TODO(snny)) and we can use that to decide if we want to keep or discard that change.

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -146,14 +145,30 @@ public synchronized void check() {
...
-          checkPercentFluctuations(timeSeries, resBuilder, logCtx);
           // Check for holes in dates, invalid dates, etc.
           checkDates(timeSeries, resBuilder, logCtx);
+
+          ValueType type = seriesSummary.getValueType();
+
+          List<String> stringSeries = new ArrayList<String>();
+          for (DataPoint dp : timeSeries) {
+            stringSeries.add(SeriesSummary.getValueOfDataPoint(dp));
+          }
```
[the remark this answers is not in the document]
enjoythecode: vestigial code; deleted.

util/src/main/java/org/datacommons/util/PlaceSeriesSummary.java
```diff
@@ -108,10 +147,15 @@ public synchronized void extractSeriesFromNode(McfGraph.PropertyValues node) {
 
     // Add the value of this StatVarObservation node to the timeseries of this node's SeriesSummary.
     String obsDate = McfUtil.getPropVal(node, Vocabulary.OBSERVATION_DATE);
-    String value = McfUtil.getPropVal(node, Vocabulary.VALUE);
+
+    // We never expect to get null here, since the node would have been dropped.
+    // The value will already have been parsed into a TypedValue, so we can depend on the fact that
+    // typedValue.getType() is not the default enum value (UNKNOWN_VALUE_TYPE).
+    McfGraph.Values nodeValues = node.getPvsOrDefault(Vocabulary.VALUE, null);
+    TypedValue typedValue = nodeValues.getTypedValues(0);
```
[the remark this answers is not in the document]
enjoythecode: good catch, done.

tool/src/test/resources/org/datacommons/tool/genmcf/measurementresult/output/report.json
enjoythecode: > Is this the input CSV?

Yes.

> Wonder why the counter is 2?

Good catch, I don't know yet, will investigate this further.

An interesting thing here from the report.json; the `lineNumber` of rows 3 and 4 are both logged as 3. Relevant section; (marked the lineNumber rows)
```
"values": [{
          "value": {
            "type": "RESOLVED_REF",
            "value": "Acre1.0",
            "column": "Acre_MeasurementResult_StatVar"
          },
          "locations": [{
            "file": "acre.csv",
            "lineNumber": "2"       <========
          }]
        }, {
          "value": {
            "type": "RESOLVED_REF",
            "value": "Acre10.0",
            "column": "Acre_MeasurementResult_StatVar"
          },
          "locations": [{
            "file": "acre.csv",
            "lineNumber": "3"       <========
          }]
        }, {
          "value": {
            "type": "RESOLVED_REF",
            "value": "Acre100.0",
            "column": "Acre_MeasurementResult_StatVar"
          },
          "locations": [{
            "file": "acre.csv",
            "lineNumber": "3"       <========
          }]
        }]
      }, {
```


> Also, can we add more data points to trigger inconsistent value types in a series?

TODO(snny)

tool/src/test/resources/org/datacommons/tool/genmcf/successtmcf/output/summary_report.html
enjoythecode: > So, in the default flag mode, where we don't allow string values for obs, does it error out?

Yes, however, the report.json counter value is still displayed as `65`, which is confusing. See relevant counter below.

```
    "level": "LEVEL_ERROR",
    "location": {
      "file": "SuccessTmcf.csv",
      "lineNumber": "13"
    },
    "userMessage": "Found a non-numeric value for the value field of StatVarObservation but allowNanSVObs was false :: value: '65', property: 'value', node: 'E:SVTest->E0'",
    "counterKey": "Sanity_SVObs_Value_NotANumber"
```

pradh: That's ok. As long as it gets parsed as a string.  If there's some way to report the original value that might be nice.

enjoythecode: I am working on reporting the type as it is parsed, since it is directly relevant to the error. Original value would likely touch  deep in the parsing logic.

enjoythecode: See: 870477d368074f84653e35e838c78fcdb3d8ad0a

tool/src/test/resources/org/datacommons/tool/genmcf/measurementresult/output/report.json
enjoythecode: > Also, can we add more data points to trigger inconsistent value types in a series?

Done

util/src/main/java/org/datacommons/util/StatChecker.java
enjoythecode: Done.

tool/src/test/resources/org/datacommons/tool/genmcf/measurementresult/output/report.json
enjoythecode: > > Wonder why the counter is 2?

> Good catch, I don't know yet, will investigate this further.

See #177 for a discussion about what is happening here and potential future directions. Marking this conversation resolved since that thread now has a place and the other request was completed.

pradh (approved):

Issue #169, opened 2022-07-15
Change import tool plot from diamond to smaller circle


shifucun:
For long time series, the plot get dense the diamonds can overlap with each other.

![image](https://user-images.githubusercontent.com/5951856/179273105-7c3a458d-0118-4f3e-b70e-b29ce79e896b.png)

enjoythecode:
Hi @shifucun, thanks for the request. Are you able to provide a dataset to reproduce this?

In terms of solutions, I have some ideas (and likely more things we can do), roughly ordered in complexity + completeness;
1. Simply decrease the diamond size and/or make into a circle
2. Add a flag to customize the size of the diamond/circle
3. Move to JavaScript based plotting that is easier to customize either automatically based on the number of points, or manually using a button

enjoythecode:
I will create 2 draft PRs, one for making the diamond smaller, one for replacing it with a small circle.

To reproduce those changes on your dataset, you can run clone those branches, `mvn package`, and then run the tool as you are used to.

enjoythecode:
See draft PRs;

1. Small diamond: https://github.com/datacommonsorg/import/pull/171
2. Small circle: https://github.com/datacommonsorg/import/pull/170

Note that these are too small for datasets with few datapoints; so the final implementation should have a command line argument to specify the size. We can use these to see which size better supports large timeseries, and then I can write a PR with the flags.

shifucun:
I like the small circle option. Looking at the test html file, seems the dot and line are disconnected. After fixing that, I would vote for the small circle change.

![image](https://user-images.githubusercontent.com/5951856/179284064-9cb9754e-5bec-422f-81b1-ceae3ad43f3e.png)

enjoythecode:
Closed by #170

PR #170, opened 2022-07-15
diamond sizing option #2: small circle

enjoythecode:


shifucun:
Can you try make the dot a different color as the line, like black? Otherwise, this looks much better.

shifucun (approved):

shifucun:
I like this a lot. Please let Prashanth check too.

enjoythecode:
Closes #169

util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
```diff
@@ -95,15 +95,37 @@ public static String constructSVGChartFromTimeSeries(TimeSeries timeSeries) {
...
+      JFreeChart chart = ChartFactory.createTimeSeriesChart("", "", "", null);
       XYPlot plot = chart.getXYPlot();
-      // create and use a renderer to draw each data point on the time series chart as a diamond and
-      // remove the legend
-      XYItemRenderer renderer = new XYLineAndShapeRenderer(true, true);
-      renderer.setSeriesShape(0, ShapeUtils.createDiamond(5));
-      renderer.setSeriesVisibleInLegend(0, false);
-      plot.setRenderer(renderer);
-      // change the background color of the chart to be white
+      // create and use separate renderers for the line and the shapes to be able
```
pradh, line 100: nit: Maybe this could be in a helper function? setRenderers(plot)

pradh:
> I like this a lot. Please let Prashanth check too.

+1.  I like this too!

pradh (approved):

util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
enjoythecode: My thought is to that this is perhaps for future development around plotting; because this function is the only place in the codebase we deal with SVG/charting so far.

PR #172, opened 2022-07-15
[beast] make table headers editable

chejennifer:
- make the table headers editable
- generate clean csv if table headers have been edited or if column id is different from column header

[screen-recording (45).webm](https://user-images.githubusercontent.com/69875368/179317727-236b3a7f-1cbe-4016-a8c7-fed05868cb10.webm)

updated at: https://chejennifer.github.io/import/wizard/


wizard/src/components/mapping_table_header.tsx
```diff
@@ -18,36 +18,65 @@
...
 export function MappingTableHeader(
   props: MappingTableHeaderProps
 ): JSX.Element {
+  const [editHeaderState, setEditHeaderState] = useState<EditHeaderState>({
+    colIdx: -1,
+    header: "",
+  });
+  const orderedColumns = Array.from(props.columnInfo.keys()).sort((a, b) => {
+    if (a > b) {
+      return 1;
```
pradh, line 48: nit: Trying to understand why the default sort() function or sort().reverse() won't work given these are numbers?

wizard/src/types.ts
```diff
@@ -34,9 +34,12 @@ export const MAPPED_THING_NAMES = {
...
 export interface Column {
-  // id of the column
+  // Id of the column.
+  // If in the original csv, there are two or more columns with the same header,
+  // id will be <header>_<columnIdx>. Otherwise, id will be the same as the
+  // header.
   id: string;
-  // original column header
+  // column header name
   header: string;
```
pradh, line 43: Comment that this may include user column name corrections?

wizard/src/components/mapping_table_header.tsx
```diff
@@ -56,31 +85,81 @@ export function MappingTableHeader(
...
+    const info = props.columnInfo.get(editHeaderState.colIdx);
+    const updatedColumn = {
+      ...info.column,
+      id: editHeaderState.header,
+      header: editHeaderState.header,
+    };
+    props.onColumnUpdated(editHeaderState.colIdx, {
+      ...info,
+      column: updatedColumn,
+    });
```
pradh, line 160: To confirm, if after user edit there are duplicates how do we handle it?

wizard/src/components/preview_section.tsx
```diff
@@ -85,19 +106,38 @@ export function PreviewSection(props: PreviewSectionProps): JSX.Element {
...
+            link.onclick = () => {
+              setTimeout(() => window.URL.revokeObjectURL(url));
+            };
+            link.click();
+            link.remove();
+          })
+          .catch(() => {
+            setIsGeneratingFiles(false);
+            alert("Sorry, there was a problem generating the import package.");
+          });
```
pradh, line 136: For my understanding, since this is wrapped with outside catch(), can we just let the exception here propagate up and skip this?

pradh (commented):
Nice!

wizard/src/components/mapping_table_header.tsx
chejennifer: the default sort function sorts alphabetically so the sorted order would be like [1, 100, 2, 250, 26, ..]

wizard/src/components/mapping_table_header.tsx
chejennifer: Currently, this would just end up being a duplicate column header. Some other ideas to handle this: 

a. automatically attach the column index to the header name if it's a duplicate
b. validate that all headers are unique and show an error message if there's a duplicate

wizard/src/types.ts
chejennifer: done

wizard/src/components/mapping_table_header.tsx
beets: since these are numbers, you could also do `.sort((a,b)=> a-b)`

beets (approved):

pradh (approved):
Thanks for the changes!

wizard/src/components/mapping_table_header.tsx
chejennifer: done

PR #173, opened 2022-07-19
Filter out impossible date pattern checks for performance gains

enjoythecode:
We use exceptions to determine invalid matches of date patterns in
StringUtil. This has a high CPU cost as every exception requires filling
the stack trace. This change reduces the potential number of failures by
filtering out changes that we know won't work with a length-based
heuristic.

In testing, I used open-covid-19 data available at [1]. I ran the tool
on the first 10k rows of the open_covid_19_data.csv file, with the other
.mcf and .tmcf files.

Over 3 runs before and after this commit, I saw the time average time
taken by the tool from 53.2 to 48.3 seconds. This indicates a 9% expected
performance gain.

Note that the chosen dataset has dates in the format YYYY-MM-dd, and it
does not have any invalid dates. The expected speed-up from this change
is higher when there are many invalid dates, when the date formats are
longer (since the previous approach tries roughly increasing pattern
lenghts). The expected speed-up is less if the date format is shorter,
but never negative (ie, this change will not lead to more checks being
done than the previous approach).

[1] https://github.com/datacommonsorg/data/tree/master/scripts/google_research/open-covid-19


util/src/main/java/org/datacommons/util/StringUtil.java
```diff
@@ -25,25 +25,41 @@
...
+          "yyyy-M", // 6, 7
+          "yyyy-MM-dd", // 10
+          "yyyyMMdd", // 8
+          "yyyy-M-d", // 8, 9, 10
+          "yyyy-MM-dd'T'HH:mm", // 16
+          "yyyy-MM-dd'T'HH:mm:ss", // 19
+          "yyyy-MM-dd'T'HH:mm:ss.SSS", // 23
+          "yyyy-MM-dd'T'HH:mm:ss.SSSXXX"); // 29 (`XXX` adds 6 becuase it means ""+HH:MM"")
+
+  private static final Map<Integer, List<String>> DATE_PATTERNS_LENGTH_MAP =
```
pradh, line 51: Can we compute this from DATE_PATTERNS?  (ideally at init once)

pradh (commented):
Nice.

enjoythecode:
Changed implementation to use a single `Map<String, List<Integer>>`. This both preserves the culling of unnecessary exceptions being thrown (which is the source of the performance gain) AND minimizes the amount of code needed to make this optimization happen.

util/src/main/java/org/datacommons/util/StringUtil.java
```diff
@@ -25,25 +25,34 @@
...
+  // https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatterBuilder.html#appendPattern-java.lang.String-
+  // Key is the pattern, value is a list of lengths that this pattern could
+  // potentially match. This is used in filtering out impossible date checks by
+  // comparing the length of the input date string against the Set<Integer> here.
+  private static final Map<String, Set<Integer>> DATE_PATTERNS =
+      Map.ofEntries(
+          Map.entry("yyyy", Set.of(4)),
+          Map.entry("yyyy-MM", Set.of(7)),
+          Map.entry("yyyyMM", Set.of(6)),
+          Map.entry("yyyy-M", Set.of(6, 7)),
```
pradh, line 47: Why is this 7? Won't that be handled by yyyy-MM ?

Ditto for line 50

util/src/main/java/org/datacommons/util/StringUtil.java
```diff
@@ -106,7 +115,7 @@ public static boolean isBool(String val) {
   }
 
   public static String getValidISO8601DatePattern(String dateValue) {
-    for (String pattern : DATE_PATTERNS) {
+    for (String pattern : DATE_PATTERNS.keySet()) {
```
pradh, line 118: Curious why not optimize this function? It gets called from mcf checker in the fast path too?

util/src/main/java/org/datacommons/util/StringUtil.java
```diff
@@ -124,25 +133,27 @@ public static String getValidISO8601DatePattern(String dateValue) {
...
-                .appendPattern(pattern)
-                .parseDefaulting(ChronoField.DAY_OF_MONTH, 1)
-                .parseDefaulting(ChronoField.MONTH_OF_YEAR, 1)
-                .parseDefaulting(ChronoField.HOUR_OF_DAY, 0)
-                .toFormatter(Locale.ENGLISH);
-        return LocalDateTime.parse(dateValue, dateFormat);
-      } catch (DateTimeParseException ex) {
-        // Pass through
+    for (Entry<String, Set<Integer>> entry : DATE_PATTERNS.entrySet()) {
+      if (entry.getValue().contains(dateValue.length())) {
```
pradh, line 137: From runtime perspective, I slightly prefer your prior approach of looking up a single map instead of looping over all patterns, and then doing set lookups.

I was originally mistaken that the lengths were just strlen, and thus my suggestion to "compute" it.

So, we could compute len->list-of-patterns from the map you have.  Or with lesser code, just make the Map as a List of pairs so we try YYYY first, the YYYYMM (and that's the most common).

enjoythecode: I was going for the lesser code with my revision, so I will implement the Map as a List of pairs so that we prioritize common cases and maintain the performance gain with less speed. (also would make the tests deterministic again and allow for no diffs to the tests)

Computing len>list-of-patterns map on the first call to the function and then using that map for every call, while marginally faster, involves more code as you mentioned. The optimization focus of this change is to have as few exceptions thrown as possible, so I will go with the less code option above. We can revise as needed.

util/src/main/java/org/datacommons/util/StringUtil.java
enjoythecode: You are correct; I missed this path. I will apply the same optimization and also re-run the performance tests.

util/src/main/java/org/datacommons/util/StringUtil.java
enjoythecode: replying just for reference; the patterns do not map exactly 1:1 to characters they match; ie yyyy-M might match both 2022-7 and 2022-07.

pradh: Yes, but lets make it match yyyy-MM.  That way we'll have 1 length per pattern, no?

util/src/main/java/org/datacommons/util/StringUtil.java
enjoythecode: Moved to List of pairs.

util/src/main/java/org/datacommons/util/StringUtil.java
enjoythecode: Done.

enjoythecode:
I have re-ran the speed test (3 times before and after this PR), and the new speed up is estimated at ~26%! (45.144s to 33.565s). The drastic increase in the performance is likely because I had missed the optimization in the getValidISO8601DatePattern function which is also on the fast path.

enjoythecode:
Flame graph with the new code. (compare to graph in #127; notice the absence of a "red peak" in the middle in the new flame graph below). Datasets with no incorrect dates are likely to experience this performance boost to its fullest! 

image

util/src/main/java/org/datacommons/util/StringUtil.java
pradh: Is this resolved?

enjoythecode: I missed your earlier comment.

Currently, the two patterns that are breaking 1-length-per-pattern are `yyyy-M` and `yyyy-M-d`. These are both redundant if we are not interested in matching dates of the form `2022-7`/`2022-7-5` (no leading zero in single digit month and/or day components).

I have not deleted these redundant patterns since there are test cases for such cases, but I can certainly do so if supporting those dates is no longer a requirement.

util/src/main/java/org/datacommons/util/StringUtil.java
```diff
@@ -25,25 +25,34 @@
...
-          "yyyy-MM-dd'T'HH:mm:ss",
-          "yyyy-MM-dd'T'HH:mm:ss.SSS",
-          "yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
+          // Sort these entries starting with the most common cases for runtime
+          // speed performance.
+          Map.entry("yyyy-M", Set.of(6, 7)),
+          Map.entry("yyyy-M-d", Set.of(8, 9, 10)),
+          Map.entry("yyyy", Set.of(4)),
+          Map.entry("yyyyMM", Set.of(6)),
+          Map.entry("yyyyMMdd", Set.of(8)),
```
pradh, line 50: nit: Order this as:
- yyyy
- yyyy-M
- yyyy-M-d
- yyyyMM
- yyyymmdd

So the 3 most common formats come on top.

pradh (approved):

util/src/main/java/org/datacommons/util/StringUtil.java
enjoythecode: done.

PR #174, opened 2022-07-19
list value types in summary_report.html

enjoythecode:
Builds on top of #168. The changes unique to this PR are limited to 23587094f05a1e0573215fcdf316167d62eb89fc. Once that PR is merged, this PR will merge with master and be ready for review.

## Changes
Surface value types in both in the SV summary table (list of value types associated with the SV), and in the sample places table (what is the value type of this facet?). This ensures that the values were parsed correctly now that we are allowing NaN in StatChecker.

## Examples
Non-number types can be observed in e2e tests `genmcf/measurementresult` and `genmcf/successtmcf`.

### Example from `genmcf/successtmcf` [[Live link](https://xwk34q.csb.app/)]

image
image
(see [this comment](https://github.com/datacommonsorg/import/pull/168#discussion_r924873580) about RESOLVED_REFERENCE with the `65` value)


enjoythecode:
> Perhaps we add a column only when something is a non-NUMBER? That is, if the measurement-result flag is enabled.

TODO(snny)

enjoythecode:
low priority

PR #178, opened 2022-07-22
Fix TmcfCsvParser last line number is not incremented

enjoythecode:
### What this PR Fixes

https://github.com/datacommonsorg/import/blob/d995b5c52d2750cb8ec4b5cce564209edda18e02/util/src/main/java/org/datacommons/util/TmcfCsvParser.java#L35-L36

### Implementation

Switch implementation to use a private variable because csvParser.getCurrentLineNumber() does not increment last line when the row does not end in a newline.

### Testing

Unit tests pass. I reviewed the goldens by hand to make sure that the only changes were the IDs of the last rows in the inputs. Specifically, these should increment by one since they were previously one less than the correct value.

### DIFF
--- a/util/src/main/java/org/datacommons/util/TmcfCsvParser.java
+++ b/util/src/main/java/org/datacommons/util/TmcfCsvParser.java
@@ -37,6 +37,9 @@ public class TmcfCsvParser {
 
   private Mcf.McfGraph tmcf;
   private char delimiter;
+  // We track the line number using a private variable instead of using
+  // csvparser.getCurrentLineNumber() because that function does not increment
+  // the line number for the last line that does not end in a newline.
   private long currentLineNumber;
   private String csvFileName;
   private CSVParser csvParser;
@@ -373,9 +376,6 @@ public class TmcfCsvParser {
 
   // Returns the current line in the CSV.
   private long getCurrentLineNumber() {
-    // We track the line number using a private variable instead of using
-    // csvparser.getCurrentLineNumber() because that function does not increment
-    // the line number for the last line that does not end in a newline.
     return currentLineNumber;
   }
 }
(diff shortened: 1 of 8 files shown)


util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -375,6 +368,14 @@ private String newUUID() {
...
   private void addLog(Debug.Log.Level level, String counter, String message) {
-    logCtx.addEntry(level, counter, message, csvFileName, csvParser.getCurrentLineNumber());
+    logCtx.addEntry(level, counter, message, csvFileName, getCurrentLineNumber());
+  }
+
+  // Returns the current line in the CSV.
+  private long getCurrentLineNumber() {
+    // We track the line number using a private variable instead of using
+    // csvparser.getCurrentLineNumber() because that function does not increment
+    // the line number for the last line that does not end in a newline.
```
pradh, line 378: I would move this comment to the definition of the variable.

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
```diff
@@ -375,6 +368,14 @@ private String newUUID() {
   }
 
   private void addLog(Debug.Log.Level level, String counter, String message) {
-    logCtx.addEntry(level, counter, message, csvFileName, csvParser.getCurrentLineNumber());
+    logCtx.addEntry(level, counter, message, csvFileName, getCurrentLineNumber());
+  }
+
+  // Returns the current line in the CSV.
+  private long getCurrentLineNumber() {
```
pradh, line 375: nit: up to you, but if all accesses are local, why not just do this.currentLineNumber in call sites?

pradh (approved):
Thanks for the fix!

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
enjoythecode: normally, I'd also think YAGNI, but in this case, we did just swap out the implementation detail. this felt like enough justification to have this as a point of flexibility. also, it is "what should have been", and I actually got to write this change faster because I changed this function, rather than editing the 6 locations across the file.

util/src/main/java/org/datacommons/util/TmcfCsvParser.java
enjoythecode: done.

PR #179, opened 2022-07-26
StatType support Step 3: measurementResult existence checks

enjoythecode:
builds off of #163 and #168 by performing existence checks on SVO values if their SV has `statType: measurementResult` (based on a flag).


enjoythecode:
`measurementResult` is the end-to-end test for these changes. Live link to its `summary_report.html`: https://fe5ier.csb.app/

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -185,6 +188,70 @@ public synchronized void check() {
     }
   }
 
+  // This feature is not within check() because it uses ExistenceChecker instead
+  // of using the StatChecker.check() mechanism of counter keys.
```
pradh, line 192: Hmm, actually how about calling this from check()?  It is actually checking stats (e.g., the values for measurementResult StatVarObs).

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -185,6 +188,70 @@ public synchronized void check() {
...
+                new LogCb(logCtx, Debug.Log.Level.LEVEL_ERROR, fileName, lineNumber)
+                    .setDetail(LogCb.PREF_KEY, Vocabulary.MEASUREMENT_RESULT)
+                    .setDetail(LogCb.VALUE_KEY, value)
+                    .setCounterSuffix("value_StatType_measurementResult");
+            existenceChecker.submitNodeCheck(value, logCb);
+          }
+        }
+      }
+    }
+    existenceChecker.drainRemoteCalls();
```
pradh, line 252: Add a comment why/if we need this drain?

util/src/main/java/org/datacommons/util/StatVarState.java
```diff
@@ -8,18 +13,126 @@
...
+    JsonObject statTypeJson = statTypeValuesJson.get(0).getAsJsonObject();
+
+    // StatType is a reference property, so the schema described under "Structure 2"
+    // in https://docs.datacommons.org/api/rest/property_value.html applies to us.
+    // This means that we can use the "name" field to get the value to compare against
+    // what is defined in Vocabulary.*
+    if (!statTypeJson.has("name")) {
+      return null;
+    }
+    String statType = statTypeJson.get("name").getAsString();
```
pradh, line 110: We should use dcid

util/src/main/java/org/datacommons/util/StatVarState.java
```diff
@@ -8,18 +13,126 @@
...
+      if (!typeOf.equals(Vocabulary.STAT_VAR_TYPE)) {
+        continue;
+      }
+
+      String dcid = McfUtil.getPropVal(node.getValue(), Vocabulary.DCID);
+      if (dcid.isEmpty()) {
+        continue;
+      }
+
+      String statType = McfUtil.getPropVals(node.getValue(), Vocabulary.STAT_TYPE).get(0);
```
pradh, line 130: Why not getPropVal here? also check for empty?

pradh (commented):
Nice.

util/src/main/java/org/datacommons/util/StatChecker.java
enjoythecode: The main complication I am avoiding is having to pass `args` (or `args.checkMeasurementResult`) down to StatChecker. I realize I did not communicate this in the comment. I felt that it is better if the concern of when we execute this behavior is isolated from the `StatChecker` module.

I am curious what you think about this

util/src/main/java/org/datacommons/util/StatChecker.java
enjoythecode: added something; but I am unsure if it addresses your concern. anything else that might be unclear here?

util/src/main/java/org/datacommons/util/StatChecker.java
pradh: If we designate StatChecker the responsibility of checking measurementResult type of stats, then I feel it makes sense to pass whatever is needed for the checking to StatChecker at construction.  So the caller (Processor) does not have to explicitly call into StatChecker again for just validating measurementResult.

pradh: And we do all the looping once...

enjoythecode: that makes a lot of sense, thank you for the guidance. implemented.

util/src/main/java/org/datacommons/util/StatChecker.java
enjoythecode: this is more clear now that the `checkMeasurementResult` callsite is in `StatChecker.check`. resolving unless further comments.

util/src/main/java/org/datacommons/util/StatVarState.java
```diff
@@ -127,8 +126,10 @@ public synchronized void addLocalGraph(Mcf.McfGraph graph) {
         continue;
       }
 
-      String statType = McfUtil.getPropVals(node.getValue(), Vocabulary.STAT_TYPE).get(0);
-      addStatType(dcid, statType);
+      String statType = McfUtil.getPropVal(node.getValue(), Vocabulary.STAT_TYPE);
+      if (statType != "") {
```
pradh, line 130: !statType.equals("") ?

pradh (approved):
Thanks for the changes!

util/src/main/java/org/datacommons/util/StatChecker.java
```diff
@@ -196,60 +222,54 @@ public synchronized void check() {
...
-
-          // If StatType is null, we were not able to determine the statType of SV
-          // from the local cache OR the API. Log an error and continue to next SV.
-          if (statType == null) {
-
-            List<Location> locations = new ArrayList<Location>();
-            if (timeSeries != null) {
-              locations = timeSeries.get(0).getValues(0).getLocationsList();
-            }
+    if (existenceChecker == null) {
```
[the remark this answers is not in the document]
enjoythecode: does 9073bfbb452a04ac2333b255739f97c5e2049b2d get at what you mentioned here? thanks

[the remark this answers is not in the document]
pradh: yup. thanks.

pradh (approved):

PR #185, opened 2022-08-17
Fix a couple of bugs in resolution

pradh:
1. When a node has a dcid, we should never try to resolve it. We were, due to an edge case, if that node also has an unresolved local ref (McfResolver change).
2. A counter reset bug in external-id-checker batching was disabling batching after the first batch was sent hurting performance (ExternalIdResolver change).


ajaits (approved):
Thanks for the fix!

Spaceenter (approved):

PR #186, opened 2022-08-28
Make import mvn work with JDK 16+

Spaceenter:
References:

https://github.com/google/google-java-format#as-a-library

https://github.com/spotify/fmt-maven-plugin (com.coveo has been moved to com.spotify.fmt).


README.md
```diff
@@ -51,13 +51,9 @@ Detailed documentation on the Import Tool is available [here](docs/usage.md).
    - For MacOS: `brew install maven`
 2. The tools use protobuf and require that `protoc` be installed.
    - For MacOS: `brew install protobuf`
-3. Make sure Java 11+ (but not Java 16) is installed
+3. Make sure Java 11+ is installed
```
shifucun, line 54: shall we suggest always use 16+?

shifucun (approved):
nice!

README.md
Spaceenter: We probably should support a wide range of Java versions... Maybe leave it for now...

Spaceenter:
> @Spaceenter In the past we've run into issues with the Import tool across versions. Did we validate that this doesn't break the tool?

Thanks @pradh -- Could you elaborate on what should be done to validate? Now 'mvn test' passes, and cloud build is green -- maybe I miss other necessary validations?

PR #194, opened 2022-12-14
Update usage.md

lucy-kind:
Update usage instructions with .jar file asset download link.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -16,6 +16,8 @@ The tool is a command line application built with Java. See below for usage inst
 
 # Usage
 
+Make sure you've downloaded the .jar file under Assets [here](https://github.com/datacommonsorg/import/releases/tag/0.1-alpha.1h). Note the path to .jar.
+
 Use the import tool from the command line, like so:
 
   ```bash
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -16,6 +16,8 @@ The tool is a command line application built with Java. See below for usage inst
 
 # Usage
```
juliawu, line 17: Perhaps we should move the section you just added to a new 'Installation' section.

docs/usage.md
```diff
@@ -16,6 +16,8 @@ The tool is a command line application built with Java. See below for usage inst
...
 
+Make sure you've downloaded the .jar file under Assets [here](https://github.com/datacommonsorg/import/releases/tag/0.1-alpha.1h). Note the path to .jar.
```
juliawu, line 19: Let's link to just the releases page (https://github.com/datacommonsorg/import/releases) incase we ever release newer versions. We can ask users to download the latest.

juliawu (changes requested):
Thanks for the much needed update!

docs/usage.md
lucy-kind: Done! ptal

juliawu (approved):
Thanks for the changes!

PR #198, opened 2023-03-03
Avoid calling getValueOfDataPointAsNumber non-numeric value

pradh:
If a series has a mix of numeric and non-numeric values, after https://github.com/datacommonsorg/import/pull/168 it can crash stats check, like below:

This happens when running with the WHO CSVs.

```
20:14:44.510 [main] INFO  org.datacommons.tool.Processor - Performing stats checks
java.lang.NumberFormatException: For input string: "'<0.01 [<0.01–0.01]'"
        at java.base/jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2055)
        at java.base/jdk.internal.math.FloatingDecimal.parseDouble(FloatingDecimal.java:111)
        at java.base/java.lang.Double.parseDouble(Double.java:615)
        at org.datacommons.util.PlaceSeriesSummary$SeriesSummary.getValueOfDataPointAsNumber(PlaceSeriesSummary.java:107)
        at org.datacommons.util.StatChecker.getStats(StatChecker.java:449)
        at org.datacommons.util.StatChecker.checkSigmaDivergence(StatChecker.java:416)
        at org.datacommons.util.StatChecker.check(StatChecker.java:189)
```


chejennifer (approved):
thanks for the fix!

PR #199, opened 2023-03-10
Support BOM, propagate exceptions and allow empty column names

pradh:
This PR makes 3 fixes:

1. Strips out BOM characters apparently found at the beginning of old Excel sheets
2. Propagates exceptions during CSV processing so we fail loudly
3. Allow empty column names in CSVs

The first two issues were found from Techsoup CSVs.


chejennifer (approved):
nice!

shifucun (approved):
Very subtle!!

pradh (commented):
Thanks!

PR #202, opened 2023-06-21
Only load gcs client when needed since this requires GCP credential

shifucun:


simple/util/filehandler.py
```diff
@@ -94,15 +95,15 @@ def list_files(self, extension: str = None) -> list[str]:
...
     # Using print instead of logging since the class is loaded before logging is initialized.
-    print("Using GCP Project:", gcs_client.project)
 
     def __init__(self, path: str) -> None:
         if not path.startswith(_GCS_PATH_PREFIX):
             raise ValueError(f"Expected {_GCS_PATH_PREFIX} prefix, got {path}")
         bucket_name, blob_name = path[len(_GCS_PATH_PREFIX):].split('/', 1)
-        self.bucket = GcsFileHandler.gcs_client.bucket(bucket_name)
+        gcs_client = storage.Client()
+        logging.info("Using GCP Project: %s", gcs_client.project)
```
keyurva, line 105: Remove this log otherwise it will be printed on the creation of every new file handler. Earlier, it was at class level so it was only logged once.

Also, wondering if creating a client instance on every file handler object could be expensive?

keyurva (approved):

simple/util/filehandler.py
shifucun: Ideally we want one client. I wonder if should pass the client as an argument to make_file() function?

keyurva: That will require the caller to be aware of the implementation and we want them to be agnostic.

One approach I can think of is to let it be a class level attribute but initialize it lazily only when accessed. The first instantiation will create a new client but subsequent ones will just reuse it.

shifucun: That's a good idea. Can add it in the next PR!

Issue #203, opened 2023-07-13
Support name and lat/lng based resolution in dc-import tool


pradh:
Just like how we support resolution by external ID (like [isoCode](https://github.com/datacommonsorg/import/blob/master/tool/src/test/resources/org/datacommons/tool/genmcf/resolution/input/covid.tmcf#L8-L10)) using [ExternalIdResolver](https://github.com/datacommonsorg/import/blob/master/util/src/main/java/org/datacommons/util/ExternalIdResolver.java), we would like to support resolution by:

1. `name` property, using [FindEntities recon API](https://github.com/datacommonsorg/mixer/blob/master/proto/v2/resolve.proto#L31-L33)
2. `latitude` / `longitude` property pair, using [coordinate recon API](https://github.com/datacommonsorg/mixer/blob/master/proto/v2/resolve.proto#L27-L29)

This is so that datasets that identify a location by only name or its lat/lng can be transparently resolved without the user needing to do anything special.

For `name` recon it is important to carefully propagate resolution failures though!

PR #205, opened 2023-07-22
Create resolver hierarchy and support chunked calls in recon client.

keyurva:
* For #203 
* Create a generic `Resolver` base class with concrete implementations for local, entity and coordinates resolver. 
* These resolvers are not yet integrated with `ExternalIdResolver`. Integration will be done in the next PR.
* The `ReconClient` now supports resolving both entities and coordinates. It also partitions requests into manageable chunks and issues the chunked calls in parallel.

### DIFF
--- a/util/src/main/java/org/datacommons/util/CoordinatesResolver.java
+++ b/util/src/main/java/org/datacommons/util/CoordinatesResolver.java
@@ -1,0 +1,112 @@
+package org.datacommons.util;
+
+import java.io.IOException;
+import java.util.LinkedHashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.datacommons.proto.Mcf.McfGraph.PropertyValues;
+import org.datacommons.proto.Mcf.McfGraph.Values;
+import org.datacommons.proto.Mcf.ValueType;
+import org.datacommons.proto.Recon.ResolveCoordinatesRequest;
+import org.datacommons.proto.Recon.ResolveCoordinatesRequest.Coordinate;
+import org.datacommons.proto.Recon.ResolveCoordinatesResponse;
+
+/** Resolves nodes with lat-lngs by calling the DC coordinates resolution API. */
+public class CoordinatesResolver {
+  private final AtomicBoolean resolved = new AtomicBoolean(false);
+
+  private final Set<Coordinate> resolveCoordinates = ConcurrentHashMap.newKeySet();
+
+  private final ConcurrentHashMap<Coordinate, Set<String>> resolvedCoordinates =
+      new ConcurrentHashMap<>();
+
+  private final ReconClient client;
+
+  public CoordinatesResolver(ReconClient client) {
+    this.client = client;
+  }
+
+  public void submitNode(PropertyValues node) {
+    if (resolved.get()) {
+      throw new IllegalStateException("submitNode called after remote resolution.");
+    }
+    getCoordinate(node).ifPresent(resolveCoordinates::add);
+  }
+
+  // TODO: Pick the ID based on a preferred list.
+  public String resolveNode(PropertyValues node) {
+    return getCoordinate(node)
+        .filter(resolvedCoordinates::containsKey)
+        .flatMap(coordinate -> resolvedCoordinates.get(coordinate).stream().findFirst())
+        .orElse("");
+  }
+
+  // TODO: Support chunking in batches of max size 500.
+  public void remoteResolve() throws IOException, InterruptedException {
+    if (resolved.get()) {
+      throw new IllegalStateException("remoteResolve called after remote resolution.");
+    }
+    resolved.set(true);
+
+    if (resolveCoordinates.isEmpty()) {
+      return;
+    }
+
+    ResolveCoordinatesRequest request =
+        ResolveCoordinatesRequest.newBuilder().addAllCoordinates(resolveCoordinates).build();
+    ResolveCoordinatesResponse response = client.resolveCoordinates(request);
+    populateResolvedCandidates(response);
+  }
+
+  boolean isResolved() {
+    return resolved.get();
+  }
+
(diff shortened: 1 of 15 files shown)


util/src/main/java/org/datacommons/util/CoordinatesResolver.java
```diff
@@ -0,0 +1,112 @@
...
+    if (node.containsPvs(Vocabulary.LATITUDE) && node.containsPvs(Vocabulary.LONGITUDE)) {
+
+      Optional<Double> optLat = getDoubleValue(node.getPvsMap().get(Vocabulary.LATITUDE));
+      Optional<Double> optLng = getDoubleValue(node.getPvsMap().get(Vocabulary.LONGITUDE));
+
+      if (optLat.isPresent() && optLng.isPresent()) {
+        double lat = optLat.get();
+        double lng = optLng.get();
+
+        if (!Double.isNaN(lat) && !Double.isNaN(lng)) {
```
pradh, line 93: We should do some sanity checks on the lat/lng?  Add TODO?

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
```diff
@@ -0,0 +1,112 @@
...
+
+  public void submitNode(PropertyValues node) {
+    if (resolved.get()) {
+      throw new IllegalStateException("submitNode called after remote resolution.");
+    }
+    getCoordinate(node).ifPresent(resolveCoordinates::add);
+  }
+
+  // TODO: Pick the ID based on a preferred list.
+  public String resolveNode(PropertyValues node) {
```
pradh, line 39: Not fully clear what resolveNode does, maybe document public API usage?

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+    resolved.set(true);
+
+    if (resolveCoordinates.isEmpty()) {
+      return;
+    }
+
+    ResolveCoordinatesRequest request =
+        ResolveCoordinatesRequest.newBuilder().addAllCoordinates(resolveCoordinates).build();
+    ResolveCoordinatesResponse response = client.resolveCoordinates(request);
```
pradh, line 59: Where are we making the call?

Also, should we limit the number of things added?

pradh (commented):
Thanks!

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
keyurva: There's a TODO on the method below (`getDoubleValue`) to support other formats.

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
keyurva: I've added comments on the class. The idea is that the resolver will work in 3 stages - submitNode, execute and getResolvedNode. The latter 2 were called remoteResolve and resolveNode in the earlier commit.

Once this PR is submitted, I'll create a `Resolver` base class and have different types of resolvers (entities, coordinates and others like name in the future) as concrete resolvers. The `ExternalIdResolver` will then delegate to these.

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
keyurva: I was planning on adding chunking in the next PR but have done it in this PR itself. PTAL.

util/src/main/java/org/datacommons/util/ReconClient.java
```diff
@@ -22,30 +27,99 @@ public class ReconClient {
...
+      return CompletableFuture.completedFuture(defaultResponse);
+    }
+
+    return toFutureOfList(
+            partition(request.getEntitiesList(), chunkSize).stream()
+                .map(chunk -> request.toBuilder().clearEntities().addAllEntities(chunk).build())
+                .map(
+                    chunkedRequest ->
+                        callApi(RESOLVE_ENTITIES_API_URL, chunkedRequest, defaultResponse))
+                .collect(toList()))
```
pradh, line 94: I'm a bit unsure about the logic here.  Partly my ignorance in having never dealt with futures, and partly because its unclear to me whether doing doing this sort of async thing is warranted?   Worth also noting that these functions will be called concurrently, and might need to reason about that too.

So generally I wonder why not keep the logic addition separate from optimizations that change the threading/synchrony?  Get the basic lat/lng thing working e2e, measure perf etc, then do the refactors / optimizations.

util/src/main/java/org/datacommons/util/EntityResolver.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import org.datacommons.proto.Mcf.McfGraph.PropertyValues;
+import org.datacommons.proto.Recon.IdWithProperty;
+import org.datacommons.proto.Recon.ResolveEntitiesRequest;
+import org.datacommons.proto.Recon.ResolveEntitiesResponse;
+import org.datacommons.proto.Recon.ResolveEntitiesResponse.ResolvedEntity;
+import org.datacommons.proto.Recon.ResolveEntitiesResponse.ResolvedId;
+
+/** Resolves nodes by calling the DC entity resolution API. */
+final class EntityResolver extends Resolver {
```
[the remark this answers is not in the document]
keyurva: `ExternalIdResolver` broadly speaking does 4 things - remote entity resolution, local resolution, batching and calling DC. `EntityResolver` only does the first thing in the list. Leads to separation of concerns and modular design.

util/src/main/java/org/datacommons/util/ReconClient.java
keyurva: It's not about optimization but simplicity. What makes concurrent programming complicated is not parallelism but mutation. Mutation requires management - mutexes, locks, mutual exclusion. Here we are dealing with protos which are immutable and hence it is inherently concurrent. It just turns out that this simplicity also enables parallelism.

Anyway, I'm fine going with whichever approach you feel is the right thing to do. Let's discuss in office and go with that.

PR #209, opened 2023-08-10
Add support for coordinates resolution.

keyurva:
* For #203 
* Coordinates resolution is only enabled if a new `--coordinates-resolution` flag is set.
* The implementation is intentionally minimally invasive to the existing resolution code path and is executed only when the above flag is set.
* Currently it only supports numeric `latitude` and `longitude` properties and basic validation. Support for more formats and validations will be added subsequently.

### DIFF
--- a/util/src/main/java/org/datacommons/util/CoordinatesResolver.java
+++ b/util/src/main/java/org/datacommons/util/CoordinatesResolver.java
@@ -1,0 +1,96 @@
+package org.datacommons.util;
+
+import static java.util.stream.Collectors.toList;
+
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import org.datacommons.proto.Mcf.McfGraph.PropertyValues;
+import org.datacommons.proto.Mcf.McfGraph.Values;
+import org.datacommons.proto.Mcf.ValueType;
+import org.datacommons.proto.Recon.ResolveCoordinatesRequest;
+import org.datacommons.proto.Recon.ResolveCoordinatesRequest.Coordinate;
+import org.datacommons.proto.Recon.ResolveCoordinatesResponse;
+import org.datacommons.proto.Recon.ResolveCoordinatesResponse.Place;
+
+/** Resolves nodes with lat-lngs by calling the DC coordinates resolution API. */
+final class CoordinatesResolver {
+  private final Set<Coordinate> resolveCoordinates = ConcurrentHashMap.newKeySet();
+
+  private final ConcurrentHashMap<Coordinate, Set<String>> resolvedCoordinates =
+      new ConcurrentHashMap<>();
+
+  private final ReconClient client;
+
+  CoordinatesResolver(ReconClient client) {
+    this.client = client;
+  }
+
+  boolean submit(PropertyValues node) {
+    return getCoordinate(node).map(resolveCoordinates::add).map(unused -> true).orElse(false);
+  }
+
+  void drain() {
+    if (!resolveCoordinates.isEmpty()) {
+      populateResolvedCandidates(
+          client.resolveCoordinates(
+              ResolveCoordinatesRequest.newBuilder()
+                  .addAllCoordinates(resolveCoordinates)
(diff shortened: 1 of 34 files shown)


util/src/main/java/org/datacommons/util/ReconClient.java
```diff
@@ -1,51 +1,130 @@
...
- * <p>Currently it only resolves coordinates.
+ * <p>Currently it only resolves coordinates. If the number of coordinates to resolve are greater
+ * than {@code chunkSize}, the API calls will be partitioned into max {@code chunkSize}d batches.
  */
 public class ReconClient {
   private static final String RESOLVE_COORDINATES_API_URL =
       "https://api.datacommons.org/v1/recon/resolve/coordinate";
 
+  private static final String RESOLVE_ENTITIES_API_URL =
+      "https://api.datacommons.org/v1/recon/entity/resolve";
```
pradh, line 33: Is this used?

util/src/main/java/org/datacommons/util/ReconClient.java
```diff
@@ -1,51 +1,130 @@
...
+        .thenApply(
+            chunkedResponses ->
+                ResolveCoordinatesResponse.newBuilder()
+                    .addAllPlaceCoordinates(
+                        chunkedResponses.stream()
+                            .flatMap(
+                                chunkedResponse ->
+                                    chunkedResponse.getPlaceCoordinatesList().stream())
+                            .collect(toList()))
+                    .build());
```
pradh, line 88: please add some comments for this block?

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import org.datacommons.proto.Mcf.McfGraph.PropertyValues;
+import org.datacommons.proto.Mcf.McfGraph.Values;
+import org.datacommons.proto.Mcf.ValueType;
+import org.datacommons.proto.Recon.ResolveCoordinatesRequest;
+import org.datacommons.proto.Recon.ResolveCoordinatesRequest.Coordinate;
+import org.datacommons.proto.Recon.ResolveCoordinatesResponse;
+import org.datacommons.proto.Recon.ResolveCoordinatesResponse.Place;
+
+/** Resolves nodes with lat-lngs by calling the DC coordinates resolution API. */
```
pradh, line 15: Add TODO to update counters on errors?

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
```diff
@@ -0,0 +1,96 @@
...
+      new ConcurrentHashMap<>();
+
+  private final ReconClient client;
+
+  CoordinatesResolver(ReconClient client) {
+    this.client = client;
+  }
+
+  boolean submit(PropertyValues node) {
+    return getCoordinate(node).map(resolveCoordinates::add).map(unused -> true).orElse(false);
```
pradh, line 29: nit: This looks like a slightly convoluted syntax to add to a set?

pradh (approved):
Nice!

Please consider updating docs in a follow up.

https://github.com/datacommonsorg/import/tree/master/docs

util/src/main/java/org/datacommons/util/ReconClient.java
keyurva: Good catch. Removed.

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
keyurva: Done.

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
keyurva: Updated. I typically don't like to use get() on Optional but it makes sense to do here for readability. Thanks for calling it out.

PR #210, opened 2023-08-24
Add support for calling the v2 resolve API from the import tool.

keyurva:
* Calling this API is needed for adding support for name resolution.
* Coordinates resolver (and potentially, id resolver) will be switched to using this API in subsequent PRs.


util/src/main/java/org/datacommons/util/ReconClient.java
```diff
@@ -49,6 +56,42 @@ public ReconClient(HttpClient httpClient, LogWrapper logWrapper, int chunkSize)
...
+  }
+
+  public CompletableFuture<ResolveResponse> resolveAsync(ResolveRequest request) {
+    ResolveResponse defaultResponse = ResolveResponse.getDefaultInstance();
+    if (request.getNodesCount() < 1) {
+      return CompletableFuture.completedFuture(defaultResponse);
+    }
+
+    return toFutureOfList(
+            // Partition request into chunkSize batches.
```
Spaceenter, line 74: Curious, is this a parallel call for different chunks?

Spaceenter (approved):

util/src/main/java/org/datacommons/util/ReconClient.java
keyurva: Yes it is. It roughly works like this:

```
Request -> partition -> 
List<Request> -> call DC in parallel for each (chunked) Request -> List<Future<Response>> -> 
Future<List<Response>> -> join -> Future<Response>
```

I'll perhaps do a PR to capture these states into named variables so it's easier to follow.

PR #211, opened 2023-08-25
Use V2Resolve API for CoordinatesResolver.

keyurva:
* Note that the V2 API returns candidates in a different order than the V1 API, hence the outputs in some of the tests have been changed.
* Removed V1 API calls from `ReconClient`.
* Made various steps in `ReconClient` more explicit with comments and named variables so the logic is easier to follow.


util/src/main/java/org/datacommons/util/CoordinatesResolver.java
```diff
@@ -37,11 +43,15 @@ boolean submit(PropertyValues node) {
...
-          client.resolveCoordinates(
-              ResolveCoordinatesRequest.newBuilder()
-                  .addAllCoordinates(resolveCoordinates)
-                  .build()));
+      ResolveRequest request =
+          ResolveRequest.newBuilder()
+              .addAllNodes(resolveCoordinates)
+              .setProperty("<-geoCoordinate->dcid")
+              .build();
+      System.out.println(request);
```
Spaceenter, line 51: nit: This seems to be a debugging line during development, shall we remove it?

util/src/main/java/org/datacommons/util/CoordinatesResolver.java
```diff
@@ -37,11 +43,15 @@ boolean submit(PropertyValues node) {
...
+      ResolveResponse response = client.resolve(request);
+      System.out.println(response);
```
Spaceenter, line 53: nit: This seems to be a debugging line, shall we remove it?

Spaceenter (approved):

PR #213, opened 2023-09-15
Add resolver to resolve nodes with names.

keyurva:
* Will be integrated with the tool in the next PR.

### DIFF
--- a/util/src/main/java/org/datacommons/util/NameResolver.java
+++ b/util/src/main/java/org/datacommons/util/NameResolver.java
@@ -1,0 +1,92 @@
+package org.datacommons.util;
+
+import static java.util.stream.Collectors.toList;
+
+import java.util.LinkedHashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import org.datacommons.proto.Mcf.McfGraph.PropertyValues;
+import org.datacommons.proto.Mcf.McfGraph.TypedValue;
+import org.datacommons.proto.Mcf.McfGraph.Values;
+import org.datacommons.proto.Mcf.ValueType;
+import org.datacommons.proto.Resolve.ResolveRequest;
+import org.datacommons.proto.Resolve.ResolveResponse;
+import org.datacommons.proto.Resolve.ResolveResponse.Entity.Candidate;
+
+/** Resolves nodes with names by calling the DC resolution API. */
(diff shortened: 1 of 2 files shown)


util/src/main/java/org/datacommons/util/NameResolver.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+    return false;
+  }
+
+  void drain() {
+    if (!resolveNames.isEmpty()) {
+      ResolveRequest request =
+          ResolveRequest.newBuilder()
+              .addAllNodes(resolveNames)
+              .setProperty("<-description->dcid")
```
shifucun, line 46: make "<-description->dcid" constant?

shifucun, line 46: Resolve based on type would be very useful. https://github.com/datacommonsorg/mixer/blob/master/internal/server/v2/resolve/golden/resolve_description_test.go#L50

Maybe add in follow up.

util/src/main/java/org/datacommons/util/NameResolver.java
```diff
@@ -0,0 +1,92 @@
...
+              .build();
+      ResolveResponse response = client.resolve(request);
+      populateResolvedNames(response);
+    }
+  }
+
+  Optional<String> resolve(PropertyValues node) {
+    return getName(node)
+        .filter(resolvedNames::containsKey)
+        .flatMap(name -> resolvedNames.get(name).stream().findFirst());
```
shifucun, line 56: Sometimes, many resolved candidates are returned, not sure if we always want to pick first one, or return these to clients

util/src/main/java/org/datacommons/util/NameResolver.java
keyurva: Done.

keyurva: Good point. Added TODO.

util/src/main/java/org/datacommons/util/NameResolver.java
keyurva: This will be called by `ExternalIdResolver` which needs only one value to be returned. But I think what we can do is allow clients to optionally set what type of value they want and resolve based on that. I've added a TODO.

shifucun (approved):

PR #214, opened 2023-09-18
Create RSI importer for simple entities.

keyurva:


simple/entities/main.py
```diff
@@ -0,0 +1,147 @@
...
+    OUTPUT_DIR = os.path.join(DATA_DIR, "output")
+
+
+flags.DEFINE_enum("entity_type", None, ["Country", "City"],
+                  "The type of entities in the CSV.")
+flags.DEFINE_string("input_dir", Defaults.INPUT_DIR, "The input directory.")
+flags.DEFINE_string("output_dir", Defaults.OUTPUT_DIR, "The output directory.")
+
+
+class Const:
```
shifucun, line 44: Instead using Const class, may just make these as module variable?

simple/entities/main.py
```diff
@@ -0,0 +1,147 @@
...
+# For importing util
+_CODEDIR = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(1, os.path.join(_CODEDIR, "../"))
+
+from util import dcclient as dc
+
+FLAGS = flags.FLAGS
+
+
+class Defaults:
```
shifucun, line 32: Instead using Const class, may just make these as module variable?

simple/entities/sample/observations.csv
```diff
@@ -0,0 +1,21 @@
...
+country/AGO,var2,2023,6.0
+country/ZMB,var2,2023,34.0
+country/ZWE,var2,2023,76.0
+country/ALB,var2,2023,34.0
+country/DZA,var2,2023,92.0
+country/AND,var2,2023,9.0
+country/ASM,var2,2023,34.0
+country/AIA,var2,2023,42.0
+country/WLF,var2,2023,75.0
+country/ESH,var2,2023,65.0
```
shifucun, line 21: why are there more rows than input?

keyurva: This is because the input CSVs are "wide" (i.e. one column per SV) where the output CSV is "long" (i.e. one row per SV value).

shifucun: I see, there are two var in the input

simple/entities/sample/observations.csv
```diff
@@ -0,0 +1,21 @@
+entity,variable,date,value
```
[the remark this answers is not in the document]
keyurva: Renamed to `dcid`. Can change name again later if needed.

simple/entities/main.py
```diff
@@ -0,0 +1,147 @@
...
+    COLUMN_DATE = "date"
+    COLUMN_VALUE = "value"
+    OBSERVATIONS_FILE_NAME = "observations.csv"
+    REPORT_ITEMS = "items"
+    REPORT_ITEM_TITLE = "title"
+    REPORT_ITEM_DETAILS = "details"
+    REPORT_FILE_NAME = "report.json"
+
+
+class SimpleEntitiesImporter:
```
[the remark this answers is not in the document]
keyurva: Thanks! BTW, moved this to a separate file.

simple/entities/main.py
keyurva: Done.

simple/entities/main.py
```diff
@@ -0,0 +1,147 @@
...
+from absl import app
+from absl import logging
+import pandas as pd
+import json
+
+# For importing util
+_CODEDIR = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(1, os.path.join(_CODEDIR, "../"))
+
+from util import dcclient as dc
```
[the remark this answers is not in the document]
keyurva: Done.

simple/entities/main.py
keyurva: Done.

simple/entities/README.md
```diff
@@ -0,0 +1,25 @@
+# Simple Entities Importer
```
pradh, line 1: Add some more user-oriented documentation?  Like what is the context of this tool?

pradh, line 1: Also, why is it called entities importer?  This is Simple Observations Importer really?

simple/entities/constants.py
```diff
@@ -0,0 +1,28 @@
...
+DEFAULT_INPUT_DIR = os.path.join(DEFAULT_DATA_DIR, "input")
+DEFAULT_OUTPUT_DIR = os.path.join(DEFAULT_DATA_DIR, "output")
+
+OBSERVATIONS_FILE_NAME = "observations.csv"
+
+# Observations CSV columns.
+COLUMN_DCID = "dcid"
+COLUMN_VARIABLE = "variable"
+COLUMN_DATE = "date"
+COLUMN_VALUE = "value"
```
pradh, line 28: nit: if this is a generated CSV, match the column name to the property name, aka:

observationAbout, variableMeasured, observationDate, value

simple/entities/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Outputs `observations.csv` for simple entities to be used by [RSI sqlite][sqlite].
+
+[sqlite]: https://github.com/datacommonsorg/mixer/tree/a768446c56095aa23add8c59cf6a0630a17a726b/internal/sqlite
+
+## Usage
```
pradh, line 7: How are we dealing with unit, have a TODO maybe?

simple/entities/main.py
```diff
@@ -0,0 +1,48 @@
...
+
+# For importing util
+_CODEDIR = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(1, os.path.join(_CODEDIR, "../"))
+
+from util import dc_client as dc
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_enum("entity_type", None, ["Country", "City"],
```
pradh, line 30: Should we make this the first column name?  Might be nice to have the CSV completely self-contained, no?

CC @shifucun

simple/entities/constants.py
keyurva: I like this idea but will require changes to the mixer code. @shifucun?

simple/entities/README.md
keyurva: Eventually, we might have more importers for different types of data and they will all be producing observations. This one is called entities importer since it resolves entities using the resolution APIs. I'm open to other names though.

simple/entities/README.md
keyurva: Added TODO.

The original RSI doc talks about specifying them at the column or cell level. Another approach could be to allow separate columns for different properties. Let's decide on the approach and then add support in here.

simple/entities/constants.py
shifucun: We do need a bunch of change for consistency; For now, maybe leave as is. And when function is complete, we can do a one-time update?

keyurva: SG. Will add a TODO.

Also, for the entity/dcid column, should we call it `dcid` or leave it as `entity` for now?

simple/entities/main.py
shifucun: Not sure if we want user to enter type schema, either as command line or as a column. Even I don't know what the type for zip code is exactly :). 

I feel the user flow is still an open ended question. And later on, maybe just have a script to focus on resolving! Otherwise, this folder will becoming more and more complicated and can only be used by us.

shifucun (approved):
I think there are still a bunch of open questions. I will LG for now to unblock the FLORET work.

simple/entities/README.md
pradh: Hmmm, the reason I wonder is, this is importing "observations", whereas, importing entities (like triples in the graph like food banks) is what we've been calling "entities" import.

simple/entities/main.py
pradh: SG

pradh (approved):
LG

simple/entities/README.md
keyurva: Renamed to `SimpleStatsImporter` after an offline chat.

keyurva:
Thanks guys. Let's learn from the floret import process and address the open questions over time.

PR #219, opened 2023-09-23
Output csvs in wide format by default.

keyurva:
* Outputs csvs in wide format by default (i.e. without "unpivotting").
* Allows overriding API root via env variable (e.g. to call autopush api).
* Supports debug flag (in code) to write entities to file for debugging.

### DIFF
--- a/simple/util/dc_client.py
+++ b/simple/util/dc_client.py
@@ -17,25 +17,41 @@
 import os
 import requests
 from absl import logging
+import json
 
 from .ngram_matcher import NgramMatcher
 
-# Environment variable for API key.
+# Environment variables.
 _KEY_ENV = "DC_API_KEY"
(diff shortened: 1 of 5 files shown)


simple/util/dc_client.py
```diff
@@ -17,26 +17,42 @@
...
 
 
-# REST API endpoint root
-_API_ROOT = "https://api.datacommons.org"
+def get_api_root():
+    return os.environ.get(_API_ROOT_ENV, _DEFAULT_API_ROOT)
+
+
+_DEBUG = True
+_DEBUG_FOLDER = ".data/debug"
```
shifucun, line 41: nit: move the constants to the top together.

shifucun (approved):

simple/util/dc_client.py
keyurva: Done.

PR #221, opened 2023-10-05
Add support for resolving lat lngs.

keyurva:
* Also added support for geoIds.
* The date field is now always output with the name `date`.


simple/stats/importer.py
```diff
@@ -136,22 +116,41 @@ def remove_pre_resolved(entity: str) -> bool:
             unresolved=unresolved_list,
         )
 
+    def _resolve(self, entity_column_name: str,
+                 entities: list[str]) -> dict[str, str]:
+        if entity_column_name.casefold(
+        ) == constants.INPUT_COLUMN_DCID.casefold():
```
shifucun, line 122: very nit, can use a function like check_column(column_name, constatns.xxx) and use casefold() in the function.

simple/stats/sample/countries/observations.csv
```diff
@@ -1,23 +1,14 @@
...
-country/ZWE,var2,2023,76.0
-country/ALB,var2,2023,34.0
-wikidataId/Q22062741,var2,2023,97.0
-country/DZA,var2,2023,92.0
-country/AND,var2,2023,9.0
-country/ASM,var2,2023,34.0
-country/AIA,var2,2023,42.0
-country/WLF,var2,2023,75.0
-country/ESH,var2,2023,65.0
+dcid,date,var1,var2
```
shifucun, line 1: thanks for the update!

simple/stats/constants.py
```diff
@@ -22,9 +22,15 @@
 OBSERVATIONS_FILE_NAME = "observations.csv"
 DEBUG_RESOLVE_FILE_NAME = "debug_resolve.csv"
 
-DCID_OVERRIDE_PREFIX = "dcid:"
+# Input columns.
+INPUT_COLUMN_DCID = "dcid"
+INPUT_COLUMN_GEO_ID = "geoId"
```
shifucun, line 27: could the geoId be even more generalized here? like we can do "wikidataId" now.

shifucun (approved):
very nice to make the input general!

simple/stats/importer.py
keyurva: I've made the logic more generic. PTAL.

simple/stats/constants.py
keyurva: Yes, I've made it more generic. PTAL.

simple/stats/constants.py
```diff
@@ -26,6 +26,22 @@
...
+
+e.g. if a column is named "geoId" and the values of that column in the CSV are: 01, 02; 
+then the importer will update the values to geoId/01, geoId/02 before importing them.
+
+Note that the lookup into these columns will always be lower-cased so ensure that 
+the keys in the dictionary are lower-case as well.
+
+Also, keep the keys sorted so it's easier to spot check should the list get large.
+"""
+PRE_RESOLVED_INPUT_COLUMNS_TO_PREFIXES = dict([("dcid", ""),
```
shifucun, line 41: for wikidataId, we still want to resolve them to be dcid using the resolve api: 
image

PR #222, opened 2023-10-06
Resolve wikidataId externally.

keyurva:


shifucun (approved):
This is great and really paves the way to resolve any property!

PR #223, opened 2023-10-07
Add support for GCS paths.

keyurva:
* Created a generic `FileHandler` abstraction with local and GCS implementations.
* GCS paths must begin with `gs://`.
* Example of GCS usage:
   + `python3 main.py --input_path=gs://datcom-floret-dev-resources/data/observations-floretcountrydata.csv --output_dir=gs://datcom-floret-dev-resources/temp/`
   + The above command read the CSV from the GCS path and output the files to another GCS path [here](https://screenshot.googleplex.com/8DZktSQaTxV2QNw).
* A mix of local and GCS paths for inputs / outputs is also supported.
   + e.g. `python3 main.py --input_path=gs://datcom-floret-dev-resources/data/observations-floretcountrydata.csv --output_dir=data/`
* Note that input folders are not yet supported. That will be implemented soon.

### DIFF
--- a/simple/util/filehandler.py
+++ b/simple/util/filehandler.py
@@ -1,0 +1,98 @@
+# Copyright 2023 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""A generic FileHandler abstraction that allows clients to work seamlessly with 
+local and GCS files and directories.
+"""
+
+import os
+import io
+import logging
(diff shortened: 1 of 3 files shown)


simple/util/filehandler.py
```diff
@@ -0,0 +1,98 @@
...
+
+import os
+import io
+import logging
+from google.cloud import storage
+
+_GCS_PATH_PREFIX = "gs://"
+
+
+class FileHandler:
```
shifucun, line 26: Add comment this is a base class to be extended.

shifucun (approved):
very clean and nice! Maybe in a follow up, can you output observations and debug info in separate folders, then we can start to test e2e!

simple/util/filehandler.py
keyurva: Done.

keyurva:
> output observations and debug info in separate folders

I'm wondering which one is better: managing multiple folders vs all in the same folder and use filename conventions (`observations*` for observations, `debug*` for debug, `status*` for status, etc.).

shifucun:
> > output observations and debug info in separate folders
> 
> I'm wondering which one is better: managing multiple folders vs all in the same folder and use filename conventions (`observations*` for observations, `debug*` for debug, `status*` for status, etc.).

multiple folders can make it easy for manual debugging, especially when there are a large number of files.

PR #227, opened 2023-10-10
Lazy load GCS client.

keyurva:


shifucun (approved):
This is very cool! Can you help verify this in the docker setup? Basically can make this change under website/import submodule, build the local docker image and run it.

PR #230, opened 2023-10-16
Generate an initial report.json when running the importer.

keyurva:
* Reports status on overall and individual file imports.
* Support for reporting more details will be added in subsequent PRs.

### DIFF
--- a/simple/stats/reporter.py
+++ b/simple/stats/reporter.py
@@ -48,6 +48,8 @@ class ImportReporter:
     self.data = {}
     self.import_files: dict[str, FileImportReporter] = {}
 
+  # Functions decorated with @_report will result in the report being saved
+  # upon function execution.
   def _report(func):
 
     @wraps(func)
@@ -132,6 +134,8 @@ class FileImportReporter:
     self.parent = parent
     self.data = {}
 
+  # Functions decorated with @_report will result in this report being reported to the parent reporter
+  # upon function execution.
   def _report(func):
 
     @wraps(func)
--- a/simple/stats/reporter.py
+++ b/simple/stats/reporter.py
@@ -1,0 +1,93 @@
+# Copyright 2023 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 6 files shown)


simple/stats/reporter.py
```diff
@@ -0,0 +1,178 @@
...
+from enum import Enum
+from functools import wraps
+import json
+import os
+import sys
+
+# For importing util
+_CODEDIR = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(1, os.path.join(_CODEDIR, "../"))
+
```
shifucun, line 26: I wonder should we start to add "__init__.py" and have module structure and not using sys path operation like this.

Then you can run main.py as module like "python3 -m simple.stats.main"

simple/stats/reporter.py
```diff
@@ -0,0 +1,178 @@
...
+  def __init__(self, report_fh: FileHandler) -> None:
+    self.status = Status.NOT_STARTED
+    self.start_time = None
+    self.last_update = datetime.now()
+    self.report_fh = report_fh
+    self.data = {}
+    self.import_files: dict[str, FileImportReporter] = {}
+
+  def _report(func):
```
shifucun, line 51: Add a comment on the decorator function?

simple/stats/runner.py
```diff
@@ -44,6 +46,8 @@ def __init__(
     self.output_dir_fh = create_file_handler(output_dir)
     self.process_dir_fh = self.output_dir_fh.make_file(
         f"{constants.PROCESS_DIR_NAME}/")
+    self.reporter = ImportReporter(report_fh=self.process_dir_fh.make_file(
+        constants.REPORT_JSON_FILE_NAME))
```
shifucun, line 50: Shall we name the report file with timestamp is it's not overwritten?

shifucun (approved):
This is fantastic!

simple/stats/reporter.py
keyurva: That's a good idea. Let me do this in a separate PR.

simple/stats/reporter.py
keyurva: Done.

simple/stats/runner.py
keyurva: I'm thinking since the mixer load endpoint will call the script with a new output dir on every execution, we won't run into the overwrite problem?

PR #232, opened 2023-10-18
Add initial support for generating triples.

keyurva:
* Triples are saved in a `triples.csv` in the output directory.
* Currently it only generates SV triples based on the SV column names.
   + SV descriptions and hierarchy support will be added next.
* Also, made `stats` into a module to eliminate the need for funky imports.


simple/sample/output/triples.csv
```diff
@@ -0,0 +1,9 @@
+subject_id,predicate,object_id,object_value
```
shifucun, line 1: Mixer now can only read from one "raw data" folder, so the triples csv would ideally be in the same folder as the observations.

shifucun (commented):
Very nice clean up on the module and folder structure!

simple/sample/output/triples.csv
keyurva: Triples csv is in the same folder as observations. Basically, the same folder where you read the observations to load the sqlite db.

shifucun: Wonder why the triples.csv here is not in the same folder of observations?

shifucun (approved):

simple/sample/output/triples.csv
keyurva: It is in the same folder: https://screenshot.googleplex.com/8G86oAmfkmgU4ht

If you're implying a different folder, let me know.

PR #235, opened 2023-10-21
Add support for SV hierarchy.

keyurva:
* Added support for SV hierarchy (groups), description, name overrides and NL sentences.
  + Updated config doc describing these.
* Added initial unit tests.
* Added options to `run_test.sh`.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -31,4 +31,64 @@ function run_lint_fix {
   deactivate
 }
 
-run_lint_fix
+# Fixes python tests
+function run_py_test {
+  python3 -m venv .env
+  source .env/bin/activate
+  
+  cd simple
+  pip3 install -r requirements.txt
+
+  echo -e "#### Running stats tests"
+  python3 -m pytest tests/stats/ -s
+
+  deactivate
+}
+
+function run_all_tests {
+  run_lint_fix
+  run_py_test
+}
+
+function help {
+  echo "Usage: $0 -afhp"
+  echo "-a              Run all tests"
+  echo "-f              Fix lint"
+  echo "-h              This usage"
+  echo "-p              Run python tests"
+  exit 1
+}
+
+if [ $# -eq 0 ]; then
+  help
+  exit 1
+fi
+
+# Always reset the variable null.
+while [[ "$#" -gt 0 ]]; do
+  case "$1" in
+    -a)
+        echo -e "### Running all tests"
+        run_all_tests
+        shift 1
+        ;;
+    -f)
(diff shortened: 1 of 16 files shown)


simple/sample/input/config.json
```diff
@@ -19,5 +19,23 @@
...
     }
+  },
+  "variables": {
+    "var1": {
+      "name": "Good var1 name",
+      "description": "Good var1 description",
+      "nlSentences": [
+        "Natural language sentence 1",
+        "Natural language sentence 2"
+      ],
```
shifucun, line 30: I still think separate out the nl descriptions into a separate csv file is cleaner.

simple/stats/config.md
```diff
@@ -17,6 +17,16 @@ The config parameters for the files to be imported should be specified in a `con
     "geoid.csv": {
       "entityType": ""
     }
+  },
+  "variables": {
```
shifucun, line 21: Can we also handle the case when variable does not have "group" field, then it is linked under "dc/g/Root" directly.

run_test.sh
```diff
@@ -31,4 +31,64 @@ function run_lint_fix {
...
+  pip3 install -r requirements.txt
+
+  echo -e "#### Running stats tests"
+  python3 -m pytest tests/stats/ -s
+
+  deactivate
+}
+
+function run_all_tests {
+  run_lint_fix
```
shifucun, line 49: I realized the run_list_fix should not be run in cloudbuild.yaml and used in here. This option is meant to be run locally to fix the lint issues.

We need another `run_list_test` option to replace run_list_fix in cloudbuild.yaml and here that executes commands like https://github.com/datacommonsorg/website/blob/master/run_test.sh#L108-L122

run_test.sh
```diff
@@ -31,4 +31,64 @@ function run_lint_fix {
   deactivate
 }
 
-run_lint_fix
\ No newline at end of file
+# Fixes python tests
+function run_py_test {
```
shifucun, line 35: thanks for adding py test, this is very good start!

simple/sample/input/config.json
keyurva: Discussed offline to keep it here for now. Will consider moving to a different file in the future if it's helpful.

simple/stats/config.md
keyurva: Done.

run_test.sh
keyurva: Done.

shifucun (approved):
thanks for the update!

PR #236, opened 2023-10-23
Rename SV column names to SV ids.

keyurva:
* Will set sv id == column name for id compatible column names in the next PR.


simple/sample/output/process/debug_resolve_powerplants.csv
```diff
@@ -1,6 +1,6 @@
 input,dcid,link
-FOO BAR,*UNRESOLVED*,
```
shifucun, line 2: this file seems to change every time, is there a way to make it stable?

simple/sample/output/process/report.json
```diff
@@ -1,37 +1,37 @@
 {
   "status": "SUCCESS",
-  "startTime": "2023-10-23 10:50:22.644858",
-  "lastUpdate": "2023-10-23 10:50:30.534262",
+  "startTime": "2023-10-23 14:43:10.816166",
```
shifucun, line 3: shall we gitignore this file?

simple/stats/nodes.py
```diff
@@ -0,0 +1,85 @@
...
+    self.variables[sv_column_name] = StatVar(
+        f"{_CUSTOM_SV_ID_PREFIX}{len(self.variables) + 1}",
+        var_cfg.name,
+        description=var_cfg.description,
+        nl_sentences=var_cfg.nl_sentences,
+        group_id=group_id,
+    )
+    return self.variables[sv_column_name]
+
+  def group(self, group_path: str) -> StatVarGroup | None:
```
shifucun, line 55: just thinking ahead, if we use column name as dcid, then the group path might to have the stat var name. for exampele:

"country_risk": "Floret/Country_Risk"

shifucun (commented):
nice simplification!

simple/sample/output/process/debug_resolve_powerplants.csv
keyurva: Yeah it's annoying. Will look into it after the next PR.

simple/sample/output/process/report.json
keyurva: Let's keep it around for now until we have reached a steady state.

One way we can address it is to have a mode that persists it with fixed timestamps for samples and tests.

simple/stats/nodes.py
keyurva: Group paths don't have stat var names. In the above case, the group name will only be "Floret" and the group ID is generated by us (e.g. "custom/g/group_1")

shifucun: Oh, now i remember, there is a separate name field for the stat var.

shifucun (approved):

PR #239, opened 2023-10-24
Generate entity typeOf triples.

keyurva:


simple/stats/importer.py
```diff
@@ -98,6 +100,25 @@ def _rename_columns(self) -> None:
...
+    if not self.entity_type:
+      logging.warning(
+          "Could not resolve entity type. Entity triples will not be imported.")
+      return
+    self.nodes.entities_with_type(self.df.iloc[:, 0].tolist(), self.entity_type)
+
+  def _resolve_entity_type(self) -> str:
+    all_entity_dcids = self.df.iloc[:, 0].tolist()
+    sample_entity_dcids = random.sample(all_entity_dcids,
+                                        min(len(all_entity_dcids), 5))
```
shifucun, line 117: nit: make 5 a constant; And worth documenting the behavior, ie, we assume the entity column has same type of entities.

shifucun (approved):
very nice!

simple/sample/output/process/debug_resolve_powerplants.csv
```diff
@@ -1,6 +1,6 @@
 input,dcid,link
-BAZ BAR,*UNRESOLVED*,
```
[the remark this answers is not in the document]
keyurva: Because we are using sets and they are unordered. I've made them sorted now so should maintain the order hereon.

simple/stats/importer.py
keyurva: Done.

PR #240, opened 2023-10-24
Build embeddings for custom DC SVs.

keyurva:


simple/stats/embeddings.py
```diff
@@ -0,0 +1,62 @@
...
+
+  for sv in svs:
+    _maybe_add(sv.name, sv.id)
+    _maybe_add(sv.description, sv.id)
+    for sentence in sv.nl_sentences:
+      _maybe_add(sentence, sv.id)
+
+  assert len(texts) == len(dcids)
+
+  logging.info("Encoding %s sentences", len(texts))
```
pradh, line 50: Should we try to use / refactor-and-reuse from https://github.com/datacommonsorg/website/tree/master/tools/nl/embeddings ?

keyurva: Do we have a common repo?

The common code will basically be this method: https://github.com/datacommonsorg/website/blob/master/tools/nl/embeddings/build_embeddings.py#L85-L92

simple/sample/output/process/report.json
```diff
@@ -1,37 +1,37 @@
 {
   "status": "SUCCESS",
-  "startTime": "2023-10-24 13:20:25.134318",
-  "lastUpdate": "2023-10-24 13:20:33.068269",
+  "startTime": "2023-10-24 15:07:01.417813",
```
shifucun, line 3: shall we start to have a mode to not update the timestamp in this file?

simple/stats/embeddings.py
pradh: I think more than that?   We could use validate embeddings, use the fine-tuned model if available, rely on the same naming with model name (so model name is clear), timestamp.

What you have is fine as a quick-one-off prototype though.

keyurva: I see.

Yes, reusing makes sense based on what you mentioned. We probably need a discussion on how similar or different main and custom DC should be and then extract the common parts.

Perhaps we can discuss once Jehangir is back and consider this a POC until then?

pradh: We can chat tomm.

pradh: For this PR ... does this code have access to our GCS paths and https://github.com/datacommonsorg/website/blob/5970947053bb81672a26fe286642d696b35c362c/deploy/nl/embeddings.yaml#L3 ?

If not, I think what you have is fine.

simple/stats/constants.py
```diff
@@ -22,6 +22,7 @@
 CONFIG_JSON_FILE_NAME = "config.json"
 REPORT_JSON_FILE_NAME = "report.json"
 TRIPLES_FILE_NAME = "triples.csv"
+EMBEDDINGS_FILE_NAME = "embeddings.csv"
```
pradh, line 25: Might be worth versioning it

simple/stats/embeddings.py
```diff
@@ -0,0 +1,62 @@
...
+from util.filehandler import FileHandler
+from util.filehandler import LocalFileHandler
+
+_MODEL_NAME = "all-MiniLM-L6-v2"
+_DCID_COL = "dcid"
+_SENTENCE_COL = "sentence"
+
+
+def build(svs: list[StatVar],
+          embeddings_fh: FileHandler | None = None) -> pd.DataFrame:
```
pradh, line 29: super nit: since only the test needs to pass None, don't default?

simple/sample/output/embeddings.csv
```diff
@@ -0,0 +1,8 @@
...83,84,85,86,87,88,89,90,91,92,93,94,95,96,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,123,124,125,126,127,128,129,130,131,132,133,134,135,136,137,138,139,140,141,142,143,144,145,146,147,148,149,150,151,152,153,154,155,156,157,158,159,160,161,162,163,164,165,166,167,168,169,170,171,172,173,174,175,176,177,178,179,180,181,182,183,184,185,186,187,188,189,190,191,192,193,194,195,196,197,198,199,200,201,202,203,204,205,206,207,208,209,210,211,212,213,214,215,216,217,218,219,220,221,222,223,224,225,226,227,228,229,230,231,232,233,234,235,236,237,238,239,240,241,242,243,244,245,246,247,248,249,250,251,252,253,254,255,256,257,258,259,260,261,262,263,264,265,266,267,268,269,270,271,272,273,274,275,276,277,278,279,280,281,282,283,284,285,286,287,288,289,290,291,292,293,294,295,296,297,298,299,300,301,302,303,304,305,306,307,308,309,310,311,312,313,314,315,316,317,318,319,320,321,322,323,324,325,326,327,328,329,330,331,332,333,334,335,336,337,338,339,340,341,342,343,344,345,346,347,348,349,350,351,352,353,354,355,356,357,358,359,360,361,362,363,364,365,366,367,368,369,370,371,372,373,374,375,376,377,378,379,380,381,382,383,dcid,sentence
```
pradh, line 1: for my understanding, what generates this?

simple/stats/constants.py
pradh: Maybe add TODO

simple/stats/embeddings.py
keyurva: It depends on the user running it. The current focus is custom DC and they won't have access to it. But when we run it in the future, we will. So the things you talked about above are valid and we need to have modes where we use fine tuned models, do validations, etc. over time.

Let's chat more tomorrow.

simple/sample/output/process/report.json
keyurva: Yes, I'll work on this following this PR.

simple/stats/constants.py
keyurva: Our current thinking is that the user (the mixer loader primarily) specifies separate output directories on each run (so that each run can be tracked separately) which implicitly acts as a versioning mechanism.

Would be good to get your thoughts on this as well.

simple/stats/embeddings.py
pradh: hmm, custom DC enabling NL might need that access to get the base embeddings

simple/sample/output/embeddings.csv
keyurva: This is generated by running RSI on the sample input files: https://github.com/datacommonsorg/import/tree/master/simple/sample

The sample files give users a quick taste of what's involved and we run it whenever we make changes to the scripts.

simple/stats/embeddings.py
keyurva: Done.

pradh (approved):
Thanks

PR #242, opened 2023-10-30
[RSI] Generate SV sentences.

keyurva:
* RSI now generates a CSV with dcids and sentences instead of embeddings.
* Embeddings will be generated further downstream by a separate tool. See https://github.com/datacommonsorg/website/pull/3738


simple/stats/nl.py
```diff
@@ -0,0 +1,49 @@
...
+_DCID_COL = "dcid"
+_SENTENCE_COL = "sentence"
+
+
+def generate_sv_sentences(svs: list[StatVar], sentences_fh: FileHandler):
+  """Generates sentences based on the name, description and NL sentences of the specified SVs.
+
+    The SV dcids and sentences are written to a CSV using the specified FileHandler
+    """
+  rows = []
```
pradh, line 31: nit: if this is the only reason to import pandas, perhaps use csv.DictWriter instead?

pradh (approved):
Thanks.

simple/stats/nl.py
keyurva: Pandas is used at other places in the package so using it here as well.

simple/stats/nl.py
```diff
@@ -0,0 +1,49 @@
...
+
+def _sv_sentences(sv: StatVar) -> str:
+  sentences = []
+  sentences.append(sv.name)
+  if sv.description:
+    sentences.append(sv.description)
+  for nl_sentence in sv.nl_sentences:
+    if nl_sentence:
+      sentences.append(nl_sentence)
+  return ";".join(sentences)
```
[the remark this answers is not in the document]
keyurva: Done.

keyurva:
Thanks!

shifucun:
Maybe fix the dependency issues https://github.com/datacommonsorg/import/pulls as they would alert on every PR submission.

PR #248, opened 2023-11-13
Create a class to insert nodes into a sqlite db.

keyurva:


simple/stats/db.py
```diff
@@ -0,0 +1,75 @@
...
+    date TEXT,
+    value TEXT,
+    provenance TEXT
+);
+"""
+
+_INSERT_OBSERVATIONS_STATEMENT = "insert into observations values(?, ?, ?, ?, ?)"
+
+
+class Db:
```
shifucun, line 44: call this SqliteDB since this is specific for sqlite and you will need one for cloudsql

shifucun: Also could this be general? In https://github.com/datacommonsorg/mixer/blob/a03c205f7783c416255a3fdf03af0cd87f29cc38/internal/sqldb/writer.go#L114-L146, they functions would work for both sqlite and cloudsql

keyurva: > call this SqliteDB since this is specific for sqlite and you will need one for cloudsql

I intentionally did not call it that. We'll keep the same `Db` interface when we add support for other dbs and make the implementation adapt to other dbs.

This keeps other parts of the importer agnostic of the actual db.

keyurva: > functions would work for both sqlite and cloudsql

`self.db.executemany` achieves that generalization (similar to the `sqlClient.Exec` call [here][exec]). We could have a helper that takes an insert statement and tuples but the for-comprehension is concise enough so didn't feel a need for that helper.

[exec]: https://github.com/datacommonsorg/mixer/blob/a03c205f7783c416255a3fdf03af0cd87f29cc38/internal/sqldb/writer.go#L89

shifucun: That's good then, the current PR covers both points!

shifucun (approved):

PR #253, opened 2023-11-20
Write imported data to DB.

keyurva:
* Writes triples and observations to DB.
   + Currently sqlite only, cloud sql coming soon.
* If output path is GCS, it first writes to a temp local file and copies the file to GCS before close.
* Observation and triples CSVs won't be needed once mixer loads the db directly. We can stop writing them once mixer does.

Insight into the sqlite db created for the sample import under `simple/sample`:

```
$ sqlite3 datacommons.db
SQLite version 3.39.5 2022-10-14 20:58:05
Enter ".help" for usage hints.
sqlite> .header on
sqlite> .mode column
sqlite> select count(*) from triples;
count(*)
--------
91      
sqlite> select count(*) from observations;
count(*)
--------
66      
sqlite> select * from triples limit 5;
subject_id   predicate  object_id  object_value       
-----------  ---------  ---------  -------------------
c/s/default  typeOf     Source                        
c/s/default  name                  Custom Data Commons
c/s/1        typeOf     Source                        
c/s/1        name                  Source1 Name       
c/s/1        url                   http://source1.com 
sqlite> select * from observations limit 5;
entity       variable  date  value  provenance
-----------  --------  ----  -----  ----------
country/AFG  var1      2023  0.19   c/p/1     
country/YEM  var1      2023  0.21   c/p/1     
country/AGO  var1      2023  0.29   c/p/1     
country/ZMB  var1      2023  0.31   c/p/1     
country/ZWE  var1      2023  0.37   c/p/1  
```


simple/stats/db.py
```diff
@@ -38,16 +44,32 @@
...
 class Db:
   """Class to insert triples and observations into a sqlite DB."""
 
   def __init__(self, db_file_path: str) -> None:
-    self.db = sqlite3.connect(db_file_path)
-    self.db.execute(_CREATE_TRIPLES_TABLE)
-    self.db.execute(_CREATE_OBSERVATIONS_TABLE)
+    self.db_file_path = db_file_path
+    # If file path is a GCS path, we create the DB in a local temp file
+    # and upload to GCS on commit.
```
shifucun, line 65: wonder why need to upload the db to GCS? For large db, this can take time and block?

simple/stats/importer.py
```diff
@@ -106,6 +111,30 @@ def _rename_columns(self) -> None:
...
+    observations_df = observations_df.melt(
+        id_vars=[
+            constants.COLUMN_DCID, constants.COLUMN_DATE,
+            constants.COLUMN_PROVENANCE
+        ],
+        var_name=constants.COLUMN_VARIABLE,
+        value_name=constants.COLUMN_VALUE,
+    ).dropna()
+
+    # Reorder columns so they are in the same order as observations
```
shifucun, line 127: since import tool has full control, can just make the order correct at the first shot?

simple/stats/db.py
```diff
@@ -12,10 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
```
shifucun, line 15: should we have a flag to disable this before the cloudsql is supported. Otherwise, custom DC would write to sqlite twice, one from here and one from mixer?

simple/stats/db.py
keyurva: Disabled upload until we have cloud SQL support.

simple/stats/importer.py
keyurva: We use the pandas `melt` method above (to unpivot the table) and it puts the variable and value columns at the end.

simple/stats/db.py
keyurva: Good idea. Done.

keyurva:
Thanks for the review.

PR #254, opened 2023-11-21
Add support for Cloud My SQL.

keyurva:
Snippet of running the sample import against a Cloud MySQL instance.

```
$ gcloud beta sql connect dc-graph --user=root
Starting Cloud SQL Proxy: [/Users/keyurs/projects/lib/google-cloud-sdk/bin/cloud_sql_proxy -instances datcom-website-dev:us-central1:dc-graph=tcp:9470 -credential_file /Users/keyurs/.config/gcloud/legacy_credentials/keyurs@google.com/adc.json]]
2023/11/21 11:06:25 Rlimits for file descriptors set to {Current = 8500, Max = 9223372036854775807}
2023/11/21 11:06:25 using credential file for authentication; path="/Users/keyurs/.config/gcloud/legacy_credentials/keyurs@google.com/adc.json"
2023/11/21 11:06:25 Listening on 127.0.0.1:9470 for datcom-website-dev:us-central1:dc-graph
2023/11/21 11:06:25 Ready for new connections
Connecting to database with SQL user [root].Enter password: 
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 127654
Server version: 8.0.31-google (Google)

Copyright (c) 2000, 2023, Oracle and/or its affiliates.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql> use datacommons;
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> select count(*) from triples;
+----------+
| count(*) |
+----------+
|       91 |
+----------+
1 row in set (0.07 sec)

mysql> select count(*) from observations;
+----------+
| count(*) |
+----------+
|       66 |
+----------+
1 row in set (0.07 sec)

mysql> select * from triples limit 5;
+-------------+-----------+-----------+---------------------+
| subject_id  | predicate | object_id | object_value        |
+-------------+-----------+-----------+---------------------+
| c/s/default | typeOf    | Source    |                     |
| c/s/default | name      |           | Custom Data Commons |
| c/s/1       | typeOf    | Source    |                     |
| c/s/1       | name      |           | Source1 Name        |
| c/s/1       | url       |           | http://source1.com  |
+-------------+-----------+-----------+---------------------+
5 rows in set (0.07 sec)

mysql> select * from observations limit 5;
+-------------+----------+------+-------+------------+
| entity      | variable | date | value | provenance |
+-------------+----------+------+-------+------------+
| country/AFG | var1     | 2023 | 0.19  | c/p/1      |
| country/YEM | var1     | 2023 | 0.21  | c/p/1      |
| country/AGO | var1     | 2023 | 0.29  | c/p/1      |
| country/ZMB | var1     | 2023 | 0.31  | c/p/1      |
| country/ZWE | var1     | 2023 | 0.37  | c/p/1      |
+-------------+----------+------+-------+------------+
5 rows in set (0.07 sec)
```


simple/stats/config.md
```diff
@@ -36,7 +36,16 @@ The config parameters for the files to be imported should be specified in a `con
         "Provenance2 Name": "http://source1.com/provenance2"
       }
     }
-  }
+  },
+  "database": {
+    "type": "cloudMySql",
+    "params": {
+        "instance": "datcom-website-dev:us-central1:dc-graph",
```
shifucun, line 43: wonder if we should write the testing data to live db? Also should remove the user and password here?

simple/stats/config.md
```diff
@@ -36,7 +36,16 @@ The config parameters for the files to be imported should be specified in a `con
         "Provenance2 Name": "http://source1.com/provenance2"
       }
     }
-  }
+  },
+  "database": {
+    "type": "cloudMySql",
```
shifucun, line 41: nit: [Cloud SQL](https://cloud.google.com/sql)

simple/stats/config.md
```diff
@@ -36,7 +36,16 @@ The config parameters for the files to be imported should be specified in a `con
...
+    "params": {
+        "instance": "datcom-website-dev:us-central1:dc-graph",
+        "user": "root",
+        "password": "secret",
+        "db": "datacommons"
```
shifucun, line 46: the user and password should be part of the `docker run` command line arg, rather than in the config file, which is usually checked in?

simple/stats/config.md
```diff
@@ -110,4 +119,36 @@ The URL of the source.
...
+
+If a database config is not specified, it writes to a sqlite db file named `datacommons.db` 
+in the output folder.
+
+### Database parameters
+
+#### `type`
+
+The type of database the importer connects to. Currently the importer supports 2 types of databases:
+* `cloudMySql` - A Google Cloud MySQL instance.
```
shifucun, line 137: just "cloudsql"? It matches the branding name and I would make it all lowercase for simplicity

simple/stats/config.md
```diff
@@ -110,4 +119,36 @@ The URL of the source.
...
+
+#### `params`
+
+##### Cloud MySQL parameters
+* `instance`: The Cloud MySQL instance to connect to. e.g. `datcom-website-dev:us-central1:dc-graph`
+* `user`: The DB user. e.g. `root`
+* `password`: The DB user's password.
+* `db`: The name of the DB. e.g. `dc-graph`
+
+> Browse or create your Google SQL instances [here](https://pantheon.corp.google.com/sql/instances).
```
shifucun, line 151: nit: should use "console.cloud.google.com". "corp.google.com" is internal link

simple/stats/config.md
```diff
@@ -110,4 +119,36 @@ The URL of the source.
...
+
+##### SQLite parameters
+* `dbFilePath`: The path of the sqlite db file. e.g. `data/output/datacommons.db`
```
shifucun, line 154: no need to set this path, as sqlite approach can just use the user file path?

simple/stats/config.md
keyurva: We are not. This is only in the doc. The actual sample continues to write to sqlite. The password is fake.

simple/stats/config.md
keyurva: Cloud SQL supports postgresql and sql server in addition to mysql. We need to explicitly add support for them if we want to support them.

simple/stats/config.md
keyurva: Ack. Let me do this in a separate PR.

simple/stats/config.md
keyurva: For sqlite, there should be no config ideally. I added it for symmetry. Perhaps it's cleanest to not support a sqlite config at all?

shifucun: I agree there should not be a config for sqlite. In general I would vote for the least and simplest config on the user side.

keyurva: Done. Removed.

simple/stats/config.md
keyurva: Good call out. Done.

simple/stats/config.md
shifucun: From user perspective, there is no difference in this config and data preparation? They seem to be internal implementation details which user does not need to be aware of.

simple/stats/config.md
shifucun: I see. Then maybe use something like
`instance: "<project:region:instance>"`
`user:<YOUR_USER_NAME>`...

Also, is  the db name "datacommons" dynamic or fixed?

keyurva: Dynamic.

simple/stats/config.md
keyurva: I've updated it to `cloudsql` for now but configs will likely be different for different DBs. We can revisit this if we have customers that want support for other cloud sql dbs.

simple/stats/config.md
keyurva: Done.

simple/stats/config.md
```diff
@@ -110,4 +119,33 @@ The URL of the source.
...
+If a database config is not specified, it writes to a sqlite db file named `datacommons.db` 
+in the output folder.
+
+### Database parameters
+
+#### `type`
+
+The type of database the importer connects to. Currently the importer supports 2 types of databases:
+* A SQLite DB - This is the default and no database config is needed for this.
+* `cloudMySql` - A Google Cloud MySQL instance.
```
shifucun, line 138: change to "cloudsql" here too?

shifucun (approved):
thanks for the update!

simple/stats/config.md
keyurva: Done.

keyurva:
Thanks for the review!

PR #256, opened 2023-11-28
Save report.json after 10 second intervals.

keyurva:


simple/stats/reporter.py
```diff
@@ -113,7 +123,12 @@ def _maybe_report(field: str, func=None):
 
   def save(self) -> None:
     self.last_update = datetime.now()
-    self.report_fh.write_string(json.dumps(self.json(), indent=2))
+    report = not self.last_reported or time.time(
+    ) - self.last_reported >= _REPORT_SAVE_INTERVAL_SECONDS or _is_done_status(
+        self.status)
```
shifucun, line 128: split and reformat this long expression? This is a bit hard to read.

shifucun: report -> should_report?

shifucun (approved):

simple/stats/reporter.py
keyurva: Done.

PR #264, opened 2023-12-12
Write import info to a new imports table.

keyurva:
* Verified against both sqlite and cloud sql.

### Cloud SQL Console snippet

```
mysql> use datacommons;
Database changed
mysql> show tables;
+-----------------------+
| Tables_in_datacommons |
+-----------------------+
| imports               |
| observations          |
| triples               |
+-----------------------+
3 rows in set (0.05 sec)

mysql> select * from imports;
+---------------------+---------+------------------------------+
| imported_at         | status  | metadata                     |
+---------------------+---------+------------------------------+
| 2023-12-12 10:32:16 | SUCCESS | {"numVars": 4, "numObs": 58} |
+---------------------+---------+------------------------------+
1 row in set (0.05 sec)

mysql> 
```

### DIFF
--- a/simple/stats/runner.py
+++ b/simple/stats/runner.py
@@ -21,6 +21,7 @@ from stats.db import create_sqlite_config
 from stats.db import Db
 from stats.db import get_cloud_sql_config_from_env
 from stats.db import get_sqlite_config_from_env
+from stats.db import ImportStatus
 from stats.importer import SimpleStatsImporter
 import stats.nl as nl
 from stats.nodes import Nodes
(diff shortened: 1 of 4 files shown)


simple/stats/runner.py
```diff
@@ -102,6 +103,9 @@ def run(self):
           list(self.nodes.variables.values()),
           self.nl_dir_fh.make_file(constants.SENTENCES_FILE_NAME))
 
+      # Write import info to DB.
+      self.db.insert_import_info(status=ImportStatus.SUCCESS)
```
shifucun, line 107: should we worry about exception when writing the import info?

shifucun (approved):

simple/stats/runner.py
keyurva: With any exception, the transaction will not be committed so we are ok.

In a later PR, I want to write the import info in a separate transaction so we can record failed imports as well.

PR #266, opened 2023-12-15
Remove support for importing individual files.

keyurva:
* In practice, RSI is used only for importing directories and not individual files (including by the load data tool).
* This PR removes support for importing individual files to eliminate this tech debt.
* This changes the command line flag from `--input_path` to `--input_dir`. To account for this change in the admin tool, we'll update [this code](https://github.com/datacommonsorg/website/blob/fa272ab47c70de1e5af47c681b887046152b6e25/server/routes/admin/html.py#L64) in the website repo when updating its submodules.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -103,7 +103,7 @@ function run_sample {
   rm -f sample/output/datacommons.db
 
   echo "Running sample."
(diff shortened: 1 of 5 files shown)


juliawu (approved):
Thanks for the changes!

PR #267, opened 2023-12-19
Create an importer hierarchy to support more types of importers.

keyurva:
* Introduces a new `Importer` base class, renames the current implementation to `ObservationsImporter` and makes it a sub-type of `Importer`.
* This PR does not introduce any new functionality. It only refactors the code to pave the way for adding more importers (like an `EventsImporter` soon).


juliawu (approved):
Thanks for the refactor!

keyurva:
Thanks for the review!

PR #268, opened 2023-12-28
Resolve place entities concurrently.

keyurva:
* Partitions place entities into chunks of 500 entities each and resolves them concurrently.
* This will be needed for importing large datasets such as [this one](https://docs.google.com/spreadsheets/d/18gtY3mCtxz6HDcs6iCd4cFi9c6SpUdsIvKl1nrT5SrQ/edit#gid=1171898663) which has 250K+ lat-lngs.
* Verified by running against the sample dataset and changing batch sizes locally.
   + Will add integration tests and run them in the cloud so such functionality will be auto tested on each PR.

### DIFF
--- a/simple/util/dc_client.py
+++ b/simple/util/dc_client.py
@@ -13,18 +13,19 @@
 # limitations under the License.
 """ Data Commons REST API Client."""
 
+import asyncio
+from itertools import islice
 import json
 import logging
 import os
 import re
 
+from httpx import AsyncClient
+from httpx import Limits
 import requests
 
 from .ngram_matcher import NgramMatcher
 from .resolvers import resolve_latlngs_2_s2cells
-from itertools import islice
-from httpx import AsyncClient, Limits
-import asyncio
 
 # Environment variables.
 _KEY_ENV = "DC_API_KEY"
--- a/simple/util/dc_client.py
+++ b/simple/util/dc_client.py
@@ -22,6 +22,10 @@ import requests
 
 from .ngram_matcher import NgramMatcher
 from .resolvers import resolve_latlngs_2_s2cells
+from itertools import islice
+from multiprocessing import Pool
+from httpx import AsyncClient, Limits
+import asyncio
 
 # Environment variables.
 _KEY_ENV = "DC_API_KEY"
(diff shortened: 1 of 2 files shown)


simple/util/dc_client.py
```diff
@@ -22,6 +22,10 @@
 
 from .ngram_matcher import NgramMatcher
 from .resolvers import resolve_latlngs_2_s2cells
+from itertools import islice
+from multiprocessing import Pool
```
shifucun, line 26: Pool is not used?

simple/util/dc_client.py
```diff
@@ -217,3 +252,28 @@ def post(path: str, data={}) -> dict:
         f'{resp.status_code}: {resp.reason}\n{response["message"]}\nRequest: {path}\n{data}'
     )
   return response
+
+
+async def post_async(client: AsyncClient, path: str, data={}) -> dict:
```
shifucun, line 257: @pradh maybe the website nl code can use this asyncClient to do concurrent api call.

simple/util/dc_client.py
```diff
@@ -79,13 +88,39 @@ def resolve_place_entities(
...
+
+  chunks = chunked(entities, _RESOLVE_BATCH_SIZE)
+
+  resolved: dict[str, str] = {}
+  async with AsyncClient(limits=_HTTPX_LIMITS) as client:
+    futures: dict[str, str] = [
+        _resolve_place_entities_chunk(client, chunk, entity_type, property_name)
+        for chunk in chunks
+    ]
+    for resolved_chunk in await asyncio.gather(*futures):
```
shifucun, line 108: looks like now if one request failed, all chunks are canceled? Is there a per request retry mechanism easily available from asyncio library?

shifucun (approved):
very nice!

simple/util/dc_client.py
keyurva: Good catch. Removed.

simple/util/dc_client.py
keyurva: There is a callback mechanism which we can use. Let me also investigate if there is an easier option for retries. Will raise a separate PR for it.

keyurva:
Thanks for the review!

PR #269, opened 2024-01-11
Fix NullPointerException when running the import and bucket flag is not passed in.

megaz:
Fix NullPointerException error for when running the import tool and a bucket is not passed in as a command line flag.


keyurva (approved):

megaz:
Hi @shifucun  @keyurva - would you be able to kindly merge this please if it is ready to ship since I don't have permissions to merge.  Thank you!

pradh (approved):
Thanks!  To confirm, do you actually use the server?

megaz:
> Thanks! To confirm, do you actually use the server?

Thank you, yes I ran the server locally for testing purposes.

PR #271, opened 2024-01-16
Add support for event count aggregations.

keyurva:
* Adds support for creating count SV from events.
* The framework is set to support other aggregate SVs (e.g. sum, average, mix, max, etc.)
* Another change is that we attempt to repurpose SV names into SV ids if replacing spaces and dashes make it a valid dcid. Otherwise we use auto generated dcids like before. Various golden changes are because of this change.
* TODO: integration into main RSI runner, event and aggregation config documentation.


simple/stats/events_importer.py
```diff
@@ -259,3 +301,14 @@ def _write_debug_csvs(self) -> None:
...
+
+# Utility methods
+def _time_period(date_str: str, period: TimePeriod) -> str:
+  date = date_parse(date_str)
+  if period == TimePeriod.DAY:
+    return date.strftime("%Y-%m-%d")
+  if period == TimePeriod.YEAR:
+    return date.strftime("%Y")
+  # Default to month
+  return date.strftime("%Y-%m")
```
shifucun, line 314: maybe default to year since we mostly have yearly data?

simple/stats/events_importer.py
```diff
@@ -259,3 +301,14 @@ def _write_debug_csvs(self) -> None:
                    self.debug_resolve_fh)
       self.debug_resolve_fh.write_string(
           self.debug_resolve_df.to_csv(index=False))
+
+
+# Utility methods
+def _time_period(date_str: str, period: TimePeriod) -> str:
```
shifucun, line 307: Since this is applied on each cell and there could be a lot of duplicate (if all from one year). Not sure if parsing date on each cell would be a performance issue. But maybe this can be optimized by having a "cache" object and a return an inner nested function like:

def _time_period():
  cache = {}
  def actual_func(...):
     # use and check cache.
      ...
   return actual_func

shifucun (approved):
Very nice! Would suggest to profile pandas in general as I found the way to use pandas can make orders of magnitude difference, especially for large input

simple/stats/events_importer.py
shifucun: github removes indention. `actual_func()` should be an inner function of _time_period and `cache` is captured.

simple/stats/config.py
```diff
@@ -70,6 +76,12 @@ def variable(self, variable_name: str) -> StatVar:
         group_path=var_cfg.get(_GROUP_FIELD, ""),
     )
 
+  def aggregation(self, variable_name: str) -> AggregationConfig:
+    aggregation_cfg = self.data.get(_VARIABLES_FIELD,
```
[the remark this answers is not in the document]
keyurva: Didn't know about this - this is great!

simple/stats/events_importer.py
keyurva: Done.

keyurva:
Thanks for the review.

PR #272, opened 2024-01-17
Integrate events importer into RSI.

keyurva:
* Also added an events import to the sample.


simple/sample/input/config.json
```diff
@@ -24,6 +24,14 @@
     "wikidataids.csv": {
       "entityType": "Country",
       "provenance": "Provenance1 Name"
+    },
+    "latlng_events.csv": {
+      "importType": "events",
```
shifucun, line 29: maybe skip importType as it can deduce from eventType?

simple/sample/input/config.json
```diff
@@ -24,6 +24,14 @@
...
+      "eventType": "CrimeEvent",
+      "entityType": "CensusZipCodeTabulationArea",
```
shifucun, line 31: with eventType, the entityType becomes a bit confusing now.. I think this worth some thoughts now.

simple/sample/output/debug/triples.csv
```diff
@@ -1,4 +1,304 @@
 subject_id,predicate,object_id,object_value
+JG497095,typeOf,CrimeEvent,""
```
shifucun, line 2: do we need to store the entire table as debug csv? That feels too much

keyurva: We actually don't need to store the `triples.csv` at all anymore. I'll updated the code so it's no longer generated.

This one however, is just the [shell script](https://github.com/datacommonsorg/import/blob/master/run_test.sh#L108) saving the tables as triples. This is to ensure that the expected changes are correctly captured and that the tables don't change when they are not expected to.

I think the name `debug` is a misnomer here. So have changed it to `tables` so it's more indicative.

simple/sample/input/config.json
keyurva: I think it's better to be explicit here to avoid confusion when we add more types of imports (like entities). So keeping this as-is.

simple/sample/input/config.json
keyurva: I agree. It's really *location* type. However, we've used `entityType` for observations and will be using it for entities as well so kept the same name for consistency.

I'll try to make sure to clarify this in the config documentation.

simple/sample/input/config.json
shifucun: but the above ones should have `importType: "observations"` to match here?

keyurva: `observations` is default so didn't add it earlier. But you're right, should match it for consistency. Let me add it.

simple/sample/output/debug/triples.csv
shifucun: I see

simple/sample/input/config.json
keyurva: Done.

shifucun (approved):

keyurva:
Thanks for the review.

PR #273, opened 2024-01-19
Ignore NA properties, adjust httpx config, add place types.

keyurva:


simple/stats/events_importer.py
```diff
@@ -187,7 +187,7 @@ def _write_event_triples(self) -> None:
       properties: dict[str, str] = {}
 
       for i, (k, v) in enumerate(row.items()):
-        if i < 2:
+        if i < 2 or pd.isna(v):
```
shifucun, line 190: wonder should filter nan first? In general prefer batch and vector operation in pandas

shifucun (approved):

simple/stats/events_importer.py
keyurva: We filter na for observations and don't insert observations that are na altogether. In this case, the NA check is for an individual property value. If it is NA, we don't insert that PV but we still import the overall event.

PR #274, opened 2024-01-24
Add initial support for main dc output.

keyurva:
* Added support for generating CSVs and TMCF for main dc import.
* Made `Db` an abstract class with concrete implementations for `SqlDb` (existing impl) and `MainDcDb` (new).
  + Should rename this to `DCStore` later since it's no longer a DB per se.
* Added a `--mode=maindc` flag to the script for generating main dc output.
* Generated main dc output for the sample files. This is under the `sample/main_dc_output` folder.
* TODOs: schema mcf, svobs format.


simple/sample/main_dc_output/observations.tmcf
```diff
@@ -0,0 +1,6 @@
+Node: E:Table->E0
+typeOf: dcs:StatVarObservation
+variableMeasured: C:Table->variable
+observationDate: C:Table->date
+observationAbout: C:Table->entity
+value: C:Table->value
```
pradh, line 6: I'm puzzled, I thought this format isn't supported?

For the current supported format I'd expect TMCF to have multiple nodes, one per SV?

keyurva: For the sql output, it is loaded into an `observations` table with the following columns: `entity`, `variable`, `date`, `value`, `provenance`. So while the input format that is supported currently is observations per column, it is converted into this representation for sql insertion.

pradh: I see!

pradh: Oh, actually, if we do passthru (places have dcids) we should find TMCF reflecting the original CSV structure right?

keyurva: We could. I'm wondering if there are benefits in always producing output in the same format though?

pradh: I'm not sure there is.  The SVObs format makes it very verbose (larger) and requires processing by RSI always on a single task.

pradh (approved):

simple/sample/main_dc_output/observations.tmcf
pradh: That's outside the scope of this PR, so for now, LGTM.

keyurva: Agree on the verbosity. Not clear about the single task comment. We can discuss it offline though.

keyurva:
Thanks for the review.

simple/sample/main_dc_output/observations.tmcf
pradh: What I mean is, if the CSV comes with place DCIDs in current format, it should be a metadata-only operation, and not produce SVObs format.

keyurva: Ah - got it. So it literally needs no processing (other than header parsing). Just copy the csv from the input to the output folder and produce a corresponding tmcf.

pradh: Yes!

keyurva: Niceness!

pradh: There might be value in RSI being a passthru for ID-based resolution as well, because they scale better.

So only datasets that need name/lat-lng resolutions have size limitations.

PR #275, opened 2024-01-24
Add support for main dc schema.mcf.

keyurva:
* Currently creates `schema.mcf` only for SVs and SVGs.


simple/sample/main_dc_output/schema.mcf
```diff
@@ -0,0 +1,46 @@
...
+typeOf: StatisticalVariable
+name: "Good var1 name"
+description: "Good var1 description"
+memberOf: custom/g/group_2
+
+Node: dcid:var2
+typeOf: StatisticalVariable
+name: "Good var2 name"
+memberOf: custom/g/group_3
+
```
pradh, line 31: This will fail validation.  SV needs properties like populationType, measuredProperty, statType?

keyurva: Added populationType and measuredProperty. PLMK if they look correct.

As for statType - should it be something users need to specify in a config? And what should be the default?

pradh: https://github.com/datacommonsorg/data/blob/master/docs/representing_statistics.md#statisticalvariable 

`measuredValue` is the default

simple/stats/data.py
```diff
@@ -31,6 +31,8 @@
 _PREDICATE_SUB_CLASS_OF = "subClassOf"
 _PREDICATE_OBSERVATION_DATE = "observationDate"
 _PREDICATE_LOCATION = "location"
+_PREDICATE_POPULATION_TYPE = "populationType"
+_PREDICATE_MEASURED_PROPERTY = "measuredProperty"
```
pradh, line 35: I we should support statType too.  And everything else in https://github.com/datacommonsorg/data/blob/master/docs/representing_statistics.md#statisticalvariable ?

pradh (approved):
If you want to add these in follow up PR so this can go in, given current RTT, approving this.

simple/sample/main_dc_output/schema.mcf
keyurva: Perfect. Filed b/322909728

simple/stats/data.py
keyurva: SG. b/322909728

keyurva:
> If you want to add these in follow up PR so this can go in, given current RTT, approving this.

Thanks. Yes, will add in a follow up PR.

PR #276, opened 2024-01-30
Create docker image for main dc RSI.

keyurva:
* Verified by building and running the docker image locally.
* Will create cloud build in a followup PR.


build/rsi/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+FROM python:3.11.4-slim as base
+
+ARG ENV
+ARG PIP_DISABLE_PIP_VERSION_CHECK=1
+ARG PIP_NO_CACHE_DIR=1
+ENV ENV=${ENV}
+
+WORKDIR /workspace
+
+COPY simple/ .
```
shifucun, line 25: now under `import` repo there are `simple` and `rsi(really simple import` concepts. Might consider renaming folders later?

build/rsi/run/README.md
```diff
@@ -0,0 +1,8 @@
+# Run RSI Docker image locally
```
shifucun, line 1: I think `run` folder should not be inside `build` as it's an "application" of the docker image not how to build it.

That said, maybe give it a more specific name other than "run"? Same apply for "run_docker.sh" which is too generic.

shifucun (commented):
Suggest to reconsider the folder naming as these are user facing files.

build/rsi/Dockerfile
pradh: How about calling it "lite", "dclite", or something similar instead of an acronym that has a more jargon feel?

keyurva: Good callouts.

> now under `import` repo there are `simple` and `rsi`

Any thoughts on reusing the "simple" and calling this `simple` or `simple_import`?

build/rsi/run/README.md
keyurva: Any thoughts on where the folder should reside? We can give it the same name that we come up with for the build folder and call it `run_that_name.sh`.

One option is to put it under a new `tools` folder. And suppose we go with `build/simple` for the build folder, this can be called `tools/simple` and the run file, `run_simple.sh`.

The full set of files will look like this:

`import/simple` (code)
`import/build/simple/Dockerfile`
`import/tools/simple/run_simple.sh`
`import/build/ci/cloudbuild.simple.yaml`

etc.

shifucun: I like this idea!

build/rsi/Dockerfile
shifucun: Given the proposal in the other comment, reusing "simple" is simple.

keyurva: Done.

build/rsi/run/README.md
keyurva: Done.

shifucun (approved):
the structure is simple now!

PR #277, opened 2024-02-02
Add support for SV properties.

keyurva:


simple/stats/config.md
```diff
@@ -29,6 +29,12 @@ The config parameters for the files to be imported should be specified in a `con
       "description": "Var 3 Description",
       "nlSentences": ["Sentence 1", "Sentence 2"],
       "group": "Parent Group/Child Group 2",
+      "properties": {
+        "populationType": "schema:Person",
```
shifucun, line 33: is "schema:" necessary here? @pradh

shifucun: Since this is an example, I would suggest adding some instruction on when to use "schema:" "dc:" or without any prefix (if needed)

shifucun (approved):

keyurva:
Thanks for the reviews.

PR #281, opened 2024-02-12
Improve performance of writing resolved entities.

keyurva:
* Replaced the pandas `replace` method with the `map` method.
  + Without this change, the code got stuck when writing a dataset with 250K entities.
* Also, we are now renaming columns early so we can refer to them by name vs by index.


shifucun (approved):
nice!

PR #285, opened 2024-02-26
Simplify env variables to be the same for both local and GCS paths.

keyurva:
* `GCS_INPUT_DIR` and `GCS_OUTPUT_DIR` vars no longer needed.
* Both GCS and local paths can be specified using `INPUT_DIR` and `OUTPUT_DIR`.


shifucun (approved):
nice!

PR #287, opened 2024-03-04
Add support for input file wildcards.

keyurva:
* Currently, config needs to be specified on a per file basis.
* With this PR, the same config can be applied to all files that match a pattern.
* Note that we only support simple wildcard patterns and not full regex so it's simple and intuitive.


simple/stats/config.py
```diff
@@ -120,7 +129,32 @@ def database(self) -> dict:
...
 
   def _input_file(self, input_file_name: str) -> dict:
-    return self.data.get(_INPUT_FILES_FIELD, {}).get(input_file_name, {})
+    # Exact match.
+    input_file_config = self._input_files_config.get(input_file_name, {})
+    if input_file_config:
+      return input_file_config
+
+    # Wildcard match
+    if input_file_name not in self._input_file_name_keys.keys():
```
shifucun, line 138: no need to do this "if" check since it's covered above?

simple/stats/config.py
```diff
@@ -120,7 +129,32 @@ def database(self) -> dict:
...
+      if "*" not in input_file_pattern:
+        continue
+      regex = self._input_file_pattern_to_regex(input_file_pattern)
+      if re.match(regex, input_file_name) is not None:
+        return input_file_pattern
+    return None
+
+  def _input_file_pattern_to_regex(self, input_file_pattern: str) -> str:
+    """
+    Transforms a string of the form "a*b.c" to the regex "a.*b\.c".
```
shifucun, line 155: why need the transform here?

simple/stats/config.py
keyurva: It is needed. The first time an input file is accessed and it is not individually specified, it will not be in the `self._input_file_name_keys` dict and this conditional will create and cache an entry for it. Subsequent calls will just use this entry.

simple/stats/config.py
keyurva: Because we do regex matching internally and the literal "a*b.c" is not one. It is a simple glob-like wildcard pattern which won't work for regex matching.

shifucun (approved):

keyurva:
Thanks for the review.

PR #291, opened 2024-03-12
Add support for importing entities.

keyurva:


simple/tests/stats/test_data/entities_importer/input/without_id_column.csv
```diff
@@ -0,0 +1,4 @@
+CASE,IUCR,PRIMARY DESCRIPTION,SECONDARY DESCRIPTION
```
shifucun, line 1: why without_id_column.csv the same as with_id_column.csv?

simple/stats/config.md
```diff
@@ -92,6 +120,32 @@ and this field associates one of the provenances defined there to this file.
...
+
+#### `idColumn`
+
+Applicable to events and entities.
+
+Every event and entity is identified by a unique ID.
+If one of the columns encodes that ID, it can be indicated with the `idColumn` field.
+If none is specified, IDs will be auto generated.
+
+#### `entityColumns`
```
shifucun, line 137: should this call "propertyColumns"?

simple/stats/config.md
```diff
@@ -73,12 +93,20 @@ Example:
...
 
+#### `importType`
+
+This importer supports importing observations, events and entities.
+The type of import can be specified with the `importType` parameter.
+Possible values: `observations` (default), `events`, `entities`.
+
 #### `entityType`
 
 All entities in a given file must be of a specific type. This type should be
```
shifucun, line 104: nit: "one specific type"

simple/stats/config.md
```diff
@@ -73,12 +93,20 @@ Example:
 
 ### Input file parameters
 
+#### `importType`
+
+This importer supports importing observations, events and entities.
+The type of import can be specified with the `importType` parameter.
+Possible values: `observations` (default), `events`, `entities`.
```
shifucun, line 100: should the importType be "Entity", "Observation" and "Event"?

simple/stats/config.md
keyurva: Hmm - all columns are properties so `propertyColumns` won't distinguish between values and ids. Open to other names though - `entityColumns` is not the best name.

simple/stats/config.md
```diff
@@ -92,6 +120,32 @@ and this field associates one of the provenances defined there to this file.
...
+
+Applicable to events and entities.
+
+Every event and entity is identified by a unique ID.
+If one of the columns encodes that ID, it can be indicated with the `idColumn` field.
+If none is specified, IDs will be auto generated.
+
+#### `entityColumns`
+
+Applicable to entities.
```
shifucun, line 139: what if this is not specified?

simple/stats/config.md
```diff
@@ -132,6 +186,28 @@ Use "/" as a separator to specify a multi-level hierarchy.
 An array of NL sentences to be used for creating more NL embeddings (in addition to the description)
 for the variable.
 
+## `entities`
```
shifucun, line 189: I lean towards not having this field. If needed, user should just add name and description columns in the csv?

simple/stats/config.md
keyurva: Don't want to break backward compatibility. So going with what we have.

simple/stats/config.md
keyurva: Done.

simple/stats/config.md
keyurva: Then all columns will be considered object values.

simple/stats/config.md
keyurva: Can you elaborate - not sure how they can be in the CSV?

Also, this is optional. If they don't provide one, the name will be same as the type.

simple/stats/config.md
shifucun: As discussed, we should handle both and document the new/desired one.

shifucun (approved):
LG to unblock. As discussed, we should think through the entire config and come up with a simple solution.

keyurva:
Thanks for the review.

Removed `config.md` from this PR. Bo and I will take a holistic view of the config, make any necessary changes and then checkin the updated doc.

PR #297, opened 2024-04-08
Add support for testing SVG hierarchy with MCF inputs.

keyurva:
* Added test for [main.in.mcf](https://source.corp.google.com/piper///depot/google3/datacommons/prophet/flume_generator/test_data/svg/main.in.mcf) from G3 and can add tests for more over time.
* Fixed a bug where if there were multiple SVs in the same leaf, it only added one of them. (Testing `main.in.mcf` surfaced this bug.)
* Changed the SVG output to be sorted by SVG ID - makes it easier to follow the hierarchy visually.
* Also, separating camel case words when generating SVG names now to make the names consistent with main DC generated names.

### DIFF
--- a/simple/kg_util/mcf_parser.py
+++ b/simple/kg_util/mcf_parser.py
@@ -23,8 +23,8 @@ from csv import reader
 import dataclasses
 import sys
 
-_VALUE = 'VALUE'
-_ID = 'ID'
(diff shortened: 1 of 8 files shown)


keyurva:
Thanks!

PR #301, opened 2024-04-12
Insert generated SVG hierarchy triples into DB.

keyurva:
* Generation of SVG hierarchy is controlled by a `generateHierarchy` config property.
  + I'm happy to use a different name for the property if any suggestions.
* It is not supported for main DC regardless of this config.
* The tests verify sqlite support. Cloud SQL was verified by running the sample import against a cloud sql instance.

image

The above shows that some SV memberships need to be worked on.
* e.g. SVs are both in a custom DC root SV as well as the generated hierarchy.
* Will discuss this with Prashanth when he's back.
* That does not affect the DB insertion so can implement those changes in a follow-up PR.


simple/sample/input/config.json
```diff
@@ -79,5 +79,6 @@
         "Provenance2 Name": "http://source1.com/provenance2"
       }
     }
-  }
+  },
+  "generateHierarchy": true
```
shifucun, line 83: groupStatVarByProperty? Since hierarchy is generated if MCF is provided too?

simple/stats/db.py
```diff
@@ -92,6 +93,8 @@
 
 _INSERT_IMPORTS_STATEMENT = "insert into imports values(?, ?, ?)"
 
+_SELECT_TRIPLES_WITH_TYPE_OF = "select * from triples where subject_id in (select subject_id from triples where predicate = 'typeOf' and object_id = ?)"
```
shifucun, line 96: I am bit confused on what this query does?

simple/stats/runner.py
```diff
@@ -132,6 +134,9 @@ def run(self):
       # Write triples to DB.
       self.db.insert_triples(triples)
 
+      # Generate SVG hierarchy.
+      self._generate_svg__hierarchy()
```
shifucun, line 138: have an extra underscore here?

simple/stats/db.py
keyurva: It first gets dcids of nodes whose `typeOf` is the specified type and then it gets all triples for those dcids.

For the SVG hierarchy case, we want all triples of all SVs and we make this query with `object_id = 'StatisticalVariable'` for that.

simple/sample/input/config.json
keyurva: I like that. Done.

simple/stats/runner.py
keyurva: Good catch. Removed.

simple/stats/config.py
```diff
@@ -49,6 +49,7 @@
 _ROW_ENTITY_TYPE_FIELD = "rowEntityType"
 _ENTITY_COLUMNS = "entityColumns"
 _ENTITIES_FIELD = "entities"
+_GENERATE_HIERARCHY = "groupStatVarByProperty"
```
shifucun, line 52: nit change to GROUP_STAT_VAR_BY_HIERARCHY

simple/stats/db.py
shifucun: I see, then maybe change the var name to _SELECT_TRIPLES_BY_SUBJECT_TYPE

simple/stats/db.py
```diff
@@ -226,13 +237,21 @@ def insert_import_info(self, status: ImportStatus):
   def commit_and_close(self):
     self.engine.commit_and_close()
 
+  def select_triples_with_type_of(self, type_of: str) -> list[Triple]:
```
shifucun, line 240: type_of -> subject_type

shifucun (approved):
Some nit for naming, otherwise look good.

simple/stats/db.py
keyurva: Done.

simple/stats/config.py
keyurva: Renamed to `_GROUP_STAT_VARS_BY_PROPERTY`. Also pluralized the property name. i.e. `groupStatVarByProperty` -> `groupStatVarsByProperty`.

simple/stats/db.py
keyurva: Done.

keyurva:
Thanks for the review.

PR #302, opened 2024-04-16
Support generating NL sentences for topics.

keyurva:
* Prior to this PR, the NL sentences CSV (which is subsequently used to generate embeddings) was generated only for SVs.
* With this PR, if a new config field called `generateHierarchy` is set to `true`, the sentences CSV will be generated for topics instead of SVs.
  + Note that we are not generating topics with this tool yet. But we'll do a follow-up PR which will. The `generateHierarchy` will then drive both generating topics as well as creating the sentences CSV for it.
* Also, prior to this PR, the `nlSentences` configured for SVs were used for generating the sentences CSV but they were not persisted as triples. With this PR, they're persisted as `searchDescription` triples.


simple/stats/data.py
```diff
@@ -34,6 +34,7 @@
 _PREDICATE_POPULATION_TYPE = "populationType"
 _PREDICATE_MEASURED_PROPERTY = "measuredProperty"
 _PREDICATE_STAT_TYPE = "statType"
+_PREDICATE_UTTERANCE_TEMPLATE = "utteranceTemplate"
```
shifucun, line 37: @pradh this schema is very confusing, why would this be the nl sentence? From previous discussion with Keyur, feels like this is a transformation function rather than schema or the actual description.

pradh: That's right, this can originally be "population of $Place" too.   Would we prefer to use something like "searchDescription" ?

shifucun: yea, searchDescription sg

keyurva: Updated to use `searchDescription`.

simple/stats/data.py
```diff
@@ -138,6 +139,14 @@ def triples(self) -> list[Triple]:
       triples.append(
           Triple(self.id, _PREDICATE_DESCRIPTION,
                  object_value=self.description))
+
+    # Encode NL sentences as searchDescription triples.
+    for nl_sentence in self.nl_sentences:
```
shifucun, line 144: do you want to rename nl_sentences to search_descriptions as well?

simple/stats/nl.py
```diff
@@ -12,38 +12,68 @@
...
 from util.filehandler import FileHandler
 
 _DCID_COL = "dcid"
 _SENTENCE_COL = "sentence"
 _SENTENCE_SEPARATOR = ";"
 
 
-def generate_sv_sentences(svs: list[StatVar], sentences_fh: FileHandler):
-  """Generates sentences based on the name, description and NL sentences of the specified SVs.
+def generate_nl_sentences(triples: list[Triple], sentences_fh: FileHandler):
```
shifucun, line 29: generate_search_descriptions?

simple/stats/schema_constants.py
```diff
@@ -19,21 +19,23 @@
 SV_HIERARCHY_PROPS_BLOCKLIST: set[str] = {
     "dcid", "typeOf", "isPublic", "provenance", "resMCFFile", "keyString",
```
shifucun, line 20: i wonder if you want to sort set items and show one item per line (if lint permits), so future diff is easier to see.

simple/stats/data.py
keyurva: Deprecated the `nlSentences` config property and using `searchDescriptions` now. Can remove the former completely down the line.

simple/stats/nl.py
keyurva: This generates an intermediary sentences.csv which is used to generate embeddings. So left it as-is.

shifucun (approved):

keyurva (commented):
Thanks for the review.

PR #295, opened 2024-04-02
SV hierarchy: Create SVG tree.

keyurva:
* Create a tree of SVGs.
* TODO: create SVG triples and write to DB.

PR #304, opened 2024-05-09
Put SVGs under a hierarchy root.

keyurva:
* This PR puts all hierarchy SVGs under a custom DC root so the custom hierarchy is cleanly separated from any main dc hierarchies.
* It also sets us up to do vertical tagging in the next PR.

### DIFF
--- a/simple/stats/schema_constants.py
+++ b/simple/stats/schema_constants.py
@@ -41,3 +41,7 @@ DEFAULT_POPULATION_TYPE = "Thing"
 
 CUSTOM_SVG_PREFIX = "c/g/"
 ROOT_SVG_ID = "dc/g/Root"
+
+# SV hierarchy groups will be put under this root.
+HIERARCHY_ROOT_SVG_ID = f"{CUSTOM_SVG_PREFIX}hierarchy"
+HIERARCHY_ROOT_SVG_NAME = "Hierarchy"
--- a/simple/stats/schema_constants.py
+++ b/simple/stats/schema_constants.py
@@ -42,6 +42,5 @@ DEFAULT_POPULATION_TYPE = "Thing"
 CUSTOM_SVG_PREFIX = "c/g/"
 ROOT_SVG_ID = "dc/g/Root"
 
-# SV hierarchy groups will be put under this root.
-HIERARCHY_ROOT_SVG_ID = f"{CUSTOM_SVG_PREFIX}hierarchy"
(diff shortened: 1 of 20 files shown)


simple/stats/schema_constants.py
```diff
@@ -41,3 +41,7 @@
 
 CUSTOM_SVG_PREFIX = "c/g/"
 ROOT_SVG_ID = "dc/g/Root"
+
+# SV hierarchy groups will be put under this root.
+HIERARCHY_ROOT_SVG_ID = f"{CUSTOM_SVG_PREFIX}hierarchy"
+HIERARCHY_ROOT_SVG_NAME = "Hierarchy"
```
keyurva, line 47: Let me know if you can think of a better name for this.

pradh: Variables?  Custom DC name?

simple/stats/stat_var_hierarchy_generator.py
```diff
@@ -47,9 +47,25 @@ def _generate_internal(triples: list[Triple]) -> "StatVarHierarchy":
   svgs = _create_all_svgs(svs)
   # Sort by SVG ID so it's easier to follow the hierarchy.
   svgs = dict(sorted(svgs.items()))
+  # Create hierarchy root SVG.
```
pradh, line 50: hmm, why not directly under dc/g/Root?

pradh (approved):
Could you add rationale for the change?

simple/stats/stat_var_hierarchy_generator.py
keyurva: I'm fine putting it directly under `dc/g/Root`. Was putting it under a different root following an earlier discussion: https://github.com/datacommonsorg/import/pull/295#discussion_r1556161136

Let me know your preference.

keyurva: Chatted offline - decided to put hierarchy under custom dc root `c/g/Root` and always create one if generating sv hierarchy.

Note that the dcid was changed from `custom/g/Root` to `c/g/Root` to follow convention and goldens have changed accordingly.

simple/stats/schema_constants.py
keyurva: The name will be `"Custom Variables"` (that of the custom dc default root).

keyurva:
Thanks for the review.

PR #307, opened 2024-05-16
Integrate vertical tagging into the main runner.

keyurva:
* The vertical specs file can be configured via a new `verticalSpecsFile` field.


simple/stats/constants.py
```diff
@@ -82,3 +82,8 @@
 # DC links
 DC_HOME = "https://datacommons.org"
 DC_BROWSER = "https://datacommons.org/browser"
+
+# "Special" filenames.
+# i.e. if files have these names, they are handled in specific ways.
+VERTICAL_SPECS_FILE_NAME = "dc.vertical_specs.json"
+SPECIAL_FILE_NAMES = set([VERTICAL_SPECS_FILE_NAME])
```
pradh, line 89: Wonder why not refer to a field in the config for the file name instead of assuming a specific filename?

simple/stats/runner.py
```diff
@@ -182,28 +185,52 @@ def _generate_svg_hierarchy(self):
...
+          if not self._is_special_file_handler(fh):
+            input_fhs.append(fh)
         for input_file in sorted(input_handler.list_files(extension=".mcf")):
-          input_mcf_fhs.append(input_handler.make_file(input_file))
+          fh = input_handler.make_file(input_file)
+          if not self._is_special_file_handler(fh):
+            input_mcf_fhs.append(fh)
+        for input_file in sorted(input_handler.list_files(extension=".json")):
+          fh = input_handler.make_file(input_file)
+          self._is_special_file_handler(fh)
```
pradh, line 233: hmm, what does it mean for a bool to be by itself?

simple/stats/constants.py
keyurva: Going with the overall theme of keeping the config minimal. Not a strong preference though - I can go either way here. Let me know your preference.

simple/stats/runner.py
keyurva: It's for the side effect of adding it to the `special_handlers`.

See method doc: `"as a side effect, it appends it to the self.special_handlers dict as well."`

pradh (approved):

simple/stats/runner.py
keyurva: I like that - changed method name to `_maybe_set_special_fh`.

Special files could be CSVs - e.g. when we add support for generating descriptions.

simple/stats/constants.py
keyurva: Chatted with Bo offline and decided to specify it in config.

shifucun (approved):

keyurva:
Thanks for the reviews.

PR #308, opened 2024-05-20
Generate special svg names.

keyurva:
* Mapping to schema names will be in a follow up PR.

### DIFF
--- a/simple/stats/stat_var_hierarchy_generator.py
+++ b/simple/stats/stat_var_hierarchy_generator.py
@@ -120,10 +120,26 @@ class SVPropVals:
       svg_name = f"{svg_name} With {pvs_str}"
     return svg_name
 
+  def gen_specialized_name(self, parent_pvs: Self) -> str:
+    parent_parts = parent_pvs._get_pv_parts()
+    child_parts = self._get_pv_parts()
+    parts = [part for part in child_parts if part not in parent_parts]
+    return ", ".join(map(lambda part: _capitalize_and_split(part), parts))
+
   # Creates and returns a new SVPropVals object with the same fields as this object
   # except for PVs which are set to the specified list.
   def with_pvs(self, pvs: list[PropVal]) -> Self:
     return replace(self, pvs=pvs)
+
+  # Returns an ordered set of PVs as a dict (since sets don't maintain order).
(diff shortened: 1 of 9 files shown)


simple/util/dc_client.py
```diff
@@ -244,6 +244,50 @@ def resolve_entity_type(entity_dcids: list[str]) -> str:
...
+        _get_property_of_entities_chunk(client, chunk, property_name)
+        for chunk in chunks
+    ]
+    for result_chunk in await asyncio.gather(*futures):
+      result.update(result_chunk)
+
+  return result
+
+
+async def _get_property_of_entities_chunk(client: AsyncClient,
```
pradh, line 269: maybe this could be in a separate PR?

simple/stats/stat_var_hierarchy_generator.py
```diff
@@ -136,7 +152,7 @@ def __init__(self, svg_id: str, svg_name: str) -> None:
     # Maintaining order maintains results consistency and helps with tests.
     self.sv_ids: dict[str, bool] = {}
     self.parent_svg_ids: dict[str, bool] = {}
-    self.child_svg_ids: dict[str, bool] = {}
+    self.child_svg_id_2_specialized_name: dict[str, str] = {}
```
pradh, line 155: How does this get surfaced?

pradh (approved):
nice!  this part has turned out to be a lot simpler here than in the flume!

simple/stats/stat_var_hierarchy_generator.py
keyurva: Currently it does not. I'll generate a svg cache in a follow-on PR and it will get surfaced there.

simple/util/dc_client.py
keyurva: huh - this _was_ in a separate [PR](https://github.com/datacommonsorg/import/pull/309). Let me see how it got here.

keyurva (commented):
Thanks for the review!

simple/util/dc_client.py
keyurva: OK - rebased correctly now so this is gone for this PR.

PR #309, opened 2024-05-20
Add method to fetch a given property from multiple entities.

keyurva:
* This method will be used to fetch schema names to generate specialized SVG names.

### DIFF
--- a/simple/util/dc_client.py
+++ b/simple/util/dc_client.py
@@ -127,7 +127,6 @@ async def _resolve_place_entities_chunk(client: AsyncClient,
       "nodes": entities_chunk,
       "property": f"<-{property_name}{type_of}->dcid",
   }
-  # TODO: handle pagination.
   response = await post_async(client, path="/v2/resolve", data=data)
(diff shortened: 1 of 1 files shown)


simple/util/dc_client.py
```diff
@@ -244,6 +244,49 @@ def resolve_entity_type(entity_dcids: list[str]) -> str:
...
+  data = {
+      "nodes": entities_chunk,
+      "property": f"->{property_name}",
+  }
+
+  logging.info("Fetching nodes: %s", data)
+  response = await post_async(client, path="/v2/node", data=data)
+
+  result_chunk: dict[str, str] = {}
+  for entity_dcid, entity_data in response.get("data", {}).items():
```
shifucun, line 281: maybe add a note to handle pagination later..

shifucun (approved):

simple/util/dc_client.py
keyurva: Done.

PR #312, opened 2024-05-21
Add functionality to fetch entity names from sql.

keyurva:
* This will be used to generate SVG cache.


simple/stats/db.py
```diff
@@ -190,9 +192,13 @@ def commit_and_close(self):
         OBSERVATIONS_TMCF)
 
     # Not supported for main DC at this time.
```
shifucun, line 194: what does "not supported for main DC" mean?

shifucun (approved):

simple/stats/db.py
keyurva: This will be called for generating the custom dc SVG cache. We don't build the SVG cache for main DC here so not supporting it.

keyurva (commented):
Thanks for the review.

PR #315, opened 2024-05-24
Add method to get names from both sql and remote dc.

keyurva:
* The names will be used for generating svg hierarchy group names.

### DIFF
--- a/simple/stats/schema.py
+++ b/simple/stats/schema.py
@@ -1,0 +1,32 @@
+# Copyright 2024 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


simple/stats/schema.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+# Gets names of the specified dcids first from db and any remaining
+# ones from remote dc.
+def get_schema_names(dcids: list[str], db: Db) -> dict[str, str]:
+  db_dcid2name = db.select_entity_names(dcids)
+  dcids = list(filter(lambda x: x not in db_dcid2name, dcids))
+  remote_dcid2name = {}
+  if dcids:
+    remote_dcid2name = dc_client.get_property_of_entities(
```
shifucun, line 30: is this code used before import or after import? Looks like here assumes it's after import when the custom DC instance is up running?

keyurva: No it does not assume the custom dc instance is running. It looks up schema names in whatever is populated in the DB at call time and in remote DC.

shifucun (approved):

keyurva:
Thanks for the review.

PR #316, opened 2024-05-28
Wire schema names into sv hierarchy generator.

keyurva:
* Also updated the generator to return svg specialized names to the runner. This will be used in a follow up PR to generate json equivalent of the [StatVarGroups](https://source.corp.google.com/piper///depot/google3/datacommons/proto/prophet/cache_data.proto;l=338?q=StatVarGroups) cache.

### DIFF
--- a/simple/stats/data.py
+++ b/simple/stats/data.py
@@ -439,8 +439,14 @@ class VerticalSpec:
                         verticals=set(json.get("verticals", [])))
 
 
+# dict[svg id, svg specialized name]
+SVGSpecializedNames = dict[str, str]
+
+# dict[parent svg id, child SVGSpecializedNames]
+ParentSVG2ChildSpecializedNames = dict[str, SVGSpecializedNames]
+
+
 @dataclass
 class StatVarHierarchyResult:
   svg_triples: list[Triple]
-  # dict[parent svg id, dict[child svg id, child specialized name]]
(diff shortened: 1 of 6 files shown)


simple/stats/data.py
```diff
@@ -437,3 +437,10 @@ def from_json(json: dict) -> Self:
                         measured_properties=set(
                             json.get("measuredProperties", [])),
                         verticals=set(json.get("verticals", [])))
+
+
+@dataclass
+class StatVarHierarchyResult:
+  svg_triples: list[Triple]
+  # dict[parent svg id, dict[child svg id, child specialized name]]
+  svg_specialized_names: dict[str, dict[str, str]]
```
pradh, line 446: nit: child SVG-id to specialized name need to be a dict?

pradh: For readability, consider type names (like typedefs):

e.g.,
```
SVGSpecializedName = tuple[str, str]
ParentSVG2Children = dict[str, SVGSpecializedName]
```

pradh (approved):

simple/stats/data.py
keyurva: On tuple: It needs to be a dict since there could be multiple children with specialized names.

On type aliases: +1 for readability. Done.

keyurva:
Thanks for the review.

PR #317, opened 2024-05-29
Add and compile svg cache proto.

keyurva:


run_test.sh
```diff
@@ -143,6 +148,7 @@ function help {
   echo "-l              Run lint test"
   echo "-m              Run sample and generate main dc output"
   echo "-p              Run python tests"
+  echo "-protoc         Compile protos"
```
shifucun, line 151: should this be --protoc (otherwise could mean -p & -r & -o)

shifucun (approved):

run_test.sh
keyurva: Good point. Done.

keyurva:
Thanks for the review.

PR #320, opened 2024-05-31
Use common test util functions.

keyurva:


simple/tests/stats/events_importer_test.py
```diff
@@ -38,14 +39,6 @@
...
 
-def _compare_files(test: unittest.TestCase, output_path, expected_path):
-  with open(output_path) as gotf:
-    got = gotf.read()
-    with open(expected_path) as wantf:
-      want = wantf.read()
-      test.assertEqual(got, want)
-
-
 def _write_observations(db_path: str, output_path: str):
```
pradh, line 42: TODO: could _write_triples also be in the helper (I see it dup'd)?

pradh (approved):
Nice!

simple/tests/stats/events_importer_test.py
keyurva: Ack. Will do.

PR #322, opened 2024-05-31
Write SVG cache to DB.

keyurva:
* Also, realized that the root node is needed for mixer to merge it with the main dc hierarchy correctly. So including dc root node in SVG cache now.


simple/stats/cache.py
```diff
@@ -36,6 +39,25 @@ def generate_svg_cache(db: Db,
...
+  return base64.b64encode(compressed_buffer.getvalue()).decode('utf-8')
+
+
+def base64_decode_and_gunzip(encoded_data: str) -> bytes:
+  """Decodes a Base64 string, decompresses the GZIP data and returns the uncompressed bytes."""
+
+  compressed_data = base64.b64decode(encoded_data)
+
+  with gzip.GzipFile(fileobj=io.BytesIO(compressed_data), mode="rb") as gz_file:
+    return gz_file.read()
```
pradh, line 58: nit: maybe worth having this in a utils file.  TODO is fine.

pradh (approved):

simple/stats/cache.py
keyurva: Added TODO.

keyurva (commented):
Thanks for the review.

simple/stats/cache.py
```diff
@@ -27,13 +30,39 @@
 from stats.data import Triple
 from stats.db import Db
 
+STAT_VAR_GROUPS_CACHE_KEY = "StatVarGroups"
```
pradh, line 33: what does this key mean?  Does it need to match something in the mixer?  Please comment.

keyurva: This is the key used to store the SVG hierarchy in the `key_value_store` table. [Examples](https://github.com/search?type=code&q=org%3Adatacommonsorg+statvargroups+language%3Acsv+repo%3Adatacommonsorg%2Fimport)

Yes, [mixer](https://github.com/datacommonsorg/mixer/blob/fae6f51bdce56b669b00a9e5d79314a4dd6087b2/internal/sqldb/sqlquery/key_value.go#L27) uses the same key.

PR #323, opened 2024-06-01
Add support for key_value_store table.

keyurva:


shifucun (approved):
nice! Feels like a separate table makes things simpler.

PR #327, opened 2024-06-26
Update some build trigger configs; add a repo version bump step

hqpho:
- Remove build/cloudbuild.push.yaml. With the addition of build/ci/cloudbuild.push.yaml, it's confusing to have both files. I can't find any scripts or triggers that use it, so should be safe to remove.
- Copy build/cloudbuild.java.yaml to a new build/ci/cloudbuild.push.yaml
- Add a repo version bump step, same as the one added to website in https://github.com/datacommonsorg/website/pull/4386


shifucun (approved):
thanks!

PR #328, opened 2024-06-27
Remove Java checks from new push trigger config

hqpho:
The current trigger (using config cloudbuild.java.yaml) runs on branches other than master as well. Since Java checks already run on presubmit, I am removing them from this trigger's config which runs post-submit. When editing the import-push-java trigger, I'll also change it to run only on pushes to master.

### DIFF
--- a/build/ci/cloudbuild.push.yaml
+++ b/build/ci/cloudbuild.push.yaml
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Cloud build config that is triggered on a push to master.
 # This cloud build is triggered when there is code change, but can also involve
 # data change.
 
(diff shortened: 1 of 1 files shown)


shifucun:
I don't think we need to run on non-master branch though

hqpho:
> I don't think we need to run on non-master branch though

In that case, my follow-up question would be whether we want to have the Java checks duplicated between cloudbuild.java.yaml and cloudbuild.push.yaml, since cloudbuild.java.yaml is also used for a trigger that runs on pull. Do we need to be doing Java checks both before and after PRs?

shifucun:
> > I don't think we need to run on non-master branch though
> 
> In that case, my follow-up question would be whether we want to have the Java checks duplicated between cloudbuild.java.yaml and cloudbuild.push.yaml, since cloudbuild.java.yaml is also used for a trigger that runs on pull. Do we need to be doing Java checks both before and after PRs?

No need to do java checks after PR. In other repo, we only run test before PR submission.

hqpho:
> > > I don't think we need to run on non-master branch though
> > 
> > 
> > In that case, my follow-up question would be whether we want to have the Java checks duplicated between cloudbuild.java.yaml and cloudbuild.push.yaml, since cloudbuild.java.yaml is also used for a trigger that runs on pull. Do we need to be doing Java checks both before and after PRs?
> 
> No need to do java checks after PR. In other repo, we only run test before PR submission.

Ok! Updated the description to reflect this.

shifucun (approved):

PR #334, opened 2024-07-29
Populate entity with type by calling base DC.

keyurva:
* SV summary in CDC currently does not work if the data is provided in variable per row (aka svobs) format.
* This was because we were not populating entities from those files in the sql db.
* With this fix, we fetch entity types for all unique entities from base dc and record them in sql.
* The fix applies this approach to variable per column files as well for consistency.

### DIFF
--- a/simple/stats/nodes.py
+++ b/simple/stats/nodes.py
@@ -264,7 +264,7 @@ class Nodes:
 
   def entities_with_types(self, dcid2type: dict[str, str]):
     """
-    Adds each dcid2type mapping to the list of entity with their types.
+    Adds each dcid2type mapping to the list of entities with their types.
     The full list will be inserted into the DB in the final stages of the import.
     """
     for entity_dcid, entity_type in dcid2type.items():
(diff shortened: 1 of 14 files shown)


simple/stats/nodes.py
```diff
@@ -262,6 +262,10 @@ def entities_with_type(self, entity_dcids: list[str], entity_type: str):
     for entity_dcid in entity_dcids:
       self.entity_with_type(entity_dcid, entity_type)
 
+  def entities_with_types(self, dcid2type: dict[str, str]):
```
dwnoble, line 265: Nit: add a docstring (at first glance it wasn't clear to me this function mutated the object state)

dwnoble (approved):
Thanks Keyur!

simple/stats/nodes.py
keyurva: Done.

keyurva (commented):
Thanks for the review!

PR #337, opened 2024-08-01
Create db indexes programmatically.

keyurva:
* This PR creates DB indexes for better performance as part of loading data itself.
  + Without this, partners would need to create the indexes themselves and we'd have to document it.
* As an optimization, at start of load data, we first drop any existing indexes so that insertions are faster. And we add the indexes only as the final step (after all data has been loaded) before committing the transaction.

### DIFF
--- a/simple/stats/db.py
+++ b/simple/stats/db.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from dataclasses import dataclass
 from datetime import datetime
 from enum import auto
 from enum import Enum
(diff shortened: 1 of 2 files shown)


simple/stats/db.py
```diff
@@ -337,9 +367,22 @@ def __init__(self, db_params: dict) -> None:
     logging.info("Connected to SQLite: %s", self.local_db_file_path)
 
     self.cursor = self.connection.cursor()
+    # Drop indexes first so inserts are faster.
+    self._drop_indexes()
```
dwnoble, line 371: Im not sure it makes sense to do drop indexes in the constructor. i might not be fully understanding, but while i think it works for our current workflow, i could see situations where we'd want to create this object to just run read operations?

keyurva: Keeping the index slows down inserts. So dropping and then readding helps with throughput and ensures the tables are indexed at the end.

dwnoble (approved):

keyurva:
Thanks for the review.

PR #340, opened 2024-09-16
Add support for loading observation properties.

keyurva:
* Adds support for loading observation properties.
* Properties are modeled as a single column in the observations table.
* For existing tables, the script applies schema updates during initialization.
* Design doc: go/cdc-obs-props


keyurva:
Thanks Ajai.

I've created b/368084250 for the per row ask. Let's discuss it there.

PR #341, opened 2024-09-24
Use separate columns for standard observation properties.

keyurva:
* After a conversation with @shifucun and @beets, decided to create separate columns for standard observations properties (namely `unit`, `scaling_factor`, `measurement_method` and `observation_period`) and a catch-all `properties` column for all other properties.
*  Also, the `properties` column values will just be stringified json blobs vs gzipped and base64 encoded as it was previously.
   + @beets correctly noted that the blob is not going to be large in practice so no need to gzip it.
* Changes to the test file `simple/tests/stats/test_data/observations_importer/expected/obs_props/observations.db.csv` are illustrative of the key changes introduced by this PR.


simple/stats/db.py
```diff
@@ -487,9 +501,10 @@ def _schema_updates(self) -> None:
     properties_column_exists = rows is not None and len(rows) > 0
     if not properties_column_exists:
       logging.info(
-          "properties column does not exist in the observations table. Altering table to add it."
+          "properties column does not exist in the observations table. Altering table to add all property columns."
```
beets, line 504: consider spelling out all the columns that are being added to the log statement ("all property columns" is vague)

beets (commented):
sorry for the review. had to poke around more since it's my first time with the code..

simple/stats/data.py
```diff
@@ -234,25 +235,45 @@ class Observation:
...
+  observation_period: str = ""
   properties: dict[str, str] = field(default_factory=dict)
 
   def __post_init__(self):
     if not self.properties:
       self.properties = {}
-    # Properties in the DB are stored as gzipped and base64 encoded strings.
-    # Convert it to json / dict so it is available as a dict in code.
+    # Properties in the DB are stored as stringified json.
+    # Load it as a dict in code.
```
shifucun, line 248: I feel this type conversion is too subtle, should the json.loads happen outside the class object constructor?

simple/stats/data.py
```diff
@@ -234,25 +235,45 @@ class Observation:
...
-    return gzip_and_base64_encode_json(self.properties)
+      self.properties = json.loads(self.properties)
+
+    self.unit = self.unit or self.properties.get(sc.PREDICATE_UNIT, "")
+    self.scaling_factor = self.scaling_factor or self.properties.get(
+        sc.PREDICATE_SCALING_FACTOR, "")
+    self.measurement_method = self.measurement_method or self.properties.get(
+        sc.PREDICATE_MEASUREMENT_METHOD, "")
+    self.observation_period = self.observation_period or self.properties.get(
+        sc.PREDICATE_OBSERVATION_PERIOD, "")
```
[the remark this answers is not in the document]
shifucun: +1 the properties should not have the 4 special props

simple/stats/data.py
```diff
@@ -234,25 +235,45 @@ class Observation:
...
+    self.properties = self._custom_properties()
```
shifucun, line 259: feel doing too many things in the post_init, which can make the behavior too implicit? Why not have a helper function to transform properties and operate outside

simple/stats/data.py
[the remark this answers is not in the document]
keyurva: Added comment.

This file defines data model classes. I think `data.py` or `model.py` works but open to suggestions.

simple/stats/data.py
keyurva: Simplified this class and added a separate one explicitly for `ObservationProperties`.

simple/stats/data.py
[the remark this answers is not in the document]
keyurva: Simplified this class and moved properties extraction to a dedicated `ObservationProperties` class. Note that the latter is still needed since users will specify all properties in a single json object in `config.json` and the standard properties need to be parsed out.

I agree it should not be in `Observation` itself so separated it out.

simple/stats/data.py
keyurva: Agreed. Moved to an explicit `ObservationProperties` class.

simple/stats/data.py
```diff
@@ -234,25 +235,45 @@ class Observation:
...
+        for p, v in all_properties.items()
+        if p not in sc.STANDARD_OBSERVATION_PROPERTIES
+    }
+    return custom_properties
 
   def db_tuple(self):
     return (_strip_namespace(self.entity), _strip_namespace(self.variable),
             self.date, self.value, _strip_namespace(self.provenance),
-            self.properties_string())
+            _strip_namespace(self.unit), self.scaling_factor,
```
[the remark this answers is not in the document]
keyurva: This is in case they specify values like `dcs:USD`, then we strip out the namespace before inserting in the DB.

simple/stats/db.py
```diff
@@ -78,12 +78,16 @@
     date varchar(255),
     value varchar(255),
     provenance varchar(255),
+    unit varchar(255),
+    scaling_factor varchar(255),
```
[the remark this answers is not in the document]
keyurva: It's stored as a [string](https://screenshot.googleplex.com/BZVk47RMBAQSg9R) in base dc so modeling it similarly here.

simple/stats/db.py
```diff
@@ -122,11 +126,17 @@
...
 
-# The properties column was not part of the observations table originally.
-# This statement adds the column.
+# Various property columns not part of the observations table originally.
+# These statements add those columns.
 # Neither sqlite nor mysql support an 'if not exists' statement for altering tables universally,
-# so the code needs to check for existence separately before applying this statement.
-_ALTER_OBSERVATIONS_TABLE_STATEMENT = "alter table observations add column properties TEXT;"
+# so the code needs to check for existence separately before applying these statements.
+_ALTER_OBSERVATIONS_TABLE_STATEMENTS = [
```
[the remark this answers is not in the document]
keyurva: Ack - let me do this in a dedicated PR. Note that the column names are all encompassed in this class only.

simple/stats/db.py
keyurva: Done.

simple/tests/stats/db_test.py
```diff
@@ -46,7 +46,7 @@
 _OBSERVATIONS = [
     Observation("e1", "v1", "2023", "123", "p1"),
     Observation("e2", "v1", "2023", "456", "p1"),
-    Observation("e3", "v1", "2023", "789", "p1", {"prop1": "val1"})
+    Observation("e3", "v1", "2023", "789", "p1", properties={"prop1": "val1"})
```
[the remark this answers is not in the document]
keyurva: Done. Also see `simple/tests/stats/test_data/observations_importer/expected/obs_props/observations.db.csv` which covers importing standard and custom properties.

keyurva (commented):
Thanks for the reviews!

simple/stats/data.py
```diff
@@ -227,35 +230,62 @@ def triples(self) -> list[Triple]:
...
-    return gzip_and_base64_encode_json(self.properties)
+  properties: ObservationProperties = field(
+      default_factory=ObservationProperties.new)
 
   def db_tuple(self):
     return (_strip_namespace(self.entity), _strip_namespace(self.variable),
             self.date, self.value, _strip_namespace(self.provenance),
-            self.properties_string())
+            _strip_namespace(self.properties.unit),
+            self.properties.scaling_factor,
```
shifucun, line 271: why not _strip_namespace for scaling_factor?

simple/stats/db.py
```diff
@@ -59,6 +59,11 @@
 
 MAIN_DC_OUTPUT_DIR = "mainDcOutputDir"
 
+_OBSERVATION_PROPERTY_COLUMNS = [
```
shifucun, line 62: Should here use the const in schema_constants.py?

simple/stats/schema_constants.py
```diff
@@ -34,6 +34,16 @@
 PREDICATE_SEARCH_DESCRIPTION = "searchDescription"
 PREDICATE_MEASURED_PROPERTY = "measuredProperty"
 PREDICATE_DESCRIPTION = "description"
+PREDICATE_UNIT = "unit"
+PREDICATE_SCALING_FACTOR = "scalingFactor"
+PREDICATE_MEASUREMENT_METHOD = "measurementMethod"
+PREDICATE_OBSERVATION_PERIOD = "observationPeriod"
```
shifucun, line 40: (minor, not worth changing for now, and subjective) I feel the constant to too long for readability. Might just remove "PREDICATE_" everywhere

shifucun (approved):

simple/stats/db.py
keyurva: No. `schema_constants.py` has names in the KG. Here's it's the column names and in some cases they are different (e.g. `scalingFactor` vs `scaling_factor`, etc.)

simple/stats/data.py
keyurva: As I understand it, scaling factor is a value and not a reference, hence no namespaces. Correct me if that's not the case.

simple/stats/db.py
shifucun: I see

simple/stats/data.py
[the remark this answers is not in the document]
beets: got it - might be clearer if a) there was a comment on _strip_namespace, or b) called a method _process_dcid which strips the namespace (and does other processing). can be done later.

simple/stats/data.py
```diff
@@ -227,35 +230,62 @@ def triples(self) -> list[Triple]:
...
+            _strip_namespace(self.properties.measurement_method),
+            _strip_namespace(self.properties.observation_period),
+            json.dumps(self.properties.properties)
+            if self.properties.properties else "")
+
+
+def _get_flattened_dataclass_field_names(cls) -> list[str]:
```
beets, line 278: please add a test and comment for this

beets (approved):
thanks for the updates!

simple/stats/data.py
keyurva: Done.

keyurva (commented):
Thanks both for the reviews!

simple/stats/data.py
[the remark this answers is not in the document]
keyurva: Ack.

PR #344, opened 2024-10-22
Support a schema update mode in stats runner

hqpho:
- On Db object creation, continue to ensure database exists and apply any schema updates. Pull these behaviors into named methods for clarity.
- Move clearing indexes and tables in anticipation of an import out of Db constructor and into an explicit separate Db method. Maintain existing behavior for imports by having Runner call this method before importing.
- Add a mode value `schemaupdate` to both RunMode and run_stats.sh mode arg. The cdc_data docker image can pass this in based on an environment variable: https://github.com/datacommonsorg/website/pull/4686
- Add test cases for both Db and Runner that start with an existing SQL database loaded from a file.


simple/tests/stats/runner_test.py
```diff
@@ -125,6 +133,12 @@ def test_config_with_wildcards(self):
   def test_input_dir_driven(self):
     _test_runner(self, "input_dir_driven", is_config_driven=False)
 
+  def test_input_dir_driven_with_existing_old_schema_data(self):
+    _test_runner(self,
+                 "input_dir_driven_with_existing_old_schema_data",
+                 is_config_driven=False,
+                 input_db_file_name="sqlite_old_schema_populated.db")
```
keyurva, line 140: With binary `*.db` files, the tests can feel opaque. Is it possible to have the tests take csv files as input and compare against expected csv files?

keyurva (commented):
The general direction of the PR is perfect!

Is it possible to break up this PR into 2 smaller PRs - one with changes primarily to the DB and a follow-up PR that introduces the new mode?

simple/tests/stats/runner_test.py
hqpho: Sure, I can do that. Before I do so, are you able to review the non-test portions of the PR so I can make test updates all in one if necessary? Also so I can work during the time zone offset.

keyurva: Yes. The overall implementation looks good, added a couple of nits.

hqpho:
> The general direction of the PR is perfect!
> 
> Is it possible to break up this PR into 2 smaller PRs - one with changes primarily to the DB and a follow-up PR that introduces the new mode?

If this is for ease of reviewing, OK if I restructure the commits within the PR instead? If it's for ease of bisecting issues if any come up, I can split the PR instead. Having dependent PRs is a headache with squash-merges, so would prefer to keep this all in one PR if possible.

keyurva:
> so would prefer to keep this all in one PR if possible

ok. Let's keep this one as-is. For the future, let's try to break them up into smaller PRs.

(and a personal note to me: do reviews more promptly!)

simple/tests/stats/runner_test.py
keyurva: Oh - this totally works. It's text based and a truer representation of a DB vs CSVs. 

Didn't know about the `iterdump()` utility. Seems like the exact thing we need here.

simple/stats/runner.py
```diff
@@ -137,35 +138,49 @@ def _get_db_config() -> dict:
 
   def run(self):
     try:
-      # Run all data imports.
-      self._run_imports()
+      if self.mode == RunMode.SCHEMA_UPDATE:
```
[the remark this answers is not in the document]
hqpho: Added call to init_or_update. Used "storage" instead of "tables" since it's not necessarily tables (updated the clear method to use "storage" as well). Opted not to add _run_schema_update_mode since it makes it harder to tell what operations are shared and right now IMO there is not so much divergence that this method is hard to read. Can revisit in the future if there is more divergence!

simple/stats/db.py
```diff
@@ -285,8 +291,13 @@ class SqlDb(Db):
 
   def __init__(self, config: dict) -> None:
     self.engine = create_db_engine(config)
+    self.engine.init_or_update_tables()
```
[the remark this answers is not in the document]
hqpho: Hmm, so I tried this but it changed an unexpectedly large surface and seemed to make it easier to "hold it wrong" when creating any Db, even if it's not a SqlDb. I have tried instead to make it very clear with naming and comments what is happening. WDYT?

keyurva (approved):
Thanks for the updates!

hqpho:
> Thanks for the updates!

Thanks for the review – promise to break it up into smaller pieces next time!

PR #346, opened 2024-10-28
Add support for specifying obs properties per row.

keyurva:
* Note that this is only supported for the `variablePerRow` format at this time.


dwnoble (approved):
Thanks Keyur!

PR #347, opened 2024-10-29
Add support for generating topic cache.

keyurva:
* For an example topic cache json, refer [this one](https://github.com/datacommonsorg/website/blob/204ee2668d6e800f1a554c60066057d638f424ad/server/tests/lib/test_data/topic_cache/input/merge_with_undata/datacommons/nl/custom_dc_topic_cache.json#L4).
* The generated json does not include the `dc` field, so that it is merged with the main dc cache itself. [reference](https://github.com/datacommonsorg/website/blob/204ee2668d6e800f1a554c60066057d638f424ad/server/lib/topic_cache.py#L238)
* Currently the cache is only generated for topics and not for SV peer groups.
  + @dwnoble - let me know if this is needed and I can add it in a follow-up PR.

### DIFF
--- a/simple/stats/nl.py
+++ b/simple/stats/nl.py
@@ -14,6 +14,7 @@
 
 from dataclasses import dataclass
 from dataclasses import field
+import json
 import logging
 
 import pandas as pd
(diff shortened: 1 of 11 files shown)


simple/stats/nl.py
```diff
@@ -120,3 +150,41 @@ def sentences(self) -> str:
       sentences = [self.name]
 
     return _SENTENCE_SEPARATOR.join(sentences)
+
+
+@dataclass
+class TopicCacheNode:
```
dwnoble, line 156: Thanks for defining the dataclass!

PR #349, opened 2024-10-31
Include stat var peer groups in topic cache.

keyurva:


dwnoble (approved):
Thanks Keyur!

keyurva:
Thanks for the review!

PR #351, opened 2024-11-05
Add utils for PyFilesystem and config key matching

hqpho:
This lays the foundation for [supporting multiple input dirs](https://github.com/datacommonsorg/import/pull/352).

### DIFF
--- a/simple/tests/util/file_match_test.py
+++ b/simple/tests/util/file_match_test.py
@@ -1,3 +1,4 @@
+import tempfile
 import unittest
 
 from util.file_match import match
@@ -22,6 +23,7 @@ class TestFileMatch(unittest.TestCase):
         match(file, pattern)
 
     # File path: temp://path/to/foo.csv
+    # Absolute path: /<generated>/path/to/foo.csv
 
     # No slashes: match file name
     yes("foo.csv")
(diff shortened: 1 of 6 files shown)


simple/tests/util/file_match_test.py
```diff
@@ -0,0 +1,124 @@
...
+    yes("//*/**/foo.csv")
+    yes("//**/*/*/foo.csv")
+    yes("//*/**/*/foo.csv")
+    yes("//**/*/**/*/**/foo.csv")  # gettin silly with it
+    no("//**/*/**/*/**/*/**/foo.csv")
+    no("//*/**/*/*/foo.csv")
+    no("//**/*/*/*/foo.csv")
+
+    # Leading single slash - equivalent to no leading slash
+    yes("/foo.csv")  # File is not in root dir
```
dwnoble, line 60: Should a single leading slash be an absolute path? I could see users testing locally and wanting to load data from a known location somewhere else on the file system

hqpho: I'm inclined to use leading double slash to match absolute path instead, for consistency with non-system paths e.g. gs://bucket/some/dir. But we can use the single vs double slash distinction for whether to match the full absolute path or the full relative path (relative to the root of the input dir). WDYT?

simple/util/file_match.py
```diff
@@ -0,0 +1,126 @@
...
+      - Double wildcards for name or protocol are invalid; use single wildcards instead.
+    - A single wildcard (*) in the directory portion of a pattern can be any non-slash character.
+    - A double wildcard (**) in the directory portion of a pattern can be any character including slashes, i.e. multiple dir levels.
+
+    Since matching is ultimately done with regular expressions, it may be
+    possible to do more complex matching by passing in patterns containing special
+    characters, but this behavior is not officially supported and may break at any time.
+
+    Examples:
+    - Match any CSV file in any directory or subdirectory: "*.csv"
```
keyurva, line 36: Will this change the current default behavior where it does not look for files in sub directories? If it does, it may pick up files that were not previously picked.

On the larger point of sub-directories, are you thinking it will be an opt-in (via a config field) or will that be the default moving forward?

hqpho: Yes, this would change that default behavior. For that reason, we would also need to stop allowing output dir to overlap with input dir. I imagined this would be the default moving forward.

keyurva (approved):

simple/tests/util/file_match_test.py
dwnoble: That's fine with me as long as it's clear to the user.  The word root is used in a few places in the comments which was throwing me off. Maybe we can use "Filesystem root" for a leading `//`, and "Relative to the root of the input dir" (like you have worded in your comment above) for a leading `/`?

hqpho: Done, and updated comments that reference "root" too!

PR #352, opened 2024-11-07
Migrate to PyFilesystem and support multiple input dirs

hqpho:
- Replace all uses of FileHandler with new PyFilesystem wrapper classes.
- Update config key lookups to use new file matching util. This allows specifying matching by either file name (already supported) or path (new), with support for wildcards.
- Put input subdirectory traversal behind an environment variable `INCLUDE_INPUT_SUBDIRS` which defaults to false.

Note that this PR also temporarily contains changes that are in https://github.com/datacommonsorg/import/pull/351. I will rebase this branch after that one is merged.


keyurva:
Will review this once it is rebased on #351. Is there a diffbase like utility for git?

hqpho:
> Will review this once it is rebased on #351. Is there a diffbase like utility for git?

I don't know of one. In the future I plan to create stacked PRs using Graphite which will take care of that, but this one was already created when I installed Graphite.

If it helps, there are no additional changes in this PR to any of the files changed in #351.

dwnoble:
Also when running an import i see log messages like:

```
Importing file: myfile.csv
```

but the file is located at `datasource/data/myfile.csv` - can you update the logging to show the full path?

dwnoble:
> Also when running an import i see log messages like:
> 
> ```
> Importing file: myfile.csv
> ```
> 
> but the file is located at `datasource/data/myfile.csv` - can you update the logging to show the full path?

Same for MCF files

Example:
```
I1113 23:34:47.937138 281473247215648 runner.py:317] Importing MCF file: custom_topics.mcf
```

but the file is at `datasource/schema/custom_topics.mcf`

dwnoble (changes requested):
(see comments above)

simple/stats/runner.py
```diff
@@ -57,62 +48,69 @@ class Runner:
...
-    self.process_dir_fh = self.output_dir_fh.make_file(
-        f"{constants.PROCESS_DIR_NAME}/")
-
-    self.output_dir_fh.make_dirs()
-    self.nl_dir_fh.make_dirs()
-    self.process_dir_fh.make_dirs()
+    output_store = create_store(output_dir_path, create_if_missing=True)
+    if self.include_input_subdirs:
+      for input_store in self.input_stores:
+        _check_not_overlapping(input_store, output_store)
```
keyurva, line 105: Nice check! Perhaps log a warning that they should keep input and output dirs separate.

simple/stats/runner.py
```diff
@@ -261,100 +270,111 @@ def _vertical_specs_dcids(self,
...
-    logging.info("Importing file: %s", input_fh.basename())
-    self._create_importer(input_fh).do_import()
-
-  def _run_single_mcf_import(self, input_mcf_fh: FileHandler):
-    logging.info("Importing MCF file: %s", input_mcf_fh.basename())
-    self._create_mcf_importer(input_mcf_fh, self.output_dir_fh,
+        input_files.append(input_store.as_file())
+
+    for input_file in input_files:
+      self._check_if_special_file(input_file)
```
keyurva, line 295: You probably want _check_if_special_file to return a bool and not do the other matches if it is a special file. Otherwise, if the special file is a csv or a mcf file (not the case today but future proofing), then it could be problematic.

simple/stats/runner.py
```diff
@@ -261,100 +270,111 @@ def _vertical_specs_dcids(self,
...
-  # If the fh is a "special" file, append it to the self.special_handlers dict.
-  # Returns true if it is, otherwise false.
-  def _maybe_set_special_fh(self, fh: FileHandler) -> bool:
-    file_name = fh.basename()
-    file_type = self.special_file_to_type.get(file_name)
-    if file_type:
-      self.special_handlers[file_type] = fh
-      return True
-    return False
+  def _check_if_special_file(self, file: File) -> None:
```
keyurva, line 273: This method is slightly more expensive than its predecessor but should be ok in practice.

simple/stats/runner.py
```diff
@@ -261,100 +270,111 @@ def _vertical_specs_dcids(self,
...
+
+    for input_file in input_files:
+      self._check_if_special_file(input_file)
+      if match(input_file, "*.csv"):
+        input_csv_files.append(input_file)
+      if match(input_file, "*.mcf"):
+        input_mcf_files.append(input_file)
+
+    # Sort input files alphabetically.
+    input_csv_files.sort(key=lambda f: f.name())
```
keyurva, line 302: nit: how about sorting by path? (it's a subjective call so optional)

simple/stats/runner.py
```diff
@@ -261,100 +270,111 @@ def _vertical_specs_dcids(self,
...
-    mcf_file_name = input_fh.basename()
-    output_fh = output_dir_fh.make_file(mcf_file_name)
-    reporter = self.reporter.import_file(mcf_file_name)
-    return McfImporter(input_fh=input_fh,
-                       output_fh=output_fh,
+    # Right now, this overwrites any file with the same name,
+    # so if different input sources have files with the same relative path,
+    # they will clobber each others output. Treating this as an edge case
+    # for now but could resolve by allowing input sources to be mapped to output
+    # locations.
```
keyurva, line 327: Perhaps add a comment that this only applies to main dc so ok to fixing it later.

keyurva: One idea is to use the sanitized file path approach that you have used for debug files.

simple/stats/runner.py
```diff
@@ -261,100 +270,111 @@ def _vertical_specs_dcids(self,
...
                        is_main_dc=is_main_dc)
 
-  def _create_importer(self, input_fh: FileHandler) -> Importer:
-    input_file = input_fh.basename()
+  def _create_importer(self, input_file: File) -> Importer:
     import_type = self.config.import_type(input_file)
-    debug_resolve_fh = self.process_dir_fh.make_file(
-        f"{constants.DEBUG_RESOLVE_FILE_NAME_PREFIX}_{input_file}")
-    reporter = self.reporter.import_file(input_file)
+    sanitized_path = input_file.full_path().replace("://",
```
keyurva, line 338: Creative!

simple/stats/runner.py
```diff
@@ -261,100 +270,111 @@ def _vertical_specs_dcids(self,
...
+    raise ValueError(
+        f"Unsupported import type: {import_type} ({input_file.full_path()})")
+
+
+def _check_not_overlapping(input_store: Store, output_store: Store):
+  input_path = input_store.full_path()
+  output_path = output_store.full_path()
+  if fspath.issamedir(input_path, output_path) or fspath.isparent(
+      input_path, output_path) or fspath.isparent(output_path, input_path):
+    raise ValueError(
```
keyurva, line 379: Is it possible to log a warning and ignore such directories (for input files)?

simple/stats/runner.py
```diff
@@ -57,62 +48,69 @@ class Runner:
...
-    assert config_file or input_dir, "One of config_file or input_dir must be specified"
-    assert output_dir, "output_dir must be specified"
+    assert config_file_path or input_dir_path, "One of config_file or input_dir must be specified"
+    assert output_dir_path, "output_dir must be specified"
 
     self.mode = mode
-    self.input_handlers: list[FileHandler] = []
+
+    # New option to traverse subdirs of input dir(s).
+    self.include_input_subdirs = bool(os.getenv("INCLUDE_INPUT_SUBDIRS"))
```
keyurva, line 61: Thoughts on basing it on a property in config.json vs in an env variable?

keyurva (commented):
Super cool!

This is a deep change touching a significant portion of the codebase and really well done. Once in autopush, let's thoroughly test all combinations (cloud, local, sqlite, cloud sql, etc.) before doing a stable release.

simple/stats/runner.py
hqpho: IMO it's quicker for a user to fix a broken setup if we fail immediately rather than allowing an import to finish but with missing data. WDYT?

simple/stats/runner.py
hqpho: I chose an env variable because it's similar to the approach we use for database config (e.g. `ENV_USE_CLOUDSQL`). Is there a principle you'd recommend for deciding what goes in config.json vs an env variable?

hqpho:
> If i run the importer with .csv files in subdirectories that are not matched by the inputFiles mappings, the data importer will still try to import them using a one-variable-per-column approach. Instead, can we ignore these files?

This would be a separate change to existing behavior, right? (Right now if there is a .csv file in the base directory that doesn't have a match in the inputFiles mapping, is it imported with a one-variable-per-column approach? I'm assuming yes)

dwnoble:
> > If i run the importer with .csv files in subdirectories that are not matched by the inputFiles mappings, the data importer will still try to import them using a one-variable-per-column approach. Instead, can we ignore these files?
> 
> This would be a separate change to existing behavior, right? (Right now if there is a .csv file in the base directory that doesn't have a match in the inputFiles mapping, is it imported with a one-variable-per-column approach? I'm assuming yes)

Hmm yes that is right. @keyurva wdyt about requiring us to explicitly specify what we're importing?

hqpho:
> Also when running an import i see log messages like:
> 
> ```
> Importing file: myfile.csv
> ```
> 
> but the file is located at `datasource/data/myfile.csv` - can you update the logging to show the full path?

Done, here and elsewhere. The __str__ method of File returns full path, so I changed logging to just use the file object.

hqpho:
> Super cool!
> 
> This is a deep change touching a significant portion of the codebase and really well done. Once in autopush, let's thoroughly test all combinations (cloud, local, sqlite, cloud sql, etc.) before doing a stable release.

Agreed! I also think we should let partners know this is basically a beta feature and it might have rough edges.

simple/stats/config.py
```diff
@@ -177,42 +179,37 @@ def generate_hierarchy(self) -> bool:
...
 
   def generate_topics(self) -> bool:
     return self.data.get(_GENERATE_TOPICS) or False
 
-  def _input_file(self, input_file_name: str) -> dict:
-    # Exact match.
-    input_file_config = self._input_files_config.get(input_file_name, {})
-    if input_file_config:
-      return input_file_config
+  def _per_file_config(self, input_file: File) -> dict:
```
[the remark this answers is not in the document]
hqpho: Added a comment!

simple/stats/runner.py
```diff
@@ -57,62 +48,69 @@ class Runner:
...
+
+    # File systems, both input and output. Must be closed when run finishes.
+    self.stores: list[Store] = []
```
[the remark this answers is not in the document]
hqpho: done

simple/stats/runner.py
```diff
@@ -57,62 +48,69 @@ class Runner:
...
-      self.input_handlers.append(input_dir_fh)
+      input_store = create_store(input_dir_path)
+      self.stores.append(input_store)
+      self.input_stores.append(input_store)
 
-      config_fh = input_dir_fh.make_file(constants.CONFIG_JSON_FILE_NAME)
-      if not config_fh.exists():
-        raise FileNotFoundError("Config file must be provided.")
-      self.config = Config(data=json.loads(config_fh.read_string()))
+      config_file = input_store.as_dir().open_file(
```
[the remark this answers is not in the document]
hqpho: Yes! and vice versa for as_file()

simple/stats/runner.py
hqpho: Updated comment

dwnoble:
> Member

To expand on this, I'm in favor of us explicitly specifying all input paths. If we want to keep backwards compatibility, maybe we can put in a configuration option to only read files specified by inputFiles.

simple/stats/runner.py
keyurva: A good demarcation is values that are used by the services container vs only by the importer. This falls in the latter category so leaning towards config.json.

simple/stats/runner.py
keyurva: Hmm - I like sorting just so that the order of processing is deterministic.

simple/stats/runner.py
keyurva: SG. Since this is only when including subdirs, which is new functionality, we can go with this.

Let's document this requirement alongside the doc for the new parameter.

keyurva:
> > > If i run the importer with .csv files in subdirectories that are not matched by the inputFiles mappings, the data importer will still try to import them using a one-variable-per-column approach. Instead, can we ignore these files?
> > 
> > 
> > This would be a separate change to existing behavior, right? (Right now if there is a .csv file in the base directory that doesn't have a match in the inputFiles mapping, is it imported with a one-variable-per-column approach? I'm assuming yes)
> 
> Hmm yes that is right. @keyurva wdyt about requiring us to explicitly specify what we're importing?

This can break existing deployments so we should keep the existing behavior.

keyurva (approved):

simple/stats/runner.py
hqpho: actually, nevermind! it's fine, the test I was seeing break wasn't an existing test.

dwnoble:
> > > > If i run the importer with .csv files in subdirectories that are not matched by the inputFiles mappings, the data importer will still try to import them using a one-variable-per-column approach. Instead, can we ignore these files?
> > > 
> > > 
> > > This would be a separate change to existing behavior, right? (Right now if there is a .csv file in the base directory that doesn't have a match in the inputFiles mapping, is it imported with a one-variable-per-column approach? I'm assuming yes)
> > 
> > 
> > Hmm yes that is right. @keyurva wdyt about requiring us to explicitly specify what we're importing?
> 
> This can break existing deployments so we should keep the existing behavior.

SG for this PR.

I still think it would be a good idea to add a flag that if turned on only reads in data files that are specified by the path expression.

dwnoble (approved):

keyurva:
> I still think it would be a good idea to add a flag that if turned on only reads in data files that are specified by the path expression.

I'm good with that.

PR #353, opened 2024-12-09
Update summary stats generator in the import tool

vish-cs:
*  Add hooks to generate CSV stats summary
*  Add min/max value stats to summary report


util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
```diff
@@ -47,6 +48,8 @@ public static final class StatVarSummary {
     Set<String> observationPeriods = new HashSet<>();
     // This set will not be populated when StatVarSummary is generated by PlaceSeriesSummary
     Set<String> dates = new HashSet<>();
+    double minValue = Double.MAX_VALUE;
+    double maxValue = Double.MIN_VALUE;
```
ajaits, line 52: If all values for an SV are measurementResult (not a number), would the summary have these initial values?
Should we reset to empty in the final report if they have the initial values?

ajaits (approved):

util/src/main/java/org/datacommons/util/CSVReportWriter.java
```diff
@@ -0,0 +1,62 @@
...
+                  summary.numObservations,
+                  summary.minValue,
+                  summary.maxValue,
+                  summary.dates.size(),
+                  !summary.dates.isEmpty()
+                      ? ((TreeSet<String>) (summary.getUniqueDates())).first()
+                      : "",
+                  !summary.dates.isEmpty()
+                      ? ((TreeSet<String>) (summary.getUniqueDates())).last()
+                      : "",
```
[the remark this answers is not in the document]
vish-cs: Done

util/src/main/java/org/datacommons/util/SummaryReportGenerator.java
vish-cs: I am updated the code to return a NaN value for such cases.

PR #357, opened 2025-01-03
fixed a bug where data imports failed if the GCS input path was not initialized with an empty blob.

dwnoble:
When running a data import, if the GCS input bucket wasn't created as an empty blob, the import will fail with the error:

```
FileNotFoundError: File not found: gs://YOUR_BUCKET/input
```

* We have a helper method `_fix_gcsfs_storage` that initializes "subdirectories" under the GCS input path as empty blobs, but if the input directory itself was not an empty blob the import fails.
* This PR moves the call to `_fix_gcsfs_storage` to earlier in the execution flow to account for the case where the input directory itself is not an empty blob.

Additional context b/386254730

### DIFF
--- a/simple/util/filesystem.py
+++ b/simple/util/filesystem.py
@@ -61,11 +61,14 @@ class Store:
 
     if not treat_as_file:
       try:
-        # Set strict=False in case the root_path is not initialized with an empty blob.
-        self.fs = open_fs(f"{self.root_path}?strict=False", create=create_if_missing)
-        # Fix GCS storage if needed.
+        # If the path is a GCS path, set strict=False in case the root_path is not initialized with an empty blob.
         if self.root_path.startswith(_GCS_PATH_PREFIX):
+          self.fs = open_fs(f"{self.root_path}?strict=False", create=create_if_missing)
+          # Fix GCS storage if needed.
           _fix_gcsfs_storage(gcs_fs=self.fs)
+        else:
+          self.fs = open_fs(self.root_path, create=create_if_missing)
+
         self._wrapper: _StoreWrapper = Dir(self, path="/")
         self._isdir = True
       except fserrors.CreateFailed as e:
@@ -221,7 +224,7 @@ class File(_StoreWrapper):
     dest.write_bytes(self.read_bytes())
 
 
-def _fix_gcsfs_storage(gcs_fs: GCSFS) -> None:
+def _fix_gcsfs_storage(fs: GCSFS) -> None:
   """Utility function that walks the entire `root_path` and makes sure that all intermediate directories are correctly marked with empty blobs.
 
   As GCS is no real file system but only a key-value store, there is also no concept of folders. S3FS and GCSFS overcome this limitation by adding
@@ -231,7 +234,7 @@ def _fix_gcsfs_storage(gcs_fs: GCSFS) -> None:
   """
   names = [
       blob.name
-      for blob in gcs_fs.bucket.list_blobs(prefix=fspath.forcedir(gcs_fs.root_path))
+      for blob in fs.bucket.list_blobs(prefix=fspath.forcedir(fs.root_path))
   ]
   marked_dirs = set()
   all_dirs = set()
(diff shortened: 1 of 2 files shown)


simple/util/filesystem.py
```diff
@@ -29,15 +29,29 @@
...
 from fs import open_fs
 import fs.errors as fserrors
 import fs.path as fspath
 from fs_gcsfs import GCSFS
+from urllib.parse import urlparse
 
 # Paths with this prefix refer to Google Cloud Storage.
 _GCS_PATH_PREFIX = "gs://"
 
+def split_path(path: str) -> tuple[str, str]:
```
keyurva, line 43: nit: s/split_path/split_gcs_path?

simple/util/filesystem.py
```diff
@@ -29,15 +29,29 @@
...
+  """
+  Splits a GCS path into a bucket name and a relative path.
+
+  Example:
+    gs://bucket/path/to/file -> ("bucket", "path/to/file")
+
+  Returns:
+    A tuple (bucket_name, relative_path).
+  """
+  parsed_url = urlparse(path)
```
keyurva, line 53: Should we assert that the scheme is `gs`?

simple/util/filesystem.py
```diff
@@ -55,17 +69,21 @@ class Store:
...
+    if self.root_path.startswith(_GCS_PATH_PREFIX):
+      _fix_gcsfs_storage(gcs_uri=self.root_path)
+
     if not treat_as_file:
       try:
         self.fs = open_fs(self.root_path, create=create_if_missing)
         self._wrapper: _StoreWrapper = Dir(self, path="/")
         self._isdir = True
       except fserrors.CreateFailed:
+        logging.info(f"CreateFailed exception: {str(fserrors.CreateFailed)}")
```
keyurva, line 86: This would log the `fserrors.CreateFailed` class. Can we log the error instance instead?

Also, should we log the effect of this error - i.e. treating the path as a file path?

simple/util/filesystem.py
```diff
@@ -219,14 +235,17 @@ def copy_to(self, dest: "File"):
...
   """Utility function that walks the entire `root_path` and makes sure that all intermediate directories are correctly marked with empty blobs.
 
   As GCS is no real file system but only a key-value store, there is also no concept of folders. S3FS and GCSFS overcome this limitation by adding
   empty files with the name "<path>/" every time a directory is created, see https://fs-gcsfs.readthedocs.io/en/latest/#limitations.
 
   This is the same as GCSFS.fix_storage() but with a fix for an infinite loop when root_path does not end with a slash.
   """
+  bucket_name, relative_path = split_path(gcs_uri)
+  # Set strict=False in case the root_path is not initialized with an empty blob.
+  fs = GCSFS(bucket_name=bucket_name, root_path=relative_path, strict=False)
```
keyurva, line 248: One concern is that this will open a GCSFS twice - once in the store's constructor and once here.

Can we instead call `open_fs` in the constructor itself with a [strict=False parameter](https://screenshot.googleplex.com/9dEntqYUJgtpD9K) and leave this method unchanged?

keyurva (commented):
Thanks Dan!

simple/util/filesystem.py
dwnoble: That's a good question- I tried to do open_fs("gs://bucket/path?strict=false"), but it looks like the library doesn't properly pass the strict parameter to GCSFS. See https://github.com/Othoz/gcsfs/issues/11

I'll add a comment clarifying why the separate GCSFS is created

simple/util/filesystem.py
dwnoble: done

simple/util/filesystem.py
dwnoble: Removed this method entirely

simple/util/filesystem.py
dwnoble: After our offline discussion - i was able to use the open_fs call after all!

simple/util/filesystem.py
dwnoble: Good catch. updated & added more context

keyurva (approved):
Thanks for the updates!

PR #364, opened 2025-02-03
When `REDIS_HOST` is set, flush the Redis cache on import

hqpho:
- Don't flush in schema update mode since data doesn't change
- Allow setting custom port as well


simple/stats/db_cache.py
```diff
@@ -0,0 +1,63 @@
...
+    self.connection = redis.Redis(host=host, port=port, decode_responses=True)
+    self.host = host
+    self.port = port
+    self.connection.ping()
+
+  def clear(self):
+    """Flushes all data from the Redis instance used as a database cache."""
+    logging.info("Running FLUSHDB on Redis instance with host=%s and port=%s",
+                 self.host, self.port)
+    self.connection.flushdb(asynchronous=False)
```
keyurva, line 48: nit: Add a log statement after this so users know it is done and in case it takes long they know it's because of this.

keyurva, line 48: I think the `asynchronous` flag here is about running it in the background on the client. Is there a fire-and-forget mode where you issue the flush call to redis but don't need to wait for it to complete?

simple/stats/db_cache.py
```diff
@@ -0,0 +1,63 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+
+import redis
+
+ENV_REDIS_HOST = "REDIS_HOST"
```
keyurva, line 20: Not for this PR but what are your thoughts on assembling all env variable names and defaults into a separate file to easily tell all the env vars used by the importer?

simple/stats/svg_cache.py
keyurva, line 1: It's ok to go with the new name but just wanted to mention that the intention of this file was to include methods that deal with any "cache" entries. For historical context, we initially were calling this a "cache" but eventually landed on `key_value_store` but somehow did not call this file that. You can consider using that name now or later if / when we generate new values in the `key_value_store`.

simple/tests/stats/db_cache_test.py
```diff
@@ -0,0 +1,81 @@
...
+  @patch.dict(os.environ, {ENV_REDIS_PORT: "unused"})
+  def test_get_db_cache_redis_no_host(self):
+    with patch("redis.Redis"):  # Mock redis.Redis to avoid actual connection
+      cache = get_db_cache_from_env()
+      self.assertIsNone(cache)
+
+  @patch.dict(os.environ, {ENV_REDIS_HOST: "localhost"})
+  def test_get_db_cache_redis_no_connection(self):
+    server = FakeServer()
+    server.connected = False
```
keyurva, line 67: These are some amazing low level tests!

simple/tests/stats/db_cache_test.py
```diff
@@ -0,0 +1,81 @@
...
+  @patch.dict(os.environ, {ENV_REDIS_HOST: "localhost"})
+  def test_get_db_cache_redis_clear(self):
+    fake_redis = FakeRedis()
+    fake_redis.set("somekey", "somevalue")
+    self.assertEqual(1, len(fake_redis.keys("*")))
+    with patch("redis.Redis", return_value=fake_redis):
+      cache = get_db_cache_from_env()
+      self.assertIsInstance(cache, RedisDbCache)
+      cache.clear()
+      self.assertEqual(0, len(fake_redis.keys("*")))
```
keyurva, line 81: Did I say these are amazing tests!

simple/tests/stats/runner_test.py
```diff
@@ -207,3 +209,26 @@ def test_with_subdirs_included(self):
...
+  def test_with_redis_db_cache_schema_update(self):
+    fake_redis = FakeRedis()
+    fake_redis.set("somekey", "somevalue")
+    self.assertEqual(1, len(fake_redis.keys("*")))
+    with mock.patch("redis.Redis", return_value=fake_redis):
+      _test_runner(self,
+                   "empty",
+                   output_dir_name="empty_initialized_db",
+                   run_mode=RunMode.SCHEMA_UPDATE)
+      # Redis cache should NOT be cleared in schema update mode.
```
keyurva, line 233: While this makes sense, I'm wondering if it's easier to communicate to the user that if you specify a redis host and run the data loader, it will clear the cache regardless of mode?

simple/stats/db_cache.py
dwnoble: I think we'd still want to wait for the flush call to complete in case it errored out for some reason

simple/tests/stats/runner_test.py
dwnoble: That could also get a little confusing because the environment variables get set when the cloud run job is created, meaning then if the user wanted to run schemaupdate without clearing redis they would have to explicitly un-set the redis_host value.

dwnoble (approved):

simple/stats/db_cache.py
hqpho: I think it's a great idea!

simple/stats/db_cache.py
hqpho: done!

PR #365, opened 2025-02-28
Added new db indices to improve performance of getStatVarSummaries query

dwnoble:
- Improves performance from 30s+ to <200ms
- @keyurva it turns out no SQL query changes are necessary. The query we were experimenting with the other day ended up not being exactly right, but it looks like these indices solve the performance issue
- Tested locally against undata-staging instance

Before:
```
2025/02/28 13:41:49 util.go:474: SQL: GetSVSummaries: 34.594636042s
```

After:
```
2025/02/28 13:43:55 util.go:474: SQL: GetSVSummaries: 141.155ms
```

### DIFF
--- a/simple/stats/db.py
+++ b/simple/stats/db.py
@@ -191,8 +191,8 @@ _DB_INDEXES: list[DbIndex] = [
             indexed_columns=["subject_id", "predicate"]),
     # Used by getStatVarSummaries
     DbIndex(table_name="observations",
-            index_name="observations_variable_value",
-            indexed_columns=["variable", "value"]),
+            index_name="observations_variable",
+            indexed_columns=["variable"]),
 ]
 
 
--- a/simple/stats/db.py
+++ b/simple/stats/db.py
@@ -184,7 +184,15 @@ _DB_INDEXES: list[DbIndex] = [
             indexed_columns=["entity", "variable"]),
     DbIndex(table_name="triples",
             index_name="triples_subject_id",
-            indexed_columns=["subject_id"])
+            indexed_columns=["subject_id"]),
+    # Used by getStatVarSummaries
+    DbIndex(table_name="triples",
+            index_name="triples_subject_id_predicate",
+            indexed_columns=["subject_id", "predicate"]),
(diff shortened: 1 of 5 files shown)


simple/stats/db.py
```diff
@@ -184,7 +184,15 @@ def mysql_create_index_statement(self) -> str:
...
-            indexed_columns=["subject_id"])
+            indexed_columns=["subject_id"]),
+    # Used by getStatVarSummaries
+    DbIndex(table_name="triples",
+            index_name="triples_subject_id_predicate",
+            indexed_columns=["subject_id", "predicate"]),
+    # Used by getStatVarSummaries
+    DbIndex(table_name="observations",
+            index_name="observations_variable_value",
+            indexed_columns=["variable", "value"]),
```
keyurva, line 195: I'm not sure about indexing values. Does a variable only index suffice?

dwnoble: Good catch. it looks like only indexing variable gives similar performance. I dropped value from the index

simple/stats/db.py
```diff
@@ -184,7 +184,15 @@ def mysql_create_index_statement(self) -> str:
             indexed_columns=["entity", "variable"]),
     DbIndex(table_name="triples",
             index_name="triples_subject_id",
-            indexed_columns=["subject_id"])
+            indexed_columns=["subject_id"]),
+    # Used by getStatVarSummaries
+    DbIndex(table_name="triples",
+            index_name="triples_subject_id_predicate",
+            indexed_columns=["subject_id", "predicate"]),
```
keyurva, line 191: Not for this PR but I'm wondering with this subject + predicate index, do we need the subject only index?

@hqpho

keyurva (approved):
Thanks Dan!

simple/stats/db.py
dwnoble: That would make sense to me

simple/stats/db.py
dwnoble: That would make sense to me

PR #368, opened 2025-04-02
Handle self-edges for graph ingestion

vish-cs:


pipeline/ingestion/src/main/java/org/datacommons/CacheReader.java
```diff
@@ -75,14 +75,19 @@ private static List<Entity> parseEntityCache(String row) {
...
               String subjectId, objectId;
-              if (row.startsWith("d/m/")) {
+              // Add a self edge if value is populated.
+              if (entity.getValue().isEmpty()) {
                 subjectId = dcid;
-                objectId = entity.getDcid();
-              } else { // "d/l/"
-                subjectId = entity.getDcid();
                 objectId = dcid;
+              } else {
```
n-h-diaz, line 83: (optional: can maybe flatten to reduce nesting)

n-h-diaz (approved):
Thanks!

pipeline/ingestion/src/main/java/org/datacommons/CacheReader.java
vish-cs: Keeping them as is for logical separation.

PR #371, opened 2025-04-03
Write distinct entities to spanner.

keyurva:


n-h-diaz (commented):
Have you tried with this change? How does the performance compare?

keyurva:
Ran into a FK issue with this approach: https://screenshot.googleplex.com/A3N55kM7C2WfKPX

I relooked at the structure of Entity and it makes sense. We want the node mutation to be distinct but not the entire entity: https://github.com/datacommonsorg/import/blob/e0a142b372fe62c478548d76407335f63eb044b4/pipeline/ingestion/src/main/java/org/datacommons/Entity.java#L112

Let me close this one and think about how best to model this.

PR #375, opened 2025-04-16
Refactor ingestion data classes.

keyurva:
* Splitting refactoring the ingestion pipeline into multiple PRs to keep it manageable.
* Keeping the current code in its own package until the refactoring is fully done.
* This PR focuses on refactoring the data classes into its own package.
* For entities, it creates nodes and edges at the outset instead of a wrapper entity.
* For observations, it captures dates and values into typed Java objects instead of strings.
* The refactored CacheReader is only responsible for parsing the cached rows and not deal with pcollections. This makes it possible to test it standalone without requiring a test pipeline.
* Adopts the new schema - object value hash and no id field on edges.

### DIFF
--- a/pipeline/ingestion/pom.xml
+++ b/pipeline/ingestion/pom.xml
@@ -37,6 +37,11 @@
       <version>1.7.32</version>
     </dependency>
     <dependency>
+        <groupId>com.google.guava</groupId>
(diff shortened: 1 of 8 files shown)


keyurva:
Closing this PR. Trying a few things locally first.

PR #376, opened 2025-04-18
Create graph ingestion pipeline based on mutation groups.

keyurva:


pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -0,0 +1,39 @@
...
+      this.spannerClient = spannerClient;
+    }
+
+    @ProcessElement
+    public void processElement(@Element String row, OutputReceiver<MutationGroup> c) {
+      if (CacheReader.isArcCacheRow(row)) {
+        NodesEdges nodesEdges = cacheReader.parseArcRow(row);
+        List<Mutation> mutations =
+            spannerClient.toGraphMutations(nodesEdges.getNodes(), nodesEdges.getEdges());
+        List<MutationGroup> mutationGroups = spannerClient.toGraphMutationGroups(mutations);
```
rohitkumarbhagat, line 34: - IIUC we are removing duplicates for mutation generated from a single cache row. Will there be many duplicate entries within same row? 
- Will it be more efficient if we can remove duplicates for all rows in a beam bundle? 
- Will it be simpler if this step outputs unique mutations and delegate mutation group creation to Spanner transform? Reason being, spanner transform logic to generate mutation groups takes multiple factor into consideration like  all mutations should be part of same transaction, group and sort across all mutations in a bundle etc. Hence, it might be more efficient in generating mutation groups. WDYT?

rohitkumarbhagat (approved):

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    List<MutationGroup> mutationGroups = new ArrayList<>();
+    for (List<Mutation> mutationList : mutationMap.values()) {
+      if (mutationList.size() == 1) {
+        mutationGroups.add(MutationGroup.create(mutationList.get(0)));
+      } else {
+        Mutation first = mutationList.get(0);
+        List<Mutation> rest = mutationList.subList(1, mutationList.size());
+        mutationGroups.add(MutationGroup.create(first, rest));
```
n-h-diaz, line 107: For my own understanding, could you explain why you're separating out the first mutation?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  public List<MutationGroup> toGraphMutationGroups(List<Mutation> mutations) {
+    Map<String, List<Mutation>> mutationMap = new HashMap<>();
+    Set<String> nodeIds = new HashSet<>();
+
+    for (Mutation mutation : mutations) {
+      String subjectId = mutation.asMap().get("subject_id").getString();
+
+      // Skip duplicate node mutations for the same subject_id
```
n-h-diaz, line 91: Will there be duplicate node mutations? It looks like the nodes are only being created for the EntityInfos within a cache row

n-h-diaz (approved):
Thanks for adding this!

Could you add a short readme on usage?

Also I was wondering how does the performance with MutationGroups compare with individual Mutations?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
keyurva: Thanks for the review!

Answers to your comments below:

1. For in-arcs, all arcs would point to the same subject so there could be fair amount of duplicate entries there.
2. I tried something similar with mutation groups - i.e. creating them in a bundle - but spanner was always using a global window which led to spanner initiating the writes only after all were transformed. Perhaps something we can look into later. Here's a canceled run that used this approach: http://shortn/_TdeoDy27rg
3. This is what the original pipeline did. However, this one had less lock contention and ran faster. My guess is that grouping all mutations with the same subject id even if only for a single cache row helped.

Overall, we should continue to experiment with the suggestions above and any others. Currently, this one seems to work well so we can get it in.

keyurva:
Thanks for the review!

> Could you add a short readme on usage?

Done.

> Also I was wondering how does the performance with MutationGroups compare with individual Mutations?

This one had less lock contention and ran faster than the other one.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
keyurva: It's because the MutationGroup class requires it to be created this way: https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/io/gcp/spanner/MutationGroup.html

It seems to use the first mutation as "primary" to determine partitions.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
keyurva: Yeah, there could be duplicates across different cache rows. We had tried distinct across the whole dataset earlier to address this but that led to writing to spanner being delayed until all distincts were identified.

n-h-diaz: Sorry I meant within the same row - as in, do you need to remove the duplicates here? For out-edges the nodes are created for the objects, and for in-edges the nodes are created for the subjects (so I'm not sure if there'd be any overlap)

keyurva: Oh ok. I was under the impression that for in-edges the nodes would be for the object. If there aren't going to be duplicates in the same row, I can remove this logic. LMK.

n-h-diaz: I don't think there should be duplicates, so we eventually can remove, but not super urgent

PR #379, opened 2025-04-30
Scale observation imports.

keyurva:
* Combined handling of graph and obs vs forking them earlier.
* Uses group-by for both.
* Experimented and landed on grouping keys that work well for the respective tables.
* Updated the default number of shards to a value that provides a good balance of batch size and distribution.

### DIFF
--- a/pipeline/ingestion/README.md
+++ b/pipeline/ingestion/README.md
@@ -32,7 +32,7 @@ mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.d
 
 ```shell
 mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.datacommons.ingestion.pipeline.IngestionPipeline \
--Dexec.args="--importGroupVersion=ipcc_2025_04_04_03_46_23 --skipProcessing=SKIP_GRAPH --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=20 --maxNumWorkers=20 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=n2-highmem-16"
+-Dexec.args="--importGroupVersion=ipcc_2025_04_04_03_46_23 --skipProcessing=SKIP_GRAPH --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=50 --maxNumWorkers=100 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
 ```
 
 ## org.datacommons.IngestionPipeline
--- a/pipeline/ingestion/README.md
+++ b/pipeline/ingestion/README.md
@@ -28,11 +28,11 @@ mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.d
 -Dexec.args="--importGroupVersion=ipcc_2025_04_04_03_46_23 --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=50 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
 ```
 
-### Import specific import group while skipping obs
+### Import specific import group while skipping graph
 
 ```shell
 mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.datacommons.ingestion.pipeline.IngestionPipeline \
--Dexec.args="--importGroupVersion=ipcc_2025_04_04_03_46_23 --skipProcessing=SKIP_GRAPH --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=50 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
+-Dexec.args="--importGroupVersion=ipcc_2025_04_04_03_46_23 --skipProcessing=SKIP_GRAPH --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=20 --maxNumWorkers=20 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=n2-highmem-16"
 ```
 
 ## org.datacommons.IngestionPipeline
--- a/pipeline/ingestion/README.md
+++ b/pipeline/ingestion/README.md
@@ -2,25 +2,37 @@
 
 ## org.datacommons.ingestion.pipeline.IngestionPipeline
 
-This module loads all tables (observations, nodes and edges) for all import groups into Spanner.
+This module loads all tables (observations, nodes and edges) for all or specified import groups into Spanner.
 The import group version to load from are fetched from the DC API's version endpoint.
 
-Example usage:
+Example usages:
+
+### Import all import groups
 
 ```shell
 mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.datacommons.ingestion.pipeline.IngestionPipeline \
--Dexec.args="--skipObservations=true --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=50 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
+-Dexec.args="--project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=50 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
 ```
 
-## org.datacommons.ingestion.pipeline.ImportGroupPipeline
-
-This module loads all tables (observations, nodes and edges) from a single import group version into Spanner.
-
-Example usage:
+### Import all import groups while skipping observations
 
 ```shell
-mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.datacommons.ingestion.pipeline.ImportGroupPipeline \
--Dexec.args="--importGroupVersion=auto1d_2025_03_26_02_16_23 --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=5 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
+mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.datacommons.ingestion.pipeline.IngestionPipeline \
+-Dexec.args="--skipProcessing=SKIP_OBS --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=50 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
+```
+
+### Import specific import group
+
+```shell
+mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.datacommons.ingestion.pipeline.IngestionPipeline \
+-Dexec.args="--importGroupVersion=ipcc_2025_04_04_03_46_23 --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=50 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
+```
+
+### Import specific import group while skipping obs
+
+```shell
+mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.datacommons.ingestion.pipeline.IngestionPipeline \
+-Dexec.args="--importGroupVersion=ipcc_2025_04_04_03_46_23 --skipProcessing=SKIP_GRAPH --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=50 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
 ```
 
 ## org.datacommons.IngestionPipeline
--- a/pipeline/ingestion/src/main/java/org/datacommons/ingestion/data/ProtoUtil.java
+++ b/pipeline/ingestion/src/main/java/org/datacommons/ingestion/data/ProtoUtil.java
@@ -1,0 +1,41 @@
+package org.datacommons.ingestion.data;
+
+import com.google.protobuf.Message;
+import com.google.protobuf.Parser;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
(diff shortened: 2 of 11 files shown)


pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -1,78 +1,190 @@
...
+    public void finishBundle() {
+      seenNodes.clear();
+      seenObs.clear();
+    }
+
+    @ProcessElement
+    public void processElement(
+        @Element KV<String, Iterable<Mutation>> kv, OutputReceiver<Mutation> out) {
+      for (var mutation : kv.getValue()) {
+        if (mutation.getTable().equals(spannerClient.getNodeTableName())) {
```
n-h-diaz, line 163: For nodes, would the iterable all be duplicates since the key is just subjectId? So do we need to go through all mutations or can we just output the first one?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -1,78 +1,190 @@
...
+          var subjectId = SpannerClient.getSubjectId(mutation);
+          if (seenNodes.contains(subjectId)) {
+            LOGGER.info("Duplicate node (extraction): {}", subjectId);
+            DUPLICATE_NODES_COUNTER.inc();
```
n-h-diaz, line 167: This counter seems like it's counting which nodes have multiple edges (so would be produced by multiple rows), which seems expected, so do we need to track this? Similarly, would it be helpful to count if there are duplicate edges (which would be unexpected)?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -1,78 +1,190 @@
...
+    public void processElement(@Element String row, OutputReceiver<KV<String, Mutation>> out) {
+      if (CacheReader.isArcCacheRow(row) && skipProcessing != SKIP_GRAPH) {
+        NodesEdges nodesEdges = cacheReader.parseArcRow(row);
+        var kvs = spannerClient.toGraphKVMutations(nodesEdges.getNodes(), nodesEdges.getEdges());
+        var filtered = spannerClient.filterGraphKVMutations(kvs, seenNodes);
+        filtered.forEach(out::output);
+
+        var dups = kvs.size() - filtered.size();
+        if (dups > 0) {
+          DUPLICATE_NODES_COUNTER.inc(dups);
```
n-h-diaz, line 116: Is this supposed to count duplicate nodes within the row or within the bundle? (It looks like it's doing both right now, but the former seems unexpected while the latter seems expected, so just want to check if we want to mix the two)

n-h-diaz (approved):
Thanks for the updates! It's exciting that all the data was loaded 🎉 😃 

This overall looks good - I had a few clarification questions, but there's been a lot of changes since I last saw this code, so let me know if I've misunderstood anything 

Also (can be follow up) it looks like there's some unused code (possibly experimental from before?) - should we clean this up? Or if we'd like to retain it, could you add some comments highlighting the difference between the different transform variations?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
keyurva: It's the latter. `seenNodes` is maintained per bundle.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
keyurva: Currently, yes because the group key is same as subject ID. But we could change the grouping key so keeping this logic as-is to future proof it. Yes, we'll be iterating over duplicates redundantly but the iterable is in memory so the performance impact should be negligible.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
keyurva: The dup nodes is mostly a measure of the denormalized nature of the BT cache and not something unexpected like dup obs.

Let me think about dup edges. We could count it but a `seenEdges` set could get quite large.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -1,156 +1,179 @@
...
+        } else if (mutation.getTable().equals(spannerClient.getObservationTableName())) {
+          var key = SpannerClient.getFullObservationKey(mutation);
+          if (seenObs.contains(key)) {
+            if (numObsDupsLogged < MAX_LOG_SAMPLES) {
+              LOGGER.info("Duplicate observation (extraction): {}", key);
+              numObsDupsLogged++;
+            }
+
+            DUPLICATE_OBS_COUNTER.inc();
+            continue;
```
keyurva, line 139: BTW, I was incorrectly returning from here which was incorrect since one duplicate does not mean all others in the iterable are. In practice, we've not encountered dup obs (except in auto1w) so the runs are still valid but I've fixed it in case we run into dups in the future.

keyurva (commented):
Thanks for the review!

I've cleaned up the code and added support for gzipped protos. Can you PTAL again?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
rohitkumarbhagat: Can you add this as a comment as why seenNodes/seenObs was not kept method local? Another option would be to enforce subject-id (or its prefix) to be part of the key using unit test.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -1,156 +1,179 @@
...
 import org.datacommons.ingestion.spanner.SpannerClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Transforms and DoFns for the ingestion pipeline. */
 public class Transforms {
+  private static final int MAX_LOG_SAMPLES = 1000;
+
+  static class CacheRowKVMutationsDoFn extends DoFn<String, KV<String, Mutation>> {
+    private static final Counter DUPLICATE_OBS_COUNTER =
```
rohitkumarbhagat, line 28: I guess Beam Counters are thread safe - as static counters would be shared by all instances within a vm.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -1,156 +1,179 @@
...
 import org.datacommons.ingestion.data.ImportGroupVersions;
 import org.datacommons.ingestion.data.NodesEdges;
-import org.datacommons.ingestion.data.Observation;
 import org.datacommons.ingestion.spanner.SpannerClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Transforms and DoFns for the ingestion pipeline. */
 public class Transforms {
+  private static final int MAX_LOG_SAMPLES = 1000;
```
rohitkumarbhagat, line 25: MAX_LOG_SAMPLES_PER_BUNDLE?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
```diff
@@ -147,85 +172,63 @@ public List<KV<String, Mutation>> filterGraphKVMutations(
...
+
+  private int hashShard(Object... values) {
+    return Math.abs(Objects.hash(values)) % numShards;
+  }
 
-    return subjectId + "-" + shard;
+  public static String getFullObservationKey(Mutation mutation) {
+    var parts =
+        new String[] {
+          getMutationValue(mutation, "variable_measured"),
```
rohitkumarbhagat, line 222: [optional] Can getMutationValue lookup be batched. Instead of creating mutation map for each lookup, it's more efficient to create mutation map once and lookup all fields.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
rohitkumarbhagat: Had encountered the same bug earlier with Graphs :)

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
```diff
@@ -147,85 +172,63 @@ public List<KV<String, Mutation>> filterGraphKVMutations(
...
 
     String objectId = getMutationValue(mutation, "object_id");
     String objectHash = getMutationValue(mutation, "object_hash");
-    int shard = Objects.hash(objectId, objectHash) % numEdgeShards;
+    int shard = Math.abs(Objects.hash(objectId, objectHash)) % numShards;
+
+    var key = Joiner.on("::").join(subjectId, shard);
+
+    if (numEdgeGroupByKeysLogged < MAX_LOG_SAMPLES) {
+      LOGGER.info("Edge GroupBy key: {}", key);
```
rohitkumarbhagat, line 196: If this is just for sampling, we can also enable sampling in dataflow --experiments=enable_data_sampling

rohitkumarbhagat (approved):

pipeline/ingestion/src/main/java/org/datacommons/ingestion/data/ProtoUtil.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+
+  /**
+   * Serializes the specified proto to be stored in spanner as gzipped bytes.
+   *
+   * <p>Note that we don't Base64 encode them like we do in BT since base64 encoding almost triples
+   * the size.
+   */
+  public static byte[] serializeSpannerProto(Message proto) {
```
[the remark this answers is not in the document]
keyurva: Good idea. Done.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/IngestionPipelineOptions.java
```diff
@@ -44,17 +44,17 @@ public interface IngestionPipelineOptions extends PipelineOptions {
 
   void setImportGroupVersion(String importGroupVersion);
 
-  @Description("The number of shards to generate for writing edge mutations.")
-  @Default.Integer(100)
-  int getNumEdgeShards();
+  @Description("The number of shards to generate for writing mutations.")
+  @Default.Integer(1000)
+  int getNumShards();
```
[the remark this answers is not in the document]
keyurva: It's applicable for obs as well. See `SpannerClient::getObservationKVKey`.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
keyurva: I like your `enable_data_sampling` idea so removed logging them here.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
keyurva: Yeah static counters are ok. Interestingly the counts are captured separately for each step: https://screenshot.googleplex.com/57ZpF7fevsmnF9y

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
keyurva: Done. Added comment at declaration.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
keyurva: I like that. Removed logging.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
keyurva: True. Instead of a batch call, added a method that takes a mutation map. The caller first calls `mutation.asMap()` once and then calls the overloaded method with this map.

pipeline/model/src/main/proto/storage.proto
```diff
@@ -0,0 +1,13 @@
...
+package org.datacommons.proto;
+
+// Includes protos that are used in spanner storage.
+// The protos are typically gzipped and stored in a bytes column.
+
+// Observations represent the observations (time series) stored in the observations column in the Observation table.
+message Observations {
+  // Map from date to value.
+  // Examples: "2024" -> "123", "2025-05" -> "-456.78"
+  map<string, string> values = 1;
```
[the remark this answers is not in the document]
keyurva: It does but we're compressing the proto here and storing them as bytes so it would not. Let's discuss in tomorrow's meeting if we want to support querying into this proto. The cost will be more storage and a slight performance hit when fetching the full time series. The benefit would be query flexibility.

keyurva (commented):
Thanks for the review!

pipeline/ingestion/README.md
```diff
@@ -2,25 +2,37 @@
...
+```
 
-Example usage:
+### Import specific import group while skipping obs
 
 ```shell
-mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.datacommons.ingestion.pipeline.ImportGroupPipeline \
--Dexec.args="--importGroupVersion=auto1d_2025_03_26_02_16_23 --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=5 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
+mvn -Pdataflow-runner compile exec:java -pl ingestion -am -Dexec.mainClass=org.datacommons.ingestion.pipeline.IngestionPipeline \
+-Dexec.args="--importGroupVersion=ipcc_2025_04_04_03_46_23 --skipProcessing=SKIP_GRAPH --project=datcom-store --gcpTempLocation=gs://keyurs-dataflow/temp --runner=DataflowRunner --region=us-central1  --numWorkers=50 --dataflowServiceOptions=enable_google_cloud_profiler --workerMachineType=e2-highmem-16"
```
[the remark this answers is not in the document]
keyurva: Good catch! Updated section header to say skip graph.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/IngestionPipelineOptions.java
```diff
@@ -44,17 +44,17 @@ public interface IngestionPipelineOptions extends PipelineOptions {
...
 
-  void setNumEdgeShards(int numEdgeShards);
+  void setNumShards(int numShards);
 
   @Description("If true, observations will not be ingested.")
```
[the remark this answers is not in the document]
keyurva: Thanks - done.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -102,25 +214,77 @@ public void processElement(@Element KV<String, Mutation> kv, OutputReceiver<Muta
     }
   }
 
+  static class ExtractMutationsDoFn extends DoFn<KV<String, Iterable<Mutation>>, Mutation> {
```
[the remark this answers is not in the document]
keyurva: Removed this and all other unused code.

PR #381, opened 2025-05-14
Fix differ options for file format

vish-cs:


n-h-diaz (approved):

pipeline/differ/src/main/java/org/datacommons/DifferOptions.java
vish-cs, line 1: I have updated the PR to make it boolean. Now we need to specify --useOptimizedGraphFormat (no true/false required) for tfrecord files. I have updated the README. The flag can be omitted for MCF files. Only caveat is   --useOptimizedGraphFormat=false still makes it tfrecord.

n-h-diaz: Thanks for the update!

PR #380, opened 2025-05-12
[Ingestion pipeline] Ingest observations without GroupByKey

rohitkumarbhagat:
Changes
- Use GBK only for graph mutations. Observations can be written directly.
- Limit default sharding to be 1. This helps in co-locating all rows for a subject-id in same bundle -> same spanner batch -> less splits per batch
- Filter duplicate edges before and after GBK
- Update spanner IO config for better throughput

PR #382, opened 2025-05-14
Address review comment from pr 380

rohitkumarbhagat:
- Address review comment from pr https://github.com/datacommonsorg/import/pull/380
- Added documentation for selecting working configuration.
- Updated de-duplication logic to key level (within values of a key) from bundle level

### DIFF
--- a/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
+++ b/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
@@ -26,49 +26,59 @@ public class Transforms {
         Metrics.counter(CacheRowKVMutationsDoFn.class, "dc_duplicate_obs_creation");
     private static final Counter DUPLICATE_NODES_COUNTER =
         Metrics.counter(CacheRowKVMutationsDoFn.class, "dc_duplicate_nodes_creation");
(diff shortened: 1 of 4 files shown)


pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -99,34 +99,22 @@ static class ExtractKVMutationsDoFn extends DoFn<KV<String, Iterable<Mutation>>,
...
-    public void finishBundle() {
-      seenNodes.clear();
-      seenObs.clear();
-      seenEdges.clear();
-    }
-
     @ProcessElement
     public void processElement(
         @Element KV<String, Iterable<Mutation>> kv, OutputReceiver<Mutation> out) {
+      // Storing caches at the bundle level can create long-lived objects, potentially leading to
```
keyurva, line 109: Does this approach purge duplicates in the same ballpark as the bundle approach or is the difference significant?

rohitkumarbhagat: It's roughly the same (+ 1.5B edges/nodes each). Theoretically, all edges/nodes for a subject key should be part of same value list, hence should be deduplicated.

keyurva (approved):
Thanks for the detailed worker configuration comments!

PR #383, opened 2025-05-15
Store observations as a proto column instead of compressed bytes.

keyurva:


keyurva:
Thanks for the review!

PR #384, opened 2025-05-19
[Ingestion pipeline] Add bounded LRU caches for bundle-level de-duplication

rohitkumarbhagat:
Introduces LRU-based caches (seenNodes, seenEdges, seenObs)
within `CacheRowKVMutationsDoFn` for de-duplicating graph
and observation mutations at the bundle level.

Limiting memory footprint with bounded caches. This mitigates
risks such as OutOfMemoryErrors and GC thrashing that can arise from large, long-lived objects within a bundle's lifecycle.

### DIFF
--- a/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
+++ b/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
@@ -5,7 +5,9 @@ import static org.datacommons.ingestion.pipeline.SkipProcessing.SKIP_OBS;
 
 import com.google.cloud.spanner.Mutation;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.io.TextIO;
@@ -17,9 +19,12 @@ import org.datacommons.ingestion.data.CacheReader;
 import org.datacommons.ingestion.data.ImportGroupVersions;
 import org.datacommons.ingestion.data.NodesEdges;
 import org.datacommons.ingestion.spanner.SpannerClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Transforms and DoFns for the ingestion pipeline. */
 public class Transforms {
+  private static final Logger LOGGER = LoggerFactory.getLogger(Transforms.class);
 
   static class CacheRowKVMutationsDoFn extends DoFn<String, KV<String, Mutation>> {
     private static final Counter DUPLICATE_OBS_COUNTER =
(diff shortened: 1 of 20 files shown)


pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -34,10 +39,15 @@ static class CacheRowKVMutationsDoFn extends DoFn<String, KV<String, Mutation>>
...
-    private final Set<String> seenNodes = new HashSet<>();
-    private final Set<String> seenEdges = new HashSet<>();
-    private final Set<String> seenObs = new HashSet<>();
+    // Using a bounded cache to prevent excessive memory consumption.
+    // As these are bundle-level caches, their size depends on the bundle's data.
+    // Large caches can block memory for the bundle's duration, risking OOM errors,
+    // GC thrashing, or worker failures.
+    // Note: Increasing capacity can enhance filtering performance, but might require more worker
+    // memory.
+    private final LRUCache<String, Boolean> seenNodes = new LRUCache<>(5_000_000);
```
keyurva, line 48: nit: Consider wrapping the cache into a `Set` since a set is more applicable for our use case. You can use [newSetFromMap](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collections.html#newSetFromMap(java.util.Map)) to do this.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
```diff
@@ -34,10 +39,15 @@ static class CacheRowKVMutationsDoFn extends DoFn<String, KV<String, Mutation>>
     private final SkipProcessing skipProcessing;
     private final TupleTag<KV<String, Mutation>> graphTag;
     private final TupleTag<KV<String, Mutation>> observationTag;
-    // TODO: Explore filtering duplicates across all bundles within a JVM.
-    private final Set<String> seenNodes = new HashSet<>();
-    private final Set<String> seenEdges = new HashSet<>();
-    private final Set<String> seenObs = new HashSet<>();
+    // Using a bounded cache to prevent excessive memory consumption.
```
keyurva, line 42: Were there any noticeable changes to the duplicate counters after this change?

keyurva (approved):

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
rohitkumarbhagat: - Nodes duplicate counter was nearly the same (around 1.5B).
- Edges duplicate counter reported decreased value for Biomedical (500K vs 1.5M). 
- No degradation was seen in pipeline response time. The reason could be duplicate Nodes count far outweigh duplicate edges. Moreover, duplicate edges were finally removed after GBK.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/Transforms.java
rohitkumarbhagat: Nice suggestion. We can take it up in future cleanup.

PR #385, opened 2025-05-19
Save geojson and kml fields as compressed string

rohitkumarbhagat:
Adds a new column `object_bytes` in Edges table.


pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/IngestionPipelineOptions.java
```diff
@@ -83,7 +84,7 @@ public interface IngestionPipelineOptions extends PipelineOptions {
   static class SkipPredicatePrefixesFactory implements DefaultValueFactory<List<String>> {
     @Override
     public List<String> create(PipelineOptions options) {
-      return List.of("geoJsonCoordinates", "kmlCoordinates");
+      return Collections.emptyList();
```
keyurva, line 87: Can we get rid of this option completely?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
```diff
@@ -30,6 +37,15 @@ public class SpannerClient implements Serializable {
   // Commit deadline for spanner writes. Use large value for bigger batches.
   private static final int SPANNER_COMMIT_DEADLINE_SECONDS = 120;
 
+  // Predicates for which the object value should be stored as bytes in Spanner.
+  private static final Set<String> STORE_VALUE_AS_BYTES_PREDICATES =
```
keyurva, line 41: Wondering if these should come from an external config. It's probably ok for now given that the list is small but I'll leave it up to you.

keyurva (approved):

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/IngestionPipelineOptions.java
rohitkumarbhagat: Done.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
rohitkumarbhagat: I would prefer keeping it in code for now. In case it grows, we can externalize it

PR #390, opened 2025-05-27
Add go_package to spanner storage proto

n-h-diaz:
This is so


keyurva:
> This is so the proto can be decoded in mixer when reading from spanner.
> 
> There were decoding errors when duplicating the proto in mixer - this was one option suggested by gemini, the other was to decode each field individually and decode the proto to bytes then unmarshal. Let me know if you have a preference or have another way to do this!

Is the plan to copy `storage.pb.go` from this project into mixer or something else?

I'm not a fan of the former. It makes the projects tightly coupled and not easily maintainable.

Some questions about the errors in mixer:

- What were the decoding errors like?
- Was it a like for like copy of the proto or was anything changed?

All in all, it'd be good if we can make the proto copy work instead of copying the generated go file.

PR #395, opened 2025-06-04
Create obs series nodes and edges.

keyurva:
* Creates 1 node and 4 edges for each time series.
* Using the type `StatVarObsSeries` for these nodes. We'll need this type to be defined in the schema.
* Currently creates 4 edges for each obs series - name, type, variable measured and observation about.
  + We can add any other edges we need in this PR or later.
*  Currently, the obs series DCID is used for nodes and edges but is not recorded in the Observation table.
   + We can add it later if there's a use case for it.
* This feature is behind a flag (`--writeObsGraph`, `false` by default) until we decide we want to do this or not for sure.

### DIFF
--- a/pipeline/ingestion/src/test/java/org/datacommons/ingestion/data/CacheReaderTest.java
+++ b/pipeline/ingestion/src/test/java/org/datacommons/ingestion/data/CacheReaderTest.java
@@ -126,9 +126,52 @@ public class CacheReaderTest {
                 .importName("NOAA_GFS_WeatherForecast")
                 .build());
 
+    NodesEdges expectedGraph = new NodesEdges()
+            .addNode(
+                    Node.builder()
+                            .subjectId("dc/os/Mean_PrecipitableWater_Atmosphere_geoId_sch2915390_870755143")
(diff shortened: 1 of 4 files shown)


pipeline/ingestion/src/test/java/org/datacommons/ingestion/data/CacheReaderTest.java
```diff
@@ -126,9 +126,52 @@ public void testParseTimeSeriesRow() {
                 .importName("NOAA_GFS_WeatherForecast")
                 .build());
 
+    NodesEdges expectedGraph = new NodesEdges()
+            .addNode(
+                    Node.builder()
+                            .subjectId("dc/os/Mean_PrecipitableWater_Atmosphere_geoId_sch2915390_870755143")
+                            .name("Mean Precipitable Water Atmosphere | geoId/sch2915390 | 870755143")
```
n-h-diaz, line 133: This is fine for now, but maybe we can think about how to make these more readable, especially the facet ids (or if we need the names at all)

n-h-diaz (approved):
Thanks for adding these! 

Since this is still experimental, what do you think of having these extra edges flag protected? I'm wondering how including these will affect load times/storage

PR #401, opened 2025-06-17
Generate optimized graph from import tool

vish-cs:


util/src/main/java/org/datacommons/util/GraphUtils.java
```diff
@@ -353,4 +355,40 @@ public static List<McfGraph> readMcfFile(String fileName) throws FileNotFoundExc
...
+   * StatVarObservations by their key.
+   *
+   * @param graph A list of McfGraph protos, typically representing StatVarObservations.
+   * @return An OptimizedMcfGraph proto containing the grouped StatVarObservations.
+   */
+  public static OptimizedMcfGraph buildOptimizedMcfGraph(List<McfGraph> graph) {
+    List<McfStatVarObsSeries> svo_list = new ArrayList<>();
+    for (McfGraph g : graph) {
+      for (PropertyValues pv : g.getNodesMap().values()) {
+        if (isObservation(pv)) {
```
n-h-diaz, line 370: Just wondering, is there a reason to add this check here and not in the parallel version?

n-h-diaz (approved):
Thanks for adding (and for adding both versions too)!

util/src/main/java/org/datacommons/util/GraphUtils.java
vish-cs: Good catch. That was a miss. I have added it now.

PR #408, opened 2025-07-09
Prevent generated stat var grouping ids from being greater than 255 characters to avoid database column length restrictions

dwnoble:
Changes:
- Updated stat var grouping generation to keep id lengths <= 255 characters to avoid a databse column length restriction
- Updated unit tests
- Ran linter which formatted a few unrelated files


chejennifer (approved):
nice

PR #411, opened 2025-07-22
[dc-import] Optimize memory usage in data processing pipeline through immediate observation extraction

rohitkumarbhagat:
Replaces full graph storage with pre-extracted observation series to reduce memory footprint during large data processing.

- Did some test on a csv file sized 900MiB
- Without mcf graph object list - ~ 3Gib
- With mcf graph object list > 32Gib (oom after 32Gib ). The progress was slow because of aggressive GC cycles, which were not able to free any ram.
- Made some improvement in this PR, memory consumption down to 16Gib. The task completed too in 7-8 mins. GC cycles do show cleanup.


vish-cs (commented):
Thanks for making the changes. Have you analyzed the memory footprint of the old/new implementation? Curious what the behavior is.

vish-cs (approved):

PR #414, opened 2025-07-23
Add a flag to generate optimized graph

vish-cs:


tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -304,30 +304,34 @@ private void processTable(File csvFile)
...
+        // Group observations incrementally by key to reduce memory usage
+        for (McfStatVarObsSeries obs : extractedObservations) {
+          McfStatVarObsSeries.Key key = obs.getKey();
+          groupedObservations
+              .computeIfAbsent(key, k -> McfStatVarObsSeries.newBuilder().setKey(k))
+              .addAllSvObsList(obs.getSvObsListList());
+        }
+        if (!logCtx.trackStatus(1, "rows processed")) {
+          throw new DCTooManyFailuresException("encountered too many failures");
+        }
```
rohitkumarbhagat, line 319: This log should be outside flag check.

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -304,30 +304,34 @@ private void processTable(File csvFile)
...
-    String filePath = Paths.get(args.outputDir.toString(), "optimized_graph.pb").toString();
-    logger.info("Writing optimized graph file to {}", filePath);
-    // Build and write optimized graphs directly from grouped observations
-    try (FileOutputStream output = new FileOutputStream(filePath)) {
-      for (Mcf.McfStatVarObsSeries.Builder builder : groupedObservations.values()) {
-        McfOptimizedGraph graph =
-            McfOptimizedGraph.newBuilder().setSvObsSeries(builder.build()).build();
-        graph.writeDelimitedTo(output);
+    if (args.generateOptimizedGraph) {
+      String filePath = Paths.get(args.outputDir.toString(), "optimized_graph.pb").toString();
```
rohitkumarbhagat, line 323: Each csv  file is processed in separate thread. All writing to same graph file might lead to race condition.

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -304,30 +304,34 @@ private void processTable(File csvFile)
...
+      logger.info("Writing optimized graph file to {}", filePath);
+      // Build and write optimized graphs directly from grouped observations
+      try (FileOutputStream output = new FileOutputStream(filePath)) {
+        for (Mcf.McfStatVarObsSeries.Builder builder : groupedObservations.values()) {
+          McfOptimizedGraph graph =
+              McfOptimizedGraph.newBuilder().setSvObsSeries(builder.build()).build();
+          graph.writeDelimitedTo(output);
```
rohitkumarbhagat, line 330: Optional: BufferedFileOutputStream might be optimal for writing large number of rows. Can we leave a todo in case planned in future?

rohitkumarbhagat (approved):

tool/src/main/java/org/datacommons/tool/Processor.java
vish-cs: Addressed in PR #415

tool/src/main/java/org/datacommons/tool/Processor.java
vish-cs: Addressed in PR #415

PR #417, opened 2025-07-30
Update pom.xml

n-h-diaz:
Updates pom.xml files to fix broken ingestion pipeline ([sample failed run](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2025-07-29_12_32_11-17071637354311084760;graphView=0?project=datcom-store&inv=1&invt=Ab4Gew&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22)))). It seemed like a beam error, so I updated the beam SDK to the latest (2.62.0 --> 2.66.0), but then had to update some other dependencies. 

This includes
* Using a BOM to manage beam dependencies
* Explicitly adding the google-cloud-spanner dependency
* Using dependencyManagement for protobuf dependencies in the parent datacommons-import so the same protobuf versions are used throughout, since the protos are shared across modules (and updates the protobuf version to 3.25.1) 
* Switches to use org.xolstice.maven.plugins for protobuf generation
* Adds a few other dependencies needed for the new protobuf version: gson, guava

[Succesful test run](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2025-07-29_19_25_29-8256223821596906265;logsSeverity=ERROR;graphView=0?project=datcom-store&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))&inv=1&invt=Ab4Ekw)

### DIFF
--- a/pipeline/pom.xml
+++ b/pipeline/pom.xml
@@ -10,6 +10,13 @@
     <url>https://datacommons.org</url>
     <packaging>pom</packaging>
 
+  <parent>
+      <groupId>org.datacommons</groupId>
+      <artifactId>datacommons-import</artifactId>
+      <version>0.1-alpha.1</version>
+      <relativePath>..</relativePath>
+  </parent>
+
   <modules>
     <module>differ</module>
     <module>ingestion</module>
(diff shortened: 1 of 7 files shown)


pipeline/pom.xml
```diff
@@ -10,6 +10,13 @@
     <url>https://datacommons.org</url>
     <packaging>pom</packaging>
 
+  <parent>
+      <groupId>org.datacommons</groupId>
+      <artifactId>datacommons-import</artifactId>
+      <version>0.1-alpha.1</version>
```
vish-cs, line 16: Can we define the version as a property so it's not hard coded here? This will make it easier to revise when we want to update the versions.

vish-cs (approved):

PR #420, opened 2025-08-04
Update differ to support schema diff

vish-cs:


pom.xml
```diff
@@ -110,14 +110,6 @@
             <configuration>
                 <protocArtifact>com.google.protobuf:protoc:${protoc.version}:exe:${os.detected.classifier}</protocArtifact>
                 <clearOutputDirectory>false</clearOutputDirectory>
-                <outputTargets>
```
n-h-diaz, line 113: Just curious why is this deleted?

pipeline/differ/src/main/java/org/datacommons/pipeline/differ/DifferUtils.java
```diff
@@ -20,28 +30,59 @@ enum Diff {
...
+                                .collect(Collectors.joining(","));
+                        String value = GraphUtils.getPropertyValue(pv, Property.value.name());
+                        c.output(KV.of(key, value));
+                      } else {
+                        List<String> keys = new ArrayList<>(pv.keySet());
+                        Collections.sort(keys);
+                        String value =
+                            keys.stream()
+                                .map(key -> GraphUtils.getPropertyValue(pv, key))
+                                .collect(Collectors.joining(","));
```
n-h-diaz, line 78: Since this will be written as a csv with a header with only 4 fields, it might be confusing if these also use commas to separate (and might be hard to parse the csv) - is there a different way to display the values? 

Also would it be helpful to include the properties too? (For example if the property names change but the values stay the same, or if a node adds/deletes pvs, so it's easy to match which values correspond)

pipeline/differ/src/main/java/org/datacommons/pipeline/differ/DifferUtils.java
```diff
@@ -20,28 +30,59 @@ enum Diff {
     UNMODIFIED;
   }
 
+  public static final TupleTag<KV<String, String>> OBSERVATION_NODES_TAG =
+      new TupleTag<KV<String, String>>() {};
+  public static final TupleTag<KV<String, String>> SCHEMA_NODES_TAG =
+      new TupleTag<KV<String, String>>() {};
+
+  public static final String[] GRUOPBY_PROPERTIES = {
```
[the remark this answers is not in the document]
vish-cs: fixed

pom.xml
vish-cs: I am getting build warning for this mentioned outputTargets is not supported for plugin protobuf-maven-plugin. I have reverted it. Can you please have a look?

pipeline/differ/src/main/java/org/datacommons/pipeline/differ/README.md
```diff
@@ -16,6 +16,6 @@ The pipeline can be run locally (DirectRunner) or in the cloud (DataflowRunner).
 
 
 ```
-mvn compile exec:java  -pl differ -am -Dexec.mainClass=org.datacommons.DifferPipeline -Dexec.args="--currentData=current*.gz --previousData=previous*.gz --useOptimizedGraphFormat --outputLocation=gs://bucket/differ/output"
+mvn compile exec:java  -pl differ -am -Dexec.mainClass=org.datacommons.pipeline.differ.DifferPipeline -Dexec.args="--currentData=current/*.mcf --previousData=previous/*.mcf --outputLocation=gs://bucket/differ/output"
```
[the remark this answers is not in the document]
vish-cs: Updated REAME with additionl info

pom.xml
n-h-diaz: Oh I see, I think it's okay to remove if it's causing build errors. Just curious where the protos will be stored?

n-h-diaz (approved):
Thanks!

pom.xml
vish-cs: They are generated in util/target/generated-sources/protobuf/java/org/datacommons/proto/ if nothing is specified.

vish-cs:
Thanks for the review!

PR #421, opened 2025-08-05
Fix parent versions in pom.xml

n-h-diaz:
Follow up to https://github.com/datacommonsorg/import/commit/4c2964f0c1961c084d8ec1ca0952ce08dd15cede 

It looks like parent versions need to be explicitly set to compile the ingestion module to run the ingestion pipeline. (Apparently revision is a special property that allows this, but custom properties do not) 

To keep the stable modules (0.1-alpha.1) separate from the development modules (0.1-SNAPSHOT, which is currently set for revision), I have to set the parent versions directly. 

Tested by running ingestion pipeline

### DIFF
--- a/pipeline/pom.xml
+++ b/pipeline/pom.xml
@@ -13,7 +13,7 @@
   <parent>
       <groupId>org.datacommons</groupId>
       <artifactId>datacommons-import</artifactId>
-      <version>${import.version}</version>
+      <version>0.1-alpha.1</version>
       <relativePath>..</relativePath>
   </parent>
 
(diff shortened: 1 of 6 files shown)


vish-cs (approved):
I also figured that yesterday. I was planning to make the same change. Wondering if we should just use revision to avoid too different version numbers?

n-h-diaz:
> I also figured that yesterday. I was planning to make the same change. Wondering if we should just use revision to avoid too different version numbers?

Makes sense -- updated to use revision everywhere (so now all modules have the same version). Thanks for the review!

PR #423, opened 2025-08-07
Fix logic to build optimized graph

vish-cs:
- Handle coder error
- Add unit test


pipeline/util/src/main/java/org/datacommons/pipeline/util/PipelineUtils.java
```diff
@@ -148,7 +151,8 @@ public McfGraph apply(String input) {
    * @param graph PCollection of McfGraph protos.()
    * @return PCollection of McfOptimizedGraph protos.
    */
-  public static PCollection<McfOptimizedGraph> buildOptimizedMcfGraph(PCollection<McfGraph> graph) {
+  public static PCollection<McfOptimizedGraph> buildOptimizedMcfGraph(
+      PCollection<McfGraph> graph, boolean sort) {
```
n-h-diaz, line 155: Is there ever a reason to not sort the obs or can it just be the default?

vish-cs: Makes sense. I was als wondering about it. I have changed it to be the default

PR #424, opened 2025-08-07
Update project version number

vish-cs:
Bring  all the module to a common version to avoid confusion. Also, we plan to release a new version of import tool on GitHub.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -192,6 +192,31 @@
                     <projectId>GCLOUD_CONFIG</projectId>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>flatten-maven-plugin</artifactId>
+                <version>1.6.0</version> 
+                <configuration>
+                    <updatePomFile>true</updatePomFile>
+                    <flattenMode>resolveCiFriendliesOnly</flattenMode>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>flatten</id>
+                        <phase>process-resources</phase>
+                        <goals>
+                            <goal>flatten</goal>
+                        </goals>
+                    </execution>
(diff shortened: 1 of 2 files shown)


n-h-diaz (commented):
Thanks!

Just for my own understanding, after this change what will the output versions be for all the modules? It the intention to have them all the same or will some use a snapshot version? (it looks like they're all getting set to revision, so just wanted to check) 

Also just to confirm, with these changes the modules all build? Both at the top level and for each submodule (e.g. the ingestion pipeline can still be run)

vish-cs:
> Thanks!
> 
> Just for my own understanding, after this change what will the output versions be for all the modules? It the intention to have them all the same or will some use a snapshot version? (it looks like they're all getting set to revision, so just wanted to check)
> 
> Also just to confirm, with these changes the modules all build? Both at the top level and for each submodule (e.g. the ingestion pipeline can still be run)

The idea is to bring them all to a common version to avoid confusion. Also, we want to release a new version of import tool on GitHub since the last release is from 2023.

Good catch. The repo builds but I see a version related error when trying to run the pipelines. I will hold this PR until further investigation.

pom.xml
```diff
@@ -13,7 +13,7 @@
         <maven.compiler.source>17</maven.compiler.source>
         <maven.compiler.target>17</maven.compiler.target>
         <!-- Dependency versions -->
-        <revision>0.1-alpha.1</revision>
+        <revision>0.1-SNAPSHOT</revision>
```
hqpho, line 16: Does the `SNAPSHOT` portion of the revision number get removed during the build process?

vish-cs: We override the revision property for a stable build before the release. The under development version in the repo remains SNAPSHOT.

hqpho (approved):

PR #426, opened 2025-08-15
Create spanner tables in ingestion pipeline

n-h-diaz:
The pipeline will create the database and tables if they don't already exist

So we don't have to manually create new tables and can automatically include any ddl changes

Tested with dc_graph_16 and dc_graph_17

### DIFF
--- a/pipeline/model/src/main/proto/storage.proto
+++ b/pipeline/model/src/main/proto/storage.proto
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package org.datacommons.proto;
+package org.datacommons;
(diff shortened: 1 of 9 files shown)


pipeline/model/src/main/proto/storage.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package org.datacommons.proto;
+package org.datacommons;
```
n-h-diaz, line 3: Removed "proto" since the spanner database admin was parsing this as a sql term

PR #432, opened 2025-09-10
Add test for nodes with type Thing

n-h-diaz:
Adds some tests to check that nodes don't explicitly have typeOf: Thing set.

This is to avoid import group merging issues when it's set as a placeholder type

### DIFF
--- a/pipeline/ingestion/src/main/java/org/datacommons/ingestion/data/CacheReader.java
+++ b/pipeline/ingestion/src/main/java/org/datacommons/ingestion/data/CacheReader.java
@@ -132,8 +132,11 @@ public class CacheReader implements Serializable {
               }
             }
 
-            // Add node.
-            if (!nodeId.isEmpty()) {
+            // Maybe add node.
+            // Triples with type Thing are from two different import groups.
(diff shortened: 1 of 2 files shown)


util/src/main/java/org/datacommons/util/McfChecker.java
```diff
@@ -438,6 +438,7 @@ private Set<String> checkCommon(String nodeId, Mcf.McfGraph.PropertyValues node)
       throws IOException, InterruptedException {
     Set<String> types =
         new HashSet<>(checkRequiredValueProp(nodeId, node, "Thing", Vocabulary.TYPE_OF));
+    checkTypes(nodeId, node, types);
```
vish-cs, line 441: So we simply log if if we find node with type Thing and then continue as before?

n-h-diaz: This is just for the lint tests, so I didn't to skip so we could continue validating the nodes and collecting all the syntax errors.

This will show up as an error in the report, so ideally then the user could fix the types and rerun the linter

vish-cs (approved):

PR #442, opened 2025-09-24
Add retry logic using Failsafe library for API connection failures

rohitkumarbhagat:
- Replace manual retry with Failsafe 3.3.2 library
- Implement exponential backoff (1s, 2s, 4s, 8s, 8s...)
- Configure retry only for network operations (IOException)
- Add proper logging with log4j
- Support enabling/disabling retries with ENABLE_RETRIES flag
- Fixes connection reset issues during long-running imports


ajaits (approved):
Thanks Rohit! This is very useful!

PR #488, opened 2026-03-13
Add constraint property graph mutation

vish-cs:
Add a new MCF transformation to add constraint properties in McfMutator where we already do quantity range and lat-long transformations.
Updated various test files required due to the added logic


n-h-diaz (approved):
Thank you!

PR #494, opened 2026-04-07
[DCP] Auto create tables when DB is empty

gmechali:
Modify the ingestion pipeline's logic for DB creation to be a bit more flexible. Before this change, it just checks whether the DB exists, and if not it creates it with the appropriate scheme and protodescriptor.

With this change, we retain the ability to create the DB and initialize it properly, but we also allow customers to provide an empty DB, and have it be initialized with the proper schema and proto descriptor.
This is important for DCP, where the DB should be created via Terraform at setup time in order to also connect the DCP service to the DB, however the schema is not necessary at that point, and so we can expect the ingestion pipeline to validate the schema and create it when necessary.

I've tested this on all the following scenarios:
- DB exists and is properly initialized
- DB exists but has no tables or proto descriptor
- DB exists but has no tables only
- DB does not exist.

### DIFF
--- a/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
+++ b/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
@@ -58,7 +58,7 @@ public class GraphIngestionPipeline {
             .build();
 
     LOGGER.info("Starting Spanner DDL creation...");
-    spannerClient.createDatabase();
+    spannerClient.validateOrInitializeDatabase();
     LOGGER.info("Spanner DDL creation complete.");
 
     Pipeline pipeline = Pipeline.create(options);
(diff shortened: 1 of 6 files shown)


vish-cs:
Thanks for making this change. This was one of the pending items since the previous logic was incompatible with the Terraform setup.
https://github.com/datacommonsorg/data/blob/master/import-automation/terraform/main.tf

 FYI, Terraform deployment script creates additional tables in the DB for import automation.
https://github.com/datacommonsorg/data/blob/master/import-automation/workflow/spanner_schema.sql

vish-cs (approved):

gmechali:
> Thanks for making this change. This was one of the pending items since the previous logic was incompatible with the Terraform setup. https://github.com/datacommonsorg/data/blob/master/import-automation/terraform/main.tf
> 
> FYI, Terraform deployment script creates additional tables in the DB for import automation. https://github.com/datacommonsorg/data/blob/master/import-automation/workflow/spanner_schema.sql

Got it - so the DCP Terraform script now can also just keep its current logic of creating the DataBase, but letting the ingestion script initialize the tables.

We also could have initialized the tables in the Terraform script, but I think that holds some pretty significant risk. If a user ends up making a change to the Table Schema and reapplies terraform, it does not modify the table, it recreates it, and therefore could drop all the data in the tables.
That was why I did not want the DCP TF to create the Node/Edge/Obs tables. Maybe for ImportStatus it's a bit less critical...but im not sure. 

I will put some thought into it in a few days when I get to the TF Import Status setup for DCP.

PR #495, opened 2026-04-07
[DCP] Adds JSON-LD Support to the ingestion pipeline

gmechali:
This PR adds JSON-LD support for the ingestion pipeline. The first stage of the pipeline is responsible for converting from MCF and TMCF to an MCFGraph object. we now also add the ability to convert from JSON-LD and CSV with a .jsonld template to the same MCFGraph object.

We also refactored the MCF parsing logic into a util/parser subdirectory.
Added some unit tests.
We verified the behavior by running the pipeline with following two [MCF](https://paste.googleplex.com/5353764669227008), [JSONLD](https://paste.googleplex.com/4812516044701696).

See the data in the tables under this [Spanner DB](https://pantheon.corp.google.com/spanner/instances/hpho-datacommons-2/databases/gabe-test-dcp-db/details/tables?e=13803378&mods=-monitoring_api_staging&project=datcom-website-dev)


pipeline/util/src/main/java/org/datacommons/ingestion/util/PipelineUtils.java
```diff
@@ -162,6 +172,69 @@ public McfGraph apply(String input) {
...
+        .apply("ReadJsonLdFiles-" + name, FileIO.readMatches())
+        .apply(
+            "ParseJsonLd-" + name,
+            ParDo.of(
+                new DoFn<FileIO.ReadableFile, McfGraph>() {
+                  @ProcessElement
+                  public void processElement(
+                      @Element FileIO.ReadableFile file, OutputReceiver<McfGraph> receiver) {
+                    try (InputStream is = Channels.newInputStream(file.open())) {
+                      McfGraph graph = JsonLdParser.parse(is);
```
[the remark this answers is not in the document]
gmechali: Added a TODO for this

util/src/main/java/org/datacommons/util/parser/jsonld/JsonLdTemplateParser.java
```diff
@@ -0,0 +1,182 @@
...
+          }
+
+          // Fallback to static value if not a ColumnSource
+          if (map.containsKey("@value")) {
+            addProperty(nodeBuilder, property, map.get("@value").toString(), Mcf.ValueType.TEXT);
+          } else if (map.containsKey("@id")) {
+            String idStr = map.get("@id").toString();
+            Mcf.ValueType type =
+                idStr.startsWith("l:") ? Mcf.ValueType.UNRESOLVED_REF : Mcf.ValueType.RESOLVED_REF;
+            addProperty(nodeBuilder, property, idStr, type);
```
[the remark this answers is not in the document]
gmechali: done.

util/src/main/java/org/datacommons/util/parser/jsonld/JsonLdParser.java
```diff
@@ -0,0 +1,146 @@
...
+    addProperty(nodeBuilder, "dcid", id, org.datacommons.proto.Mcf.ValueType.TEXT);
+
+    for (Map.Entry<String, Object> entry : nodeMap.entrySet()) {
+      String key = entry.getKey();
+      if (key.contains("/")) {
+        key = key.substring(key.lastIndexOf('/') + 1);
+      }
+      if (key.contains("#")) {
+        key = key.substring(key.lastIndexOf('#') + 1);
+      }
```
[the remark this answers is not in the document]
gmechali: Fixed.

dwnoble (changes requested):
Nit on the input JSON-LD files: all property names and non-literal values should either be namespaced or have the values explicitly defined in the `@context`

in the linked example, `name` , `value`, and `StatVarObservation` are missing `@context` definitions, and the "schema" and "dcid" namespaces should be defined in the context. I also think we should use the datacommons.org definitions for observationDate, observationAbout, variableMeasured, and value because there are slight differences between the schema.org and datacommons.org definitions 😅 

the way you define properties like `"variableMeasured": "http://schema.org/variableMeasured",` is totally fine (minus the nitpick about using datacommons.org instead of schema.org), but personally i like to just define the namespaces and use those everywhere in the `@graph` definition like this:

```
{
  "@context": {
    "schema": "https://schema.org/",
    "dcid": "https://datacommons.org/browser/"
  },
  "@graph": [
    {
      "@id": "dcid:TestNode",
      "@type": "schema:Thing",
      "schema:name": "Test Node Name"
    },
    {
      "@id": "dcid:TestObservation_2020",
      "@type": "dcid:StatVarObservation",
      "dcid:observationAbout": {
        "@id": "dcid:TestNode"
      },
      "dcid:variableMeasured": {
        "@id": "dcid:Count_Person"
      },
      "dcid:value": 1000,
      "dcid:observationDate": "2020"
    }
]
```

(you can see the differences between those properties by going to, for example, https://schema.org/observationAbout and https://datacommons.org/browser/observationAbout)

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -260,28 +289,32 @@ public Void call() throws Exception {
   private void processTable(File csvFile)
       throws IOException, DCTooManyFailuresException, InterruptedException {
     logger.info("Checking CSV " + csvFile.getPath());
-    TmcfCsvParser parser =
+    interface GraphSupplier {
+      Mcf.McfGraph get() throws java.io.IOException, InterruptedException;
+    }
+
+    GraphSupplier parser;
+    McfFileGroup mcfGroup = (McfFileGroup) args.fileGroup;
```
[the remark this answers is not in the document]
gmechali: We arent yet support .jsonld in the tool processor and unclear that we ever will.

dwnoble (approved):

tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -175,52 +179,73 @@ private Processor(Args args) {
...
+        files = ((JsonLdFileGroup) args.fileGroup).getJsonLds();
+      } else {
+        throw new IllegalArgumentException(
+            "Unsupported file group type for Instance MCF processing: "
+                + args.fileGroup.getClass().getName());
       }
     } else {
-      for (var f : args.fileGroup.getTmcfs()) {
-        processNodes(type, f);
+      if (!(args.fileGroup instanceof McfFileGroup)) {
```
vish-cs, line 195: Should this be a TODO: we want to eventually support templates with JSON LD?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
```diff
@@ -141,12 +145,22 @@ private static void processImport(
...
     }
 
-    // 2. Read and Split Graph:
-    // Read the graph data (TFRecord or MCF files) and split into schema and observation nodes.
-    PCollection<McfGraph> graph =
-        graphPath.contains("tfrecord")
-            ? PipelineUtils.readMcfGraph(importName, graphPath, pipeline)
-            : PipelineUtils.readMcfFiles(importName, graphPath, pipeline);
+    PCollection<McfGraph> graph;
+    switch (PipelineUtils.resolveFormat(graphPath)) {
```
vish-cs, line 149: Actually I was planning to make some changes to this logic. We have scenarios where we have to read both MCF/TMCF and TFRecord files as part of a single ingestion when the graph path is *.* There are imports like Place etc. which have both types of files. Wondering if we can incorporate that here as well keeping JSONLD support in mind.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
```diff
@@ -181,9 +195,11 @@ private static void processImport(
...
 
     // Write Edges (wait for Nodes write and Edges delete)
-    edgeMutations.apply(
-        "EdgesWaitOn-" + importName, Wait.on(List.of(writtenNodes.getOutput(), deleteEdgesWait)));
-    spannerClient.writeMutations(pipeline, "WriteEdgesToSpanner-" + importName, edgeMutations);
+    PCollection<Mutation> waitingEdges =
+        edgeMutations.apply(
+            "EdgesWaitOn-" + importName,
+            Wait.on(List.of(writtenNodes.getOutput(), deleteEdgesWait)));
+    spannerClient.writeMutations(pipeline, "WriteEdgesToSpanner-" + importName, waitingEdges);
```
vish-cs, line 202: Curious why this change..we should be writing edgeMutations instead of the wait operation?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
```diff
@@ -193,9 +209,9 @@ private static void processImport(
         GraphReader.graphToObservations(optimizedGraph, importName, spannerClient, obsCounter)
             .apply("ExtractObsMutations-" + importName, Values.create());
     // Write Observations (wait for Obs delete)
-    observationMutations.apply("ObsWaitOn-" + importName, Wait.on(deleteObsWait));
+    PCollection<Mutation> waitingObs =
+        observationMutations.apply("ObsWaitOn-" + importName, Wait.on(deleteObsWait));
 
-    spannerClient.writeMutations(
-        pipeline, "WriteObservationsToSpanner-" + importName, observationMutations);
+    spannerClient.writeMutations(pipeline, "WriteObservationsToSpanner-" + importName, waitingObs);
```
vish-cs, line 215: Ditto

pipeline/util/src/main/java/org/datacommons/ingestion/util/PipelineUtils.java
```diff
@@ -162,6 +166,55 @@ public McfGraph apply(String input) {
...
```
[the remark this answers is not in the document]
vish-cs: Thoughts on this? We are splitting each MCF file into MCF nodes before processing them. Should be do something similar here.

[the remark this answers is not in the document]
gmechali: The issue is MCF is Splittable but JSON-LD is not. The format becomes invalid if you try to split it up. 
Im not sure how the streaming parser can parse a partial JSON-LD file tbh. I'll file a bug to look into this more, and will address as a follow up.

Regardless, I know I also need to support the JSON-LD schema + CSV, which I believe is the most common upload pattern for customers. And like that we are unlikely to hit memory issue on the JSON-LD. So just the CSV needs to be streamed.

Expect more on this in the future!

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
gmechali: I dont want to add too much to this PR. Can you either submit your PR and I can rebase on it, or I submit this one and you rebase on it? 

I'm not fully aware of the use case you're describing.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
gmechali: I actually thought this was just a readability change at first, but I missed the wait.On change. However, Gemini is calling out that without this, you risk database integrity issues, because if we dont wait for the node to be written, your edge could fail. 

I can revert if you dont believe this is a real bug?

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
gmechali: (same response as a above)

tool/src/main/java/org/datacommons/tool/Processor.java
gmechali: Yes - I have a bug open to track it. Will come back to support the templates with JSON-LD.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
vish-cs: Sure..that makes sense. I can update the logic once your code is checked in.

pipeline/util/src/main/java/org/datacommons/ingestion/util/PipelineUtils.java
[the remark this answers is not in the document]
vish-cs: Sounds good.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
vish-cs: ok got it. We can keep this.

vish-cs (approved):

PR #496, opened 2026-04-09
[DCP] Add a flag to only do DB initialization

gmechali:
This will allow us to provide a clean command for DCP customers to run Terraform apply, then trigger the ingestion pipeline with --initializeDatabaseOnly=true to then be in a clean state, and ready to import data.

Note it is somewhat optional, since they dont need the tables created at that point, but it provides for a cleaner story to DCP clients that once you run this you're ready. They could immediately run it with their data as well, but if it's not ready to import, this provides a clean instruction to be done with the setup.

### DIFF
--- a/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
+++ b/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
@@ -61,6 +61,11 @@ public class GraphIngestionPipeline {
     spannerClient.validateOrInitializeDatabase();
     LOGGER.info("Spanner DDL creation complete.");
 
+    if (options.getInitializeDatabaseOnly()) {
+      LOGGER.info("Skipping Beam data ingestion. Database initialized successfully.");
+      return;
+    }
+
     Pipeline pipeline = Pipeline.create(options);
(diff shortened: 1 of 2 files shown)


vish-cs:
> This will allow us to provide a clean command for DCP customers to run Terraform apply, then trigger the ingestion pipeline with --initializeDatabaseOnly=true to then be in a clean state, and ready to import data.
> 
> Note it is somewhat optional, since they dont need the tables created at that point, but it provides for a cleaner story to DCP clients that once you run this you're ready. They could immediately run it with their data as well, but if it's not ready to import, this provides a clean instruction to be done with the setup.

If the goal is simply to initialize, would it not be better to provide a script to do that rather than running a dataflow pipeline?

vish-cs (commented):
If the goal is simply to initialize, would it not be better to provide a script to do that rather than running a dataflow pipeline?

gmechali:
> If the goal is simply to initialize, would it not be better to provide a script to do that rather than running a dataflow pipeline?

Yes, but we dont want to maintain duplicated code / a duplicate schema that can fall out of sync with what the ingestion pipeline needs. So we could factor out this into some script and have the ingestion pipeline call it, or we can just rely on the ingestion pipeline since it already does it. 
wdyt? Should we refactor now, or should this just be a future cleanup?

vish-cs (approved):

PR #503, opened 2026-04-30
[DCP Ingestion] Adds JSON-LD exporter from legacy CDC data job

gmechali:
Fully expands the CDC JSONLD for data isnerted into the SQLite db. 
This also connects the CDC data Job to the Cloud Workflow and triggers it automatically.


simple/stats/jsonld_exporter.py
```diff
@@ -0,0 +1,180 @@
...
+      return None
+    if item.startswith("http://") or item.startswith("https://"):
+      return URIRef(item)
+    if item.startswith("dcid:"):
+      return URIRef(f"{DCID_URL}{item[5:]}")
+    return URIRef(f"{DCID_URL}{item.lstrip('/')}")
+
+  shard_index = 0
+
+  # 1. Process Triples (Schema) in chunks
```
dwnoble: Nit: can you extract the triples/schema logic & observation logic blocks to functions?

simple/stats/jsonld_exporter.py
```diff
@@ -0,0 +1,180 @@
...
+  compacted_jsonld = jsonld.compact(expanded_jsonld, ns_map)
+
+  if "@graph" not in compacted_jsonld:
+    data_only = {k: v for k, v in compacted_jsonld.items() if k != "@context"}
+    compacted_jsonld = {
+        "@context": compacted_jsonld.get("@context"),
+        "@graph": [data_only]
+    }
+
+  shard_name = f"output-{index:05d}.jsonld"
```
dwnoble: can you have different filename prefixes for observations and nodes? I think that would make debugging the output easier

simple/stats/main.py
```diff
@@ -53,10 +56,12 @@
 
 def _run():
   initialize_logger()
+  logging.info("Starting Runner in mode: %s", FLAGS.mode)
```
dwnoble: what do you think about: `Starting stats data importer job in mode: %s` of `starting runner`?

simple/stats/trigger_ingestion_workflow.py
```diff
@@ -0,0 +1,98 @@
+# Copyright 2024 Google Inc.
```
dwnoble: Nit: 2026

simple/stats/jsonld_exporter.py
```diff
@@ -0,0 +1,180 @@
+# Copyright 2024 Google Inc.
```
dwnoble: Nit: 2026

simple/stats/runner.py
```diff
@@ -549,6 +575,48 @@ def _create_importer(self, input_file: File) -> Importer:
...
+    # Write import info
+    self.db.insert_import_info(status=ImportStatus.SUCCESS)
+
+    # Export to JSON-LD
+    jsonld_dir = self.output_dir.open_dir("jsonld")
+    export_to_jsonld(self.db, jsonld_dir)
+
+    # Auto-trigger workflow if output is on GCS
+    output_path = jsonld_dir.full_path()
+    if output_path.startswith("gs://"):
```
dwnoble: Instead of checking if the output path is `gs://` i think it would make more sense to check if the WORKFLOW_LOCATION is set. wdyt?

simple/stats/main.py
gmechali: Done.

simple/stats/runner.py
gmechali: I think you're right to suggest checking the workflow, but I do think we also need to ensure it's in GCS. I just added a check before that INGESTION_WORKFLOW_NAME is passed in.

simple/stats/trigger_ingestion_workflow.py
gmechali: done.

simple/stats/jsonld_exporter.py
gmechali: done.

simple/stats/jsonld_exporter.py
gmechali: Done

simple/stats/jsonld_exporter.py
gmechali: Fair point, Done.

dwnoble (approved):

simple/stats/runner.py
dwnoble: That sounds right-thanks!

PR #513, opened 2026-05-26
Add the upsert logic for updating the recent update timestamp for spanner Node table

shixiao-coder:
This PR added the mutation logic to update the timestamp based on the insert/update logic from spanner table instruction https://docs.cloud.google.com/spanner/docs/commit-timestamp-postgresql#update-row

### DIFF
--- a/pipeline/spanner/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
+++ b/pipeline/spanner/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
@@ -333,6 +333,8 @@ public class SpannerClient implements Serializable {
       return Mutation.newInsertOrUpdateBuilder(nodeTableName)
           .set("subject_id")
           .to(node.getSubjectId())
+          .set("last_update_timestamp")
+          .to(Value.COMMIT_TIMESTAMP)
           .build();
     }
     return Mutation.newInsertOrUpdateBuilder(nodeTableName)
@@ -346,6 +348,8 @@ public class SpannerClient implements Serializable {
         .to(node.getName())
         .set("types")
         .toStringArray(node.getTypes())
+        .set("last_update_timestamp")
+        .to(Value.COMMIT_TIMESTAMP)
         .build();
   }
 
--- a/pipeline/spanner/src/main/resources/spanner_schema.sql
+++ b/pipeline/spanner/src/main/resources/spanner_schema.sql
@@ -9,6 +9,7 @@ CREATE TABLE Node (
   name STRING(MAX),
   types ARRAY<STRING(1024)>,
   name_tokenlist TOKENLIST AS (TOKENIZE_FULLTEXT(name)) HIDDEN,
+  last_update_timestamp TIMESTAMP OPTIONS (allow_commit_timestamp=true),
 ) PRIMARY KEY(subject_id)
 
 CREATE TABLE Edge (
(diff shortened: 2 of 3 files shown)


pipeline/spanner/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
```diff
@@ -333,6 +333,8 @@ public Mutation toNodeMutation(Node node) {
       return Mutation.newInsertOrUpdateBuilder(nodeTableName)
           .set("subject_id")
           .to(node.getSubjectId())
+          .set("last_update_timestamp")
+          .to(Value.COMMIT_TIMESTAMP)
```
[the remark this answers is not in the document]
shixiao-coder: added

pipeline/spanner/src/main/resources/spanner_schema.sql
```diff
@@ -9,6 +9,7 @@ CREATE TABLE Node (
   name STRING(MAX),
   types ARRAY<STRING(1024)>,
   name_tokenlist TOKENLIST AS (TOKENIZE_FULLTEXT(name)) HIDDEN,
+  last_update_timestamp TIMESTAMP OPTIONS (allow_commit_timestamp=true),
```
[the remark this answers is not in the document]
shixiao-coder: we would choose to, when initializing the DB, it will have the new field

pipeline/spanner/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
gmechali, line 337: as discussed in person, I think it might be nice to propagate an update timestamp through so that all ndoes written to from the same ingestion dataflow share a last_updated_timestamp.

However, it might be hijacking the meaning of last_updated_timestamp in some ways so if others disagree, happy to proceed like this

pipeline/spanner/src/test/java/org/datacommons/ingestion/spanner/SpannerClientTest.java
```diff
@@ -64,4 +66,40 @@ public void testCheckTableExists_FalsePositive() {
...
+            .name("Node Name")
+            .types(List.of("Type1", "Type2"))
+            .build();
+
+    Mutation mutation = spannerClient.toNodeMutation(node);
+    assertNotNull(mutation);
+    assertEquals("Node", mutation.getTable());
+
+    var mutationMap = mutation.asMap();
+    assertEquals("dcid:123", mutationMap.get("subject_id").getString());
```
gmechali: nit: I think it's a more common pattern to have the actual value first, then the expected value

Same result obviously but I think it's a bit more legible
assertEquals(mutationMap.get("subject_id").getString(), "dcid:123");

gmechali (approved):
Couple nits but LGTM otherwise. 

I just want to be very careful with the rollout of this - if any instance trying to run this doesnt have the last_update_timestamp, I think it would fail.

So the question is.... is there any instance that doesnt have that column? 
For DCP, we're fine to proceed. Any base DC spanner DBs without this column? @vish-cs @keyurva

pipeline/spanner/src/test/java/org/datacommons/ingestion/spanner/SpannerClientTest.java
shixiao-coder: updated

vish-cs (approved):

pipeline/spanner/src/main/java/org/datacommons/ingestion/spanner/SpannerClient.java
shixiao-coder: the current method is using the [TrueTime](https://docs.cloud.google.com/spanner/docs/true-time-external-consistency) that force incremental time and external consistency for actual commit/update in DB.

I think both way are fine since we are doing stale read anyway. Personally I'd be leaning toward the true commit time but fine to switch if need

PR #516, opened 2026-06-01
feat: support conditional 'dc/base/' prefix during ingestion

SandeepTuniki:
This PR introduces the `IS_BASE_DC` env to conditionally control `dc/base/` prefixing in the ingestion pipelines. When set to false (applicable for DCP), it dynamically strips this prefix during import.

### DIFF
--- a/pipeline/data/src/main/java/org/datacommons/ingestion/data/Observation.java
+++ b/pipeline/data/src/main/java/org/datacommons/ingestion/data/Observation.java
@@ -182,7 +182,8 @@ public class Observation {
         && Objects.equals(importName, that.importName)
         && Objects.equals(provenanceUrl, that.provenanceUrl)
         && Objects.equals(facetId, that.facetId)
-        && Objects.equals(isDcAggregate, that.isDcAggregate);
+        && Objects.equals(isDcAggregate, that.isDcAggregate)
+        && Objects.equals(isBaseDc, that.isBaseDc);
   }
 
   @Override
@@ -198,7 +199,8 @@ public class Observation {
         importName,
         provenanceUrl,
         facetId,
-        isDcAggregate);
+        isDcAggregate,
+        isBaseDc);
   }
 
   // Builder for Observation
--- a/pipeline/data/src/main/java/org/datacommons/ingestion/data/Observation.java
+++ b/pipeline/data/src/main/java/org/datacommons/ingestion/data/Observation.java
@@ -18,7 +18,6 @@ public class Observation {
   private static final String VARIABLE_MEASURED_PREDICATE = "variableMeasured";
   private static final String NAME_PREDICATE = "name";
   private static final String TYPE_OF_PREDICATE = "typeOf";
-  private static final String PROVENANCE_DCID_PREFIX = "dc/base/";
 
   private String observationAbout;
   private String variableMeasured;
@@ -31,6 +30,7 @@ public class Observation {
   private String provenanceUrl;
   private String facetId;
   private boolean isDcAggregate;
+  private boolean isBaseDc;
   private NodesEdges obsGraph;
 
   private Observation(Builder builder) {
@@ -45,6 +45,7 @@ public class Observation {
     this.provenanceUrl = builder.provenanceUrl;
     this.facetId = builder.facetId;
     this.isDcAggregate = builder.isDcAggregate;
+    this.isBaseDc = builder.isBaseDc;
     this.obsGraph = toObsGraph();
   }
 
@@ -110,7 +111,7 @@ public class Observation {
                     replaceSlashesWithUnderscores(observationAbout),
                     facetId);
     var seriesName = Joiner.on(" | ").join(variableMeasured, observationAbout, facetId);
-    var provenanceDcid = PROVENANCE_DCID_PREFIX + importName;
+    var provenanceDcid = ProvenanceUtils.getProvenanceDcid(importName, this.isBaseDc);
 
     // Add series node
     graph.addNode(
@@ -213,6 +214,7 @@ public class Observation {
     private String provenanceUrl = "";
     private String facetId = "";
     private boolean isDcAggregate = false;
+    private boolean isBaseDc = true;
 
     public Builder observationAbout(String observationAbout) {
       this.observationAbout = observationAbout;
@@ -251,6 +253,11 @@ public class Observation {
 
     public Builder isDcAggregate(boolean isDcAggregate) {
       this.isDcAggregate = isDcAggregate;
+      return this;
+    }
+
+    public Builder isBaseDc(boolean isBaseDc) {
+      this.isBaseDc = isBaseDc;
       return this;
     }
 
--- a/pipeline/data/src/main/java/org/datacommons/ingestion/data/ProvenanceUtils.java
+++ b/pipeline/data/src/main/java/org/datacommons/ingestion/data/ProvenanceUtils.java
@@ -18,17 +18,4 @@ public class ProvenanceUtils implements Serializable {
     return prefix + importName;
   }
 
-  /**
-   * Strips the 'dc/base/' prefix from the given ID if the run is not base DC.
-   *
-   * @param value The value to strip.
-   * @param isBaseDc Whether the run is base DC.
-   * @return The stripped value if not base DC and starts with prefix; otherwise the original value.
-   */
-  public static String stripPrefix(String value, boolean isBaseDc) {
-    if (!isBaseDc && value != null && value.startsWith(BASE_DC_PREFIX)) {
-      return value.substring(BASE_DC_PREFIX.length());
-    }
-    return value;
-  }
 }
--- a/pipeline/data/src/main/java/org/datacommons/ingestion/data/ProvenanceUtils.java
+++ b/pipeline/data/src/main/java/org/datacommons/ingestion/data/ProvenanceUtils.java
@@ -17,5 +17,4 @@ public class ProvenanceUtils implements Serializable {
     String prefix = isBaseDc ? BASE_DC_PREFIX : "";
     return prefix + importName;
   }
-
 }
--- a/pipeline/data/src/main/java/org/datacommons/ingestion/data/ProvenanceUtils.java
+++ b/pipeline/data/src/main/java/org/datacommons/ingestion/data/ProvenanceUtils.java
@@ -1,0 +1,34 @@
+package org.datacommons.ingestion.data;
+
+import java.io.Serializable;
+
+/** Util functions for managing provenance prefixes. */
+public class ProvenanceUtils implements Serializable {
+  public static final String BASE_DC_PREFIX = "dc/base/";
+
+  /**
+   * Generates the provenance DCID based on import name and whether it is base DC.
+   *
+   * @param importName The name of the import.
+   * @param isBaseDc Whether the run is base DC.
+   * @return The constructed provenance DCID.
+   */
+  public static String getProvenanceDcid(String importName, boolean isBaseDc) {
+    String prefix = isBaseDc ? BASE_DC_PREFIX : "";
+    return prefix + importName;
+  }
+
+  /**
+   * Strips the 'dc/base/' prefix from the given ID if the run is not base DC.
+   *
+   * @param value The value to strip.
+   * @param isBaseDc Whether the run is base DC.
+   * @return The stripped value if not base DC and starts with prefix; otherwise the original value.
+   */
+  public static String stripPrefix(String value, boolean isBaseDc) {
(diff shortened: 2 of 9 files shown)


pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
```diff
@@ -46,6 +47,9 @@ public static void main(String[] args) {
     IngestionPipelineOptions options =
         PipelineOptionsFactory.fromArgs(args).withValidation().as(IngestionPipelineOptions.class);
 
+    boolean isBaseDc = System.getenv("IS_BASE_DC") == null || Boolean.parseBoolean(System.getenv("IS_BASE_DC"));
+    options.setIsBaseDc(isBaseDc);
```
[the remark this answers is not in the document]
SandeepTuniki: Done.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/IngestionPipeline.java
```diff
@@ -19,6 +19,9 @@ public static void main(String[] args) {
     IngestionPipelineOptions options =
         PipelineOptionsFactory.fromArgs(args).withValidation().as(IngestionPipelineOptions.class);
 
+    boolean isBaseDc = System.getenv("IS_BASE_DC") == null || Boolean.parseBoolean(System.getenv("IS_BASE_DC"));
+    options.setIsBaseDc(isBaseDc);
```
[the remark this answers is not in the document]
SandeepTuniki: Done.

pipeline/data/src/main/java/org/datacommons/ingestion/data/Observation.java
```diff
@@ -31,6 +30,7 @@ public class Observation {
   private String provenanceUrl;
   private String facetId;
   private boolean isDcAggregate;
+  private boolean isBaseDc;
```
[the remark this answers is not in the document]
SandeepTuniki: Done.

SandeepTuniki:
The failing codacy check is about a function being long (>50 lines). I'm leaving it as-is for now, because the check is failing on a test file, and that test function's length is mostly due to constructing test fixture. I'm not sure if this specific codacy check is appropriate for a test function.

pipeline/data/src/main/java/org/datacommons/ingestion/data/ProvenanceUtils.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  /**
+   * Strips the 'dc/base/' prefix from the given ID if the run is not base DC.
+   *
+   * @param value The value to strip.
+   * @param isBaseDc Whether the run is base DC.
+   * @return The stripped value if not base DC and starts with prefix; otherwise the original value.
+   */
+  public static String stripPrefix(String value, boolean isBaseDc) {
```
gmechali: Why would we ever need to strip the prefix? If we just dont add it, then there should be no need to strip it. 

DCP users won't add dc/base/... to their imports!

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
```diff
@@ -46,6 +47,10 @@ public static void main(String[] args) {
     IngestionPipelineOptions options =
         PipelineOptionsFactory.fromArgs(args).withValidation().as(IngestionPipelineOptions.class);
 
+    if (System.getenv("IS_BASE_DC") != null) {
```
gmechali: can you just run System.getenv("IS_BASE_DC") once and store it in a variable,, instead of running it twice.

pipeline/util/src/main/java/org/datacommons/ingestion/util/CacheReader.java
```diff
@@ -139,25 +146,32 @@ public NodesEdges parseArcRow(String row, Counter mcfNodesWithoutTypeCounter) {
             // Triples with type Thing are from two different import groups.
             // The corresponding Node is a placeholder and shouldn't be added.
             // Instead it will be added as part of the import group where its defined.
-            if (!nodeId.isEmpty() && !typeOf.equals(PipelineUtils.TYPE_THING)) {
+            boolean isReferenceNode = isOutArcCacheRow(row) ? !entity.getDcid().isEmpty() : true;
+            String finalNodeId = stripPrefix(nodeId);
```
gmechali: I think same as above, I dont think any of these are necessary since we should modify the importName.

pipeline/util/src/main/java/org/datacommons/ingestion/util/CacheReader.java
```diff
@@ -226,4 +241,8 @@ public static final boolean isArcCacheRow(String row) {
   public static final boolean isObsTimeSeriesCacheRow(String row) {
     return row.startsWith(OBS_TIME_SERIES_CACHE_PREFIX);
   }
+
+  private String stripPrefix(String s) {
```
gmechali: Could we just import stripPrefix from ProvenanceUtils in this file instead of duplicating this?

pipeline/util/src/main/java/org/datacommons/ingestion/util/CacheReader.java
```diff
@@ -33,9 +34,15 @@ public class CacheReader implements Serializable {
   private static final String CACHE_KEY_SEPARATOR_REGEX = "\\^";
 
   private final String gcsBucketId;
+  private final boolean isBaseDc;
 
   public CacheReader(String gcsBucketId) {
+    this(gcsBucketId, true);
+  }
+
+  public CacheReader(String gcsBucketId, boolean isBaseDc) {
```
gmechali, line 38: Why do we need another constructor? Could we just use one single constructor?

pipeline/util/src/main/java/org/datacommons/ingestion/util/GraphReader.java
```diff
@@ -128,7 +128,13 @@ public static List<Edge> graphToEdges(McfGraph graph, String provenance) {
   }
 
   public static Observation graphToObservations(McfOptimizedGraph graph, String importName) {
```
gmechali: Can we just migrate usages of this function to graphToObservations? 

Is this really needed at all still? Where would we call this function over the one on l134

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
SandeepTuniki: Thanks for the pointer. I refactored this to a single ENV access.

pipeline/util/src/main/java/org/datacommons/ingestion/util/CacheReader.java
SandeepTuniki: I was going to keep the previous one for backwards compatibility purposes, but scanning through the codebase, there's only like ~2 usages. So I refactored it to have a single constructor now.

pipeline/util/src/main/java/org/datacommons/ingestion/util/GraphReader.java
SandeepTuniki: Similar to previous comment, I was going to keep the previous one for backwards compatibility purposes, but scanning through the codebase, there's not many usages. So I refactored it to have a single method now.

pipeline/data/src/main/java/org/datacommons/ingestion/data/ProvenanceUtils.java
SandeepTuniki: I wasn't sure if the DCP users ever need to load custom data into their instances, and I noticed that CacheReader reads from GCS (which I assume is part of base DC's ingestion process), so I wanted to make sure we strip away the prefix if DCP users end up reading from the GCS buckets.

@vish-cs Does this case ever arise? Based on that, I can either keep or remove this code.

SandeepTuniki: Per offline feedback from Vishal, I removed all the stripPrefix logic now.

pipeline/util/src/main/java/org/datacommons/ingestion/util/CacheReader.java
SandeepTuniki: Removed this code as the stripPrefix logic is no longer in the code.

pipeline/util/src/main/java/org/datacommons/ingestion/util/CacheReader.java
SandeepTuniki: Reverted this change.

pipeline/data/src/main/java/org/datacommons/ingestion/data/Observation.java
vish-cs, line 33: Why do we want to include isBaseDc as a field within an observation? This is a system/env. property so we don't need to replicate in each single observation. We can simply initialize the env. variable in ProvenanceUtils. This will also minimize the changes required.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/IngestionPipelineOptions.java
```diff
@@ -100,4 +100,10 @@ public interface IngestionPipelineOptions extends PipelineOptions {
   boolean getForceCombineNodes();
 
   void setForceCombineNodes(boolean forceCombineNodes);
+
+  @Description("Whether this is a base Data Commons ingestion run")
```
vish-cs, line 104: If we are relying on the env. variable, we should not need to add it as a pipeline option. This is for runtime properties.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
```diff
@@ -46,6 +47,11 @@ public static void main(String[] args) {
     IngestionPipelineOptions options =
         PipelineOptionsFactory.fromArgs(args).withValidation().as(IngestionPipelineOptions.class);
 
+    String isBaseDcEnv = System.getenv("IS_BASE_DC");
```
vish-cs, line 50: One caveat here is that any string (even FALSE) would cause it to be true. I am wondering if we can rename the env. var to something like BASE_DC_ENV to help address it.

pipeline/data/src/main/java/org/datacommons/ingestion/data/Observation.java
SandeepTuniki: When I used Jetski to understand the code flow, it noted that the beam pipeline sends the code to many different workers, and that setting the env in the launcher won't immediately make them accessible in the workers. So I put it in PipelineOptions and passed it to Observation objects.

And from what I can tell, we need it in Observation as the ingestion also creates edges, which contains the provenance values. So we need the isBaseDc here.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/IngestionPipelineOptions.java
SandeepTuniki: As mentioned in other comment, I understand the flow like this:

GKE Pod (with Env Var) -> main() -> PipelineOptions -> Dataflow Workers

So to make the ENV accessible in the workers, we need to have the launcher set the value in Pipeline options.

pipeline/data/src/main/java/org/datacommons/ingestion/data/Observation.java
vish-cs: oh..that makes sense. Thanks for explaining.

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
vish-cs: I am fine to also leave it as is.

vish-cs (approved):

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
SandeepTuniki: Since the data repo uses `IS_BASE_DC` env, let's use the same name to keep it consistent.

PR #517, opened 2026-06-01
Update the legacy data job for processing the --importName

gmechali:
Updates the legacy data job to support the --imports flag.
It has backwards compatibility, if --imports is not provided, we will continue looking for the config.json in the input/dir.
Otherwise, we will be looking in the subfolders provided.

CLI is updated in https://github.com/datacommonsorg/datacommons/pull/99

Note that this supports following thepath in GCS to make the provenance name, but it does not yet support per-import feature yet. We currently hack te provenance name of multiple different imports. The work needed to do this blew up the PR, so I am now splitting it up. This is JUST the import job changes for support a --imports flag and fetching the configs.

The follow up PR will add full support for multiple individual imports.

### DIFF
--- a/simple/stats/main.py
+++ b/simple/stats/main.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+import os
 
 from absl import app
 from absl import flags
@@ -29,8 +30,8 @@ flags.DEFINE_string("input_dir", constants.DEFAULT_INPUT_DIR,
                     "The input directory.")
 flags.DEFINE_string("output_dir", constants.DEFAULT_OUTPUT_DIR,
                     "The output directory.")
-flags.DEFINE_string("import_name", None,
-                    "The name of the import (subdirectory under input_dir).")
+flags.DEFINE_list("imports", [],
+                   "The names of the imports (subdirectories under input_dir).")
 flags.DEFINE_enum(
     "mode",
     RunMode.CUSTOM_DC,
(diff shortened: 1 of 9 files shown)


simple/stats/runner.py
```diff
@@ -212,6 +231,119 @@ def _read_config_from_file(self,
...
+        new_key = fspath.join(rel_dir, k)
+        merged_data["inputFiles"][new_key] = v
+
+      # Merge variables
+      variables = config_data.get("variables", {})
+      for k, v in variables.items():
+        merged_data["variables"][k] = v
+
+      # Merge sources
+      sources = config_data.get("sources", {})
```
dwnoble, line 285: If two `config.json`s have different provenances, this looks like it may have a conflict and overwrite one of the source's provenances.

simple/stats/runner.py
```diff
@@ -212,6 +231,119 @@ def _read_config_from_file(self,
...
+    Args:
+      configs: A list of File objects representing the config.json files to merge.
+      base_dir: The base directory used to calculate relative paths for input files.
+    """
+    import json
+
+    import fs.path as fspath
+
+    merged_data = {
+        "importName": constants.ALL_IMPORTS,
```
dwnoble, line 253: Would it be better to set it to a combined name (e.g., ["import1", "import2"]) for better tracking?

simple/stats/runner.py
```diff
@@ -46,7 +46,10 @@
 from stats.entities_importer import EntitiesImporter
 from stats.events_importer import EventsImporter
 from stats.importer import Importer
+from stats.jsonld_exporter import DCID_URL
```
dwnoble: these 3 imports look like they're unused

simple/stats/main.py
```diff
@@ -54,10 +57,22 @@
 def _run():
   initialize_logger()
   logging.info("Starting stats data importer job in mode: %s", FLAGS.mode)
+
+  input_dir = FLAGS.input_dir
+  if FLAGS.imports == [constants.ALL_IMPORTS]:
+    logging.info("Running bulk load for all imports under: %s", input_dir)
```
dwnoble: This if / else logic looks like it's duplicated below- can you consolidate it and add the logging below as well?

simple/stats/runner.py
gmechali: Correct. This is what I tried resolving in this PR, but then it blew up, so im leaving it for a child PR. 

Basically this (and https://github.com/datacommonsorg/datacommons/pull/99/changes) is just allowing the wiring for supporting per-import ingestion. Next piece of work will actually allow per import on a data level, which means cleaning up the provenances. And this needs a lot more conversation before 😨

simple/stats/runner.py
gmechali: Good question. I think this piece will need to change anyways with hte follow up. Because we do need to preserver the import for each one. 
This going to be irrelevant shortly.

simple/stats/runner.py
gmechali: Done thanks

simple/stats/main.py
gmechali: Fixed.

simple/stats/runner.py
dwnoble: gotcha-that sounds good to leave it to a follow up

simple/stats/runner.py
dwnoble: sounds good!

dwnoble (approved):

PR #519, opened 2026-06-04
[DCP - Ingestion] DCP Bridge Mode improvements

gmechali:
# Refactor Simple Importer Parallel Ingestion, GCS Exporter & Decouple JSON-LD Streaming
## Summary

This PR optimizes the parallel ingestion pipeline, decouples JSON-LD streaming, and resolves critical connection pooling and CPU-blocking bottlenecks to speed up the GCS bulk upload phase.

## Key Changes
- GCS Upload Speedup: Replaced PyFilesystem GCSFS client (serialized by instance locks) with the native google-cloud-storage client. Configured the internal HTTP session pool size to match upload thread concurrency (32).
- Reduces GCS upload time by more than half
- Decoupled JSON-LD Exporter: Extracted JsonLdStreamDb into its own module (stats/jsonld_stream_db.py) to eliminate circular import dependencies between relational DB logic and GCS streaming exporters.
- CPU Bottleneck Fixes: Moved Pandas to_records() conversions out of the sequential generator into worker threads, and removed redundant blocking gc.collect() calls from chunk generators.
- Orchestration: Deferred workflow auto-trigger execution to the very end of Runner.run() to guarantee that Cloud Workflows/Dataflow starts only after all JSON-LD files are fully uploaded to GCS.

## Code Health & Thread Safety:
Wrapped child progress reporter updates with the parent reentrant lock.
Fixed a critical logging bug where failed files incorrectly reported Status
Moved all inline imports (threading, requests, tempfile) to top-level.


simple/stats/jsonld_stream_db.py
```diff
@@ -0,0 +1,384 @@
...
+    self.lock = threading.Lock()
+    self._obs_records = []
+    self._triples = []
+
+  def insert_observations(self, observations_df: pd.DataFrame,
+                          input_file: File):
+    if not observations_df.empty:
+      records = observations_df.to_records(index=False).tolist()
+      with self.lock:
+        self._obs_records.extend(records)
```
dwnoble, line 266: This dataframe looks like it could get pretty big and use a lot of memory. can it write to disk as it goes?

dwnoble (approved):

simple/stats/jsonld_stream_db.py
gmechali: Ooh good catch, thanks for finding this. We're definintely fine on scale for now, un data works without any issues. Let me take an AI to store on disk as we go for future resilience. I ll do it in a follow up.

PR #530, opened 2026-06-08
feat: Add place aggregations

SandeepTuniki:
I'm porting this PR from the "data" repo since the aggregation utils have been moved to this repo now.

I tested it against several test cases for place aggregations (see https://github.com/datacommonsorg/import/pull/536) and ensured its working. I'll create a separate PR for the test cases (it has hardcodings at the moment).

Note:
- I'll shortly come with up a way to configure it so that it only runs on base DC + specific DCP users who prefer it. For now, I haven't enabled the place aggregations to be run automatically.


pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
```diff
@@ -0,0 +1,128 @@
...
+            '2020' as date_val
+        );
+
+        -- Conceptual aggregation logic:
+        -- SELECT SUM(value_num) as state_population, date_val
+        -- FROM temp_county_populations
+        -- GROUP BY date_val;
+
+        SELECT 'geoId/06' as place, 'Count_Person' as variable, 39000000.0 as sum_population;
+        """
```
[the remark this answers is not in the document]
SandeepTuniki: Subsequent commit makes this suggestion obsolete.

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
```diff
@@ -0,0 +1,128 @@
...
+            '2020' as date_val
+        );
+
+        -- Conceptual aggregation logic:
+        -- SELECT SUM(value_num) as us_population, date_val
+        -- FROM temp_state_populations
+        -- GROUP BY date_val;
+
+        SELECT 'USA' as place, 'Count_Person' as variable, 330000000.0 as sum_population;
+        """
```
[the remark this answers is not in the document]
SandeepTuniki: Subsequent commit makes this suggestion obsolete.

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
```diff
@@ -0,0 +1,213 @@
...
+            spanner_options = '{{"table": "Observation_final_v2"}}' ) AS
+        SELECT 
+            "country/USA" AS entity1, 
+            ts.variable_measured,
+            ts.facets_id,
+            ts.extra_entities_id,
+            obs.date,
+            SUM(obs.double_value) AS double_value
+        FROM (
+          SELECT variable_measured, entity1, extra_entities_id, facets_id
```
rohitkumarbhagat: facet_id

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
```diff
@@ -0,0 +1,213 @@
...
+            FROM TimeSeries_final_v2
+            WHERE variable_measured = '{variable}'
+              AND STARTS_WITH(entity1, 'geoId/') 
+              AND LENGTH(entity1) = 8
+              AND JSON_VALUE(facets, '$.measurementMethod') = '{measurement_method}'
+            '''
+          )
+        ) ts
+        JOIN (
+          SELECT variable_measured, entity1, extra_entities_id, facets_id, date, double_value
```
rohitkumarbhagat: use value of type string. cast to double value

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
SandeepTuniki: Updated it to facet_id.

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
SandeepTuniki: Done. Thanks for spotting it.

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
```diff
@@ -0,0 +1,303 @@
...
+            '''
+            SELECT variable_measured, entity1, extra_entities_id, facet_id, date, value
+            FROM Observation
+            '''
+          )
+        ) obs ON 
+            ts.variable_measured = obs.variable_measured AND 
+            ts.entity1 = obs.entity1 AND 
+            ts.extra_entities_id = obs.extra_entities_id AND 
+            ts.facet_id = obs.facet_id
```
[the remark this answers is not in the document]
SandeepTuniki: This won't work as Databoost requires no joins in the spanner query.

pipeline/workflow/ingestion-helper/aggregation_test.py
```diff
@@ -243,6 +244,48 @@ def test_run_all(self):
...
+        self.assertIn("test-conn", query1)
+        self.assertIn("spanner-uri", query1)
+        self.assertIn("import1", query1)
+        self.assertIn('"country/USA" AS entity1', query1)
+
+        query2 = self.mock_executor.execute.call_args_list[1][0][0]
+        self.assertIn("test-conn", query2)
+        self.assertIn("spanner-uri", query2)
+        self.assertIn("import1", query2)
+        self.assertIn("SUBSTR(ts.entity1, 1, 8) AS entity1", query2)
```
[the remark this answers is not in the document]
SandeepTuniki: Fixed the unit test.

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
```diff
@@ -0,0 +1,303 @@
...
+        new_method_sql = f"""
+            IF(
+              ts.source_method IS NULL OR ts.source_method = '' OR ts.source_method = 'DataCommonsAggregate',
+              'DataCommonsAggregate',
+              CONCAT('dcAggregate/', ts.source_method)
+            )
+        """
+        
+        target_facet_id_sql = f"""
+            TO_HEX(MD5(CONCAT(
```
SandeepTuniki: Modify this to farm fingerprint hash

SandeepTuniki: Made the change.

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
```diff
@@ -0,0 +1,305 @@
...
+            IF(
+              ts.source_method IS NULL OR ts.source_method = '' OR ts.source_method = 'DataCommonsAggregate',
+              'DataCommonsAggregate',
+              CONCAT('dcAggregate/', ts.source_method)
+            )
+        """
+        
+        # 3. Deterministic Facet ID: Farm Fingerprint of (variable + new_method + new_provenance)
+        # We cast the resulting INT64 to STRING to match Spanner's schema.
+        target_facet_id_sql = f"""
```
vish-cs: Let's update it as discussed

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
```diff
@@ -0,0 +1,305 @@
...
+            FROM EXTERNAL_QUERY(
+              "{connection_id}",
+              '''
+              SELECT subject_id, object_id
+              FROM Edge
+              WHERE predicate = "containedInPlace"
+              '''
+            )
+            """
+        else:
```
vish-cs, line 98: Do we have a case not allow multiple to place?

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
```diff
@@ -0,0 +1,305 @@
...
+                variable_measured, 
+                facet_id, 
+                extra_entities_id,
+                JSON_VALUE(facet, '$.provenance') AS source_provenance,
+                JSON_VALUE(facet, '$.measurementMethod') AS source_method
+            FROM EXTERNAL_QUERY(
+              "{connection_id}",
+              '''
+              SELECT entity1, variable_measured, facet_id, extra_entities_id, facet
+              FROM TimeSeries
```
vish-cs: One thing to keep in mind that we are currently  overriding the default Observation/Timeseries table names with _final_v3 versions. So the queries may not work as is.

vish-cs (approved):

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
SandeepTuniki: Updated the facet id calculation per our discussion. It now follows the same pattern as https://github.com/datacommonsorg/import/blob/master/pipeline/data/src/main/java/org/datacommons/ingestion/data/TimeSeries.java

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
SandeepTuniki: These queries actually follow the schema of the `final_v3` version.

pipeline/workflow/ingestion-helper/aggregation/place_aggregation_generator.py
SandeepTuniki: Yes, the `allow_multiple_to_places` can take true of false in various scenarios.

This is the existing behaviour in the Flume pipeline, which I replicated in the query:

- When it is  false (for strict hierarchies like `County -> State`), the pipeline resolved duplicates by preferring the lexicographically first parent ID. Our query here uses `MIN(parent_id)` to match this behavior
- Where there are overlapping boundaries, such as rolling up zip codes to counties, or grid cells to states, we double count them to ensure both the parents calculate the common child's population count.

PR #537, opened 2026-06-11
Update embedding schema

shixiao-coder:
This PR updates the embedding schema based on the updated design document. It has below changes:

1. Having a new field embedding_type as primary key
2. the embedding content is now JSON formatted
3. We update DDL to be controlled by various parameter
4. Related test update and more coverage about spanner parameter IO with spanner client


pipeline/workflow/ingestion-helper/embedding_utils.py
```diff
@@ -101,18 +103,29 @@ def filter_and_convert_nodes(nodes_generator):
...
-
-
-def generate_embeddings_partitioned(database, nodes_generator, timeout):
+        name = node.get("name")
+        subject_id = node.get("subject_id")
+        if name:
+            embedding_content = json.dumps(OrderedDict([
+                ("title", subject_id),
+                ("text", OrderedDict([
+                    ("description", name)
```
clincoln8, line 112: should we call this name if it's the name field?
otherwise if we start to embed the description field then we'd need to call it something else

pipeline/workflow/ingestion-helper/schema.sql
```diff
@@ -128,10 +128,11 @@ CREATE INDEX VariableMeasuredObservationAbout ON Observation(variable_measured,
 
 CREATE TABLE NodeEmbedding (
```
clincoln8, line 129: will table name be configurable in a follow-up?

pipeline/workflow/ingestion-helper/embedding_utils.py
```diff
@@ -101,18 +103,29 @@ def filter_and_convert_nodes(nodes_generator):
...
     for node in nodes_generator:
-        if node.get("name"):
-            yield (node.get("subject_id"), node.get("name"), node.get("types"))
-
-
-def generate_embeddings_partitioned(database, nodes_generator, timeout):
+        name = node.get("name")
+        subject_id = node.get("subject_id")
+        if name:
+            embedding_content = json.dumps(OrderedDict([
```
clincoln8, line 109: should we add a todo to make the embedding content configurable one way or another

clincoln8 (approved):

PR #557, opened 2026-06-18
Optimize ingestion path for tfrecord files

vish-cs:
Special ingestion logic for tfrecord to create timeseries from tfrecord and skipping dedup. Also, added an option to skip wait operations between ingestion steps. This is used for the initial load of the data from the legacy pipelines into Spanner graph.

### DIFF
--- a/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
+++ b/pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
@@ -25,6 +25,7 @@ import org.datacommons.ingestion.util.GraphReader;
 import org.datacommons.ingestion.util.GraphTransformer;
 import org.datacommons.ingestion.util.PipelineUtils;
 import org.datacommons.proto.Mcf.McfGraph;
+import org.datacommons.proto.Mcf.McfOptimizedGraph;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
```diff
@@ -149,20 +150,29 @@ private static void processImport(
               "CreateEmptyEdgesWait-" + importName, Create.empty(TypeDescriptor.of(Void.class)));
     }
 
+    PipelineUtils.InputFormat format = PipelineUtils.resolveFormat(graphPath);
+    if (format == PipelineUtils.InputFormat.TFRECORD) {
+      TfRecordProcessingResult result =
```
rohitkumarbhagat, line 155: [optional] Can we add a check to prevent multi-entity ingestion via tfrecords? Seems like this flow does not support multi-entity fully. so better to throw error.

rohitkumarbhagat (approved):
Thanks Vishal!

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
vish-cs: tfrecord files are output from the legacy pipeline which itself does not support multi-entity schema.

PR #561, opened 2026-06-18
[DCP Ingestion] Bare minimum for multi-entity support in preprocessor

gmechali:
This is the first pass to get multi ingestion working with the preprocessor. 

There remains some testing required , mainly to ensure all columns are handled properly.


simple/stats/variable_per_row_importer.py
```diff
@@ -38,13 +39,17 @@
...
-    constants.COLUMN_ENTITY,
-    constants.COLUMN_VARIABLE,
-    constants.COLUMN_DATE,
-    constants.COLUMN_VALUE,
-]
-_DEFAULT_COLUMN_MAPPINGS = {x: x for x in _REQUIRED_COLUMNS}
+
+# Mapping from official DCID keys in config.json to internal importer columns.
+STANDARD_PROPERTY_MAPPING = {
+    "dcid:observationAbout": constants.COLUMN_ENTITY,                  # "entity"
```
dwnoble: do we need scaling factor here?

dwnoble (approved):

simple/stats/jsonld_stream_db.py
```diff
@@ -52,15 +52,18 @@
...
 
 def _uri_ref(val):
-  if not val:
+  if pd.isna(val) or val == "":
     return None
-  if is_uri_or_namespace(val):
-    return {"@id": val}
-  return {"@id": f"dcid:{val.lstrip('/')}"}
+  val_str = str(val).strip()
+  if val_str == "" or val_str.lower() in ("nan", "<na>"):
```
gmechali, line 59: This is because of how numpy imports the csv and then we fill it with nan if it's invalid on numeric processing.

We may want to revisit the CSV parsing longer term, but I'm okay with this for now. Lmk if you would prefer it be handled differently.

simple/stats/variable_per_row_importer.py
gmechali: Deferring scaling factor, But I do believe it needs to be added. Follow up PR will address

dwnoble (approved):

PR #579, opened 2026-06-23
[DCP SVG ] Seed the DB with c/g/Root and edge to dc/g/Root

gmechali:
Without this relationship, it's impossible to attach SVGs to the existing hierarchy.

Note: This is the first time we are seeding an edge. The preprocessor used to add the c/g/Root node, but now I think it belongs in seeding.


gmechali:
Closing this otu - we will rely on the customers defining their own root hierarchy

PR #581, opened 2026-06-24
Add option to skip transfoms

vish-cs:


pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
```diff
@@ -181,10 +181,16 @@ private static void processImport(
     }
 
     // Transform schema nodes (handle quantities, etc.)
-    PCollection<McfGraph> transformedGraph =
-        combinedGraph.apply(
-            "TransformNodes-" + importName,
-            org.apache.beam.sdk.transforms.ParDo.of(new GraphTransformer()));
+    PCollection<McfGraph> transformedGraph;
+    if (options.getSkipTransformation()) {
+      LOGGER.info("Skipping transformation step for import: {}", importName);
```
rohitkumarbhagat, line 186: Prefer counter instead of logging to avoid log flood. Do we need import name given pipeline runs per import?

rohitkumarbhagat (approved):

pipeline/ingestion/src/main/java/org/datacommons/ingestion/pipeline/GraphIngestionPipeline.java
vish-cs: A pipeline can run multiple imports. This log statement is per import so it will be just a small number.

pipeline/ingestion/src/test/java/org/datacommons/ingestion/pipeline/GraphIngestionPipelineTest.java
```diff
@@ -160,6 +160,65 @@ public void testBuildPipelineWithForceCombineNodes() throws Exception {
...
+    // "gender" is NOT in NON_CONSTRAINT_PROPS, so it would normally trigger constraintProperties
+    // generation.
+    boolean foundConstraintProperties = false;
+    for (Mutation m : capturedMutations) {
+      if (m.getTable().equals("Edge")
+          && m.asMap().get("subject_id").getString().equals("testStatVar")
+          && m.asMap().get("predicate").getString().equals("constraintProperties")) {
+        foundConstraintProperties = true;
+      }
+    }
```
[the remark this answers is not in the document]
vish-cs: Will do it as a follow up

PR #584, opened 2026-06-24
Add node counter to mcf import tool

vish-cs:
NumNodeSuccesses counter is used to EMPTY_IMPORT_CHECK validaiton. This counter is currently generated only during CSV processing. We need it for MCF processing flow as well for Schema imports etc.

### DIFF
--- a/tool/src/main/java/org/datacommons/tool/Processor.java
+++ b/tool/src/main/java/org/datacommons/tool/Processor.java
@@ -54,6 +54,7 @@ public class Processor {
   private final ExecutorService execService;
   private final LogWrapper logCtx;
   private HttpClient httpClient;
+  private int numMcfNodeSuccesses = 0;
 
   public static Integer process(Args args) throws IOException, TemplateException {
     Integer retVal = 0;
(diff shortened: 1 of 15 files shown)


tool/src/main/java/org/datacommons/tool/Processor.java
```diff
@@ -213,7 +220,9 @@ private void processLoadedGraph(Mcf.McfGraph n, Mcf.McfType type)
     if (existenceChecker != null && type == Mcf.McfType.INSTANCE_MCF) {
       existenceChecker.addLocalGraph(n);
```
ajaits, line 221: instance_mcf will have schema nodes from input mcf files. should the counter be updated here too?

ajaits (approved):

tool/src/main/java/org/datacommons/tool/Processor.java
vish-cs: The actual validation for these nodes happens later in the checkNodes(). I have already updated checkNodes() to increment the counter for these nodes.

PR #606, opened 2026-07-01
[DCP Ingestion] Update Ingestion History to have unique records per run and update the record by status / stage

gmechali:
This PR adds an environment variable to contrl whether to keep the current schema, and behavior for writing to Ingesiton History, or use the new proposed approach. 

We use a flag to be able to move over DCP customers and Base DC safely with the new versions.

The new proposal is for ingestionHistory to have unique records per run, and updates the status/stage thoroughly as it progresses through the workflow. This will allow a stronger visibility into the ingestion runs, and allow us to fix the issue with the computation of the stale timestamp


pipeline/workflow/ingestion-helper/clients/schema.sql
```diff
@@ -113,16 +114,18 @@ CREATE TABLE ImportStatus (
 ) PRIMARY KEY(ImportName);
 
 CREATE TABLE IngestionHistory (
-  CompletionTimestamp TIMESTAMP NOT NULL OPTIONS ( allow_commit_timestamp = TRUE ),
-  IngestionFailure Bool NOT NULL,
   WorkflowExecutionID STRING(1024) NOT NULL,
+  Timestamp TIMESTAMP OPTIONS ( allow_commit_timestamp = TRUE ),
+  IngestionFailure Bool,
+  Status STRING(1024),
```
dwnoble, line 121: Can you add constraints to `Status` and `Stage` with the valid string values? Something like:

```
CONSTRAINT ValidStatus CHECK(Status IN ('PENDING', 'RUNNING', 'SUCCESS', 'FAILED', 'RETRY'))
```

(see: https://docs.cloud.google.com/spanner/docs/check-constraint/how-to)

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -282,42 +282,64 @@ def _update(transaction: Transaction):
...
 
         logging.info(
-            f"Updating IngestionHistory table for workflow {workflow_id}")
+            f"Updating IngestionHistory table for workflow {workflow_id} with status {status}, stage {stage}")
+
+        def _update(transaction: Transaction):
+            columns = ["WorkflowExecutionID", "Status", "Stage", "Timestamp"]
+            values = [workflow_id, status, stage, spanner.COMMIT_TIMESTAMP]
+
+            if status in ('SUCCESS', 'FAILURE'):
```
dwnoble: Can you add a comment explaining the rationale here?

pipeline/workflow/ingestion-helper/routes/imports.py
```diff
@@ -29,8 +29,14 @@
 class IngestionState(str, Enum):
     SUCCESS = "SUCCESS"
     PENDING = "PENDING"
+    RUNNING = "RUNNING"
+    FAILURE = "FAILURE"
     RETRY = "RETRY"
```
dwnoble: When is `RETRY` used?

pipeline/workflow/ingestion-helper/clients/schema.sql
```diff
@@ -113,16 +114,18 @@ CREATE TABLE ImportStatus (
 ) PRIMARY KEY(ImportName);
 
 CREATE TABLE IngestionHistory (
-  CompletionTimestamp TIMESTAMP NOT NULL OPTIONS ( allow_commit_timestamp = TRUE ),
-  IngestionFailure Bool NOT NULL,
   WorkflowExecutionID STRING(1024) NOT NULL,
+  Timestamp TIMESTAMP OPTIONS ( allow_commit_timestamp = TRUE ),
+  IngestionFailure Bool,
```
dwnoble, line 120: Why remove NOT NULL?

pipeline/workflow/ingestion-helper/clients/schema.sql
```diff
@@ -113,16 +114,18 @@ CREATE TABLE ImportStatus (
 ) PRIMARY KEY(ImportName);
 
 CREATE TABLE IngestionHistory (
-  CompletionTimestamp TIMESTAMP NOT NULL OPTIONS ( allow_commit_timestamp = TRUE ),
-  IngestionFailure Bool NOT NULL,
   WorkflowExecutionID STRING(1024) NOT NULL,
+  Timestamp TIMESTAMP OPTIONS ( allow_commit_timestamp = TRUE ),
```
dwnoble: Can we track both creation & completion timestamps?

pipeline/workflow/ingestion-helper/clients/schema.sql
gmechali: I actually just replaced those to be enums. Should be better this way!

pipeline/workflow/ingestion-helper/clients/spanner.py
gmechali: Added. These 3 ifs are to control partial udpates where we only update the fields that are set.

pipeline/workflow/ingestion-helper/routes/imports.py
gmechali: That was legacy there. I m leaving it for now, but would like to know from Vishal. Will see if his retry script relies on this. More to come.

pipeline/workflow/ingestion-helper/clients/schema.sql
gmechali: I think the ingestionFailure state is meant to represent its final state.
when we write a "pending" row, it doesn't really make sense for us to set the ingestion failure to true or false. So it stays null for now. We write it only once the status is a final state of success/failure.

Though, I'm not sure now that we have a more descriptive "State" column, whether it makes sense to keep ingestionFailure. @vish-cs wdyt.

pipeline/workflow/ingestion-helper/clients/schema.sql
gmechali: Yup great point. Done.

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -282,42 +282,64 @@ def _update(transaction: Transaction):
...
             metrics: A dictionary containing metrics about the ingestion.
         """
 
         logging.info(
-            f"Updating IngestionHistory table for workflow {workflow_id}")
+            f"Updating IngestionHistory table for workflow {workflow_id} with status {status}, stage {stage}")
+
+        def _update(transaction: Transaction):
+            columns = ["WorkflowExecutionID", "Status", "Stage", "Timestamp"]
+            values = [workflow_id, status, stage, spanner.COMMIT_TIMESTAMP]
```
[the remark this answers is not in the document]
gmechali: Done.

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -282,42 +282,64 @@ def _update(transaction: Transaction):
...
 
-    def update_ingestion_history(self, workflow_id: str, job_id: str,
-                                 ingested_imports: list, metrics: dict):
+    def update_ingestion_history(self,
+                                 workflow_id: str,
+                                 status: str,
+                                 stage: str,
+                                 job_id: str = None,
+                                 ingested_imports: list = None,
+                                 metrics: dict = None):
```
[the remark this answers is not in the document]
gmechali: Done.

pipeline/workflow/ingestion-helper/routes/imports.py
```diff
@@ -81,20 +96,51 @@ def get_import_info(req: ImportInfoRequest, spanner: SpannerClient = Depends(get
...
         spanner.update_import_version_history(import_list_dicts, req.workflowId)
     return BaseResponse(status=ResponseStatus.OK)
 
+@router.post("/ingestion-history", response_model=BaseResponse)
+def update_ingestion_history(req: UpdateIngestionHistoryRequest, spanner: SpannerClient = Depends(get_spanner_client)):
+    """Updates the ingestion history record for the workflow execution."""
+    ingested_imports = _extract_import_names(req.importList)
+    
+    metrics = None
+    if req.status in (IngestionState.SUCCESS, IngestionState.FAILURE) and req.stage == IngestionStage.DATAFLOW and req.jobId:
```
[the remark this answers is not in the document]
gmechali: Done.

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -282,42 +282,64 @@ def _update(transaction: Transaction):
...
-            f"Updating IngestionHistory table for workflow {workflow_id}")
+            f"Updating IngestionHistory table for workflow {workflow_id} with status {status}, stage {stage}")
+
+        def _update(transaction: Transaction):
+            columns = ["WorkflowExecutionID", "Status", "Stage", "Timestamp"]
+            values = [workflow_id, status, stage, spanner.COMMIT_TIMESTAMP]
+
+            if status in ('SUCCESS', 'FAILURE'):
+                columns.append("IngestionFailure")
+                values.append(status == 'FAILURE')
```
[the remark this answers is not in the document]
gmechali: Done.

pipeline/workflow/ingestion-helper/config.py
```diff
@@ -75,3 +75,5 @@
 REDIS_PORT = os.environ.get('REDIS_PORT', '6379')
 GCS_OUTPUT_PREFIX = os.environ.get('GCS_OUTPUT_PREFIX', '')
 
+ENABLE_UNIQUE_INGESTION_RUNS = os.environ.get('ENABLE_UNIQUE_INGESTION_RUNS', 'false').lower() == 'true'
```
dwnoble, line 80: Can you add a comment describing what this env variable does, and when it should be phased out

pipeline/workflow/ingestion-helper/routes/imports.py
```diff
@@ -81,20 +85,61 @@ def get_import_info(req: ImportInfoRequest, spanner: SpannerClient = Depends(get
...
+def update_ingestion_history(req: UpdateIngestionHistoryRequest, spanner: SpannerClient = Depends(get_spanner_client)):
+    """Updates the ingestion history record for the workflow execution."""
+    if not config.ENABLE_UNIQUE_INGESTION_RUNS:
+        logging.warning("Received update_ingestion_history request but ENABLE_UNIQUE_INGESTION_RUNS is disabled. No-op.")
+        return BaseResponse(status=ResponseStatus.OK)
+
+    ingested_imports = _extract_import_names(req.importList)
+    
+    metrics = None
+    if req.status in (IngestionState.SUCCESS, IngestionState.RETRY) and req.stage == IngestionStage.DATAFLOW and req.jobId and req.jobId != "N/A":
```
dwnoble, line 128: Can you add a comment explaining why we do this logic check?

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -282,46 +297,130 @@ def _update(transaction: Transaction):
...
             if self.graph_database and self.graph_database.name != self.database.name:
                 self.graph_database.run_in_transaction(_insert)
             logging.info(
-                f"Updated IngestionHistory table for workflow {workflow_id}")
+                f"Updated IngestionHistory table (v1) for workflow {workflow_id}")
+        except Exception as e:
+            logging.error(f'Error updating IngestionHistory table (v1): {e}')
+            raise
+
+    def update_ingestion_history_v2(self,
```
dwnoble, line 349: It's possible that the user set ENABLE_UNIQUE_INGESTION_RUNS to true, but is still running the old schema version. 

can you add a check here that ensures we're running schema v2 (maybe it can check if the CreationTimestamp column exists)

This can be a follow-up PR, but i think it would be also worht creating a new table that tracks what 'version' of the schema we're running. So for example, if these schema changes migrate us from version "1" to version "2", we can just check that we're running at least version "2" before calling this method-what do you think?

pipeline/workflow/ingestion-helper/config.py
gmechali: Added.

pipeline/workflow/ingestion-helper/routes/imports.py
gmechali: done.

pipeline/workflow/ingestion-helper/clients/spanner.py
gmechali: The only cases for this would be our internal testing accounts, I dont think it's worth helping us migrate.

For real customers that were using the latest milestones, once we cut over a new version of the CLI / terraform, I think it'd be great if we could tell them to drop their db and start over. If that's not possible, we can just give them the statement to alter their DB to match.

What you're describing on version migration sounds great to track but I'd argue pretty low prio against the current task list. wdyt? 
The schema should** stagnate.

dwnoble: that sgtm

dwnoble (approved):

PR #613, opened 2026-07-06
Add  endpoint to rollback failed imports

vish-cs:


pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,378 @@
...
+        if not status_rows:
+            logging.warning(f"No ImportStatus found for '{short_name}'.")
+            return False
+
+        current_latest_version_path = status_rows[0][0]
+        if not current_latest_version_path:
+            logging.warning(f"ImportStatus.LatestVersion is empty for '{short_name}'.")
+            return False
+
+        current_version_name = os.path.basename(current_latest_version_path)
```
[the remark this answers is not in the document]
vish-cs: Done

pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,378 @@
...
+        if not rows_prev:
+            logging.warning(
+                f"No previous different version found for '{short_name}'. Cannot revert.")
+            return False
+
+        previous_version = rows_prev[0][0]
+
+        # Reconstruct full path for LatestVersion
+        parent_dir = os.path.dirname(current_latest_version_path)
+        new_latest_version_path = os.path.join(parent_dir, previous_version)
```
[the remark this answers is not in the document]
vish-cs: done

pipeline/scripts/retry_failed_ingestion_test.py
```diff
@@ -0,0 +1,188 @@
...
+# limitations under the License.
+
+import unittest
+from unittest.mock import MagicMock, patch
+import sys
+import os
+
+# Import the script functions
+import retry_failed_ingestion
+from google.cloud import spanner
```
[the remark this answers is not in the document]
vish-cs: Removed the test file

pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,378 @@
...
+            location=location,
+            jobId=job_id,
+            minimumImportance='JOB_MESSAGE_ERROR').execute()
+
+        for msg in messages.get('jobMessages', []):
+            text = msg.get('messageText', '')
+            # If we find a valid import name in the error message, add it.
+            for name in all_import_names:
+                if name in text:
+                    failed_imports.add(name)
```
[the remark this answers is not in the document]
vish-cs: Done

pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,378 @@
...
+    logging.info(f"Processing failed imports: {failed_imports}")
+    spanner_client = spanner.Client(project=FLAGS.spanner_project)
+    instance = spanner_client.instance(FLAGS.spanner_instance)
+    database = instance.database(FLAGS.spanner_database)
+
+    can_retrigger = True
+    for imp in failed_imports:
+        status = revert_import(database, imp, FLAGS.job_id, dry_run=FLAGS.dry_run)
+        if status == "ALREADY_REVERTED":
+            can_retrigger = False
```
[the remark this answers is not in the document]
vish-cs: Done

pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,382 @@
...
+import sys
+
+from absl import app
+from absl import flags
+from absl import logging
+from google.cloud import spanner
+from googleapiclient.discovery import build
+from googleapiclient.errors import HttpError
+
+FLAGS = flags.FLAGS
```
gmechali: What do you think about putting this logic in the ingestion helper? 

So that retrying a failed import, can just be executing a post call to the ingestion helper with a different URL param, and it would process all this logic you have here?

That way, it isnt a separate script/image, we can keep it well contained in the ingestion helper image, and use the workflow to trigger it just like everything else.

pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,382 @@
...
+# limitations under the License.
+"""Script to automate the retry process for failed Dataflow ingestion jobs.
+
+This script identifies failed imports within a specific Dataflow job,
+reverts the failed imports in the Spanner database to their last known
+good version, resets any 'RETRY' imports back to 'STAGING', and
+optionally retriggers the Spanner ingestion workflow.
+
+Usage:
+    python3 retry_failed_ingestion.py --job_id=<DATAFLOW_JOB_ID> \
```
gmechali: Is this the expected usage of this script? manual triggers?
Or is the expectation that the workflow automaitcally does this?

pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,382 @@
...
+flags.DEFINE_string("spanner_instance", "dc-kg-test", "Spanner Instance ID")
+flags.DEFINE_string("spanner_database", "dc_graph_import",
+                    "Spanner Database ID")
+flags.DEFINE_string("workflow_name", "spanner-ingestion-workflow",
+                    "Workflow name")
+flags.DEFINE_bool("dry_run", False,
+                  "Dry run mode. Display actions without executing them.")
+
+
+def get_failed_imports(dataflow, project_id, location, job_id):
```
gmechali: I would probably recommend not just putting this as a single file script under scripts/

It's already almost 400 lines, and very likely to grow in the future. I would recommend keeping a main.py as the simple entry point, and helpers to navigate the complexity. This also seems complex enough that it warrants tests imo. 

I opened anohter comment asking if this should just be in the ingestion-helper. Curious for your thoughts on that?

pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,382 @@
...
+                all_import_names = {
+                    i['importName'].split(':')[-1] for i in all_imports_data
+                }
+                logging.info(
+                    f"Imports involved in this job: {all_import_names}")
+            except Exception as e:
+                logging.error(f"Error parsing importList parameter: {e}")
+
+        logging.info(f"Fetching error messages for Dataflow job {job_id}...")
+        request = dataflow.projects().locations().jobs().messages().list(
```
gmechali: This seems to only be checking dataflow for failures. What if the failures came from elsewhere in the worfklow (spanner embeddings, ingestion helper issue, bq federation).

What's the long term plan for this, should we be checking all the stages? Should the workflow be modified to propoagate errors from dataflow? (if that's even possible)

pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,382 @@
...
+        True if the revert was successful, 'ALREADY_REVERTED' if the import
+        was recently reverted, or False if no version history was found or an
+        error occurred.
+    """
+    short_name = import_name.split(':')[-1]
+
+    def _revert_txn(transaction):
+        # 1. Get current version from ImportStatus to know what to avoid.
+        status_sql = """
+            SELECT LatestVersion FROM ImportStatus
```
gmechali: Given the changes we discussed to IngestionHistory last week, isnt that a better place to read this from? ImportStatus seems brittle given that if we have started a new workflow since, that information is gone.

pipeline/scripts/retry_failed_ingestion.py
```diff
@@ -0,0 +1,382 @@
...
+            WHERE ImportName = @importName
+        """
+        status_results = transaction.execute_sql(
+            status_sql,
+            params={'importName': short_name},
```
gmechali: How would this work for a multi-provenance import? 
Like if I import WHO and ILO at once?

pipeline/scripts/retry_failed_ingestion.py
vish-cs: That makes sense. I have refactored and moved it inside the ingestion helper. Thanks for the suggestion!

pipeline/scripts/retry_failed_ingestion.py
vish-cs: We are planning a manual trigger for base DC at the moment. This will be with the oncall. Once stabilized, we can automate it in the workflow. DCP can still add a step in the workflow to invoke it on a failure (using the ingestion helper API).

pipeline/scripts/retry_failed_ingestion.py
vish-cs: Yes, the logic is now within separate helper files with corresponding test cases.

pipeline/scripts/retry_failed_ingestion.py
vish-cs: The script checks for the dataflow logs to see if it finds a specific failed import and reverts it. Else, it falls back to reverting all the imports in that batch.

pipeline/scripts/retry_failed_ingestion.py
vish-cs: Yes, we should be taking bq failures also into account. It will require us to track aggregation status into spanner. Looking into this.

pipeline/scripts/retry_failed_ingestion.py
vish-cs: I was not aware of your PR when I raised this PR. I have gone through your update. My understanding is that the IngesitonHistory table does not include per-import version information so we will need to refer to another table for retrieving that.

pipeline/scripts/retry_failed_ingestion.py
vish-cs: I am also thinking that it may be simpler to revert the whole batch as a v1 for the rollback flow.

pipeline/scripts/retry_failed_ingestion.py
vish-cs: Updated the script to use the new schema tables

pipeline/scripts/retry_failed_ingestion.py
vish-cs: The new implementation reverts the whole batch

pipeline/scripts/retry_failed_ingestion.py
vish-cs: The new implementation is an endpoint within ingestion helper. It takes a workflow id as an argument and reverts all the imports for that workflow. It can be invoked manually or from the ingestion workflow in the failure handing logic for any type of failures.

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -456,6 +455,199 @@ def check_failed_imports(self) -> bool:
...
+                    param_types={'workflowId': STRING}
+                )
+                rows = list(results)
+                if rows and rows[0][0]:
+                    return list(rows[0][0])
+        except Exception as e:
+            logging.error(f"Error querying IngestionHistory for workflow '{workflow_id}': {e}")
+        return []
+
+    def revert_import(self,
```
gmechali: Can you move most of this logic in a separate file in util/rollback_helper.py or something?

And just keep the spanner code in here, so get_imports_for_workflow, fetch_import_version_history, set_rollback_status etc..

I think this spanner client should just be simple helper functions to execute spanner queries. The actual logic for how these helpers are called belong elsewhere.

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -456,6 +455,199 @@ def check_failed_imports(self) -> bool:
             logging.error(f'Error checking for retry imports: {e}')
             return True
 
+    def get_imports_for_workflow(self, workflow_id: str) -> list[str]:
```
gmechali, line 460: ultra nit: "from" instead of for?

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -456,6 +455,199 @@ def check_failed_imports(self) -> bool:
...
+            """
+            results_history = transaction.execute_sql(
+                sql_history,
+                params={'importName': short_name},
+                param_types={'importName': STRING})
+            rows_history = list(results_history)
+
+            if not rows_history:
+                logging.warning(
+                    f"No version history found in ImportVersionHistory for '{short_name}'. Cannot revert.")
```
gmechali: So what should happen here to the database? 

do we leave the run in FAILED state, the staleness timestamp should read before this fialed run start. Ideally they get alerted of this failuyre and have to manually intervene, is that right?

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -456,6 +455,199 @@ def check_failed_imports(self) -> bool:
...
+                return False, None
+
+            latest_version = rows_history[0][0]
+            latest_comment = rows_history[0][1] if len(rows_history[0]) > 1 else ""
+
+            if latest_comment and ("Reverted due to Dataflow failure" in latest_comment or "Reverted batch workflow" in latest_comment):
```
gmechali: Using these string comments seems super dangerous to me. Shouldn't we use the Status instead?

It's really not transparent that if you modify the comment, you ll modify actual workflow behavior. You know this now, but whoever is working on this in 9 months won't. Can't we just fetch the Status on IngestionHistory row, add a new rollback status and use that to determine the state?

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -456,6 +455,199 @@ def check_failed_imports(self) -> bool:
...
+                logging.error(
+                    f"Import '{short_name}' was already reverted recently. Aborting to prevent infinite loop."
+                )
+                return "ALREADY_REVERTED", None
```
gmechali: What exactly is this function supposed to return? I see you returning a True/False  or now thing "ALREADY_REVERTED" string.

pipeline/workflow/ingestion-helper/routes/imports.py
```diff
@@ -91,6 +91,17 @@ class ImportInfoItem(BaseModel):
...
+class RollbackRequest(BaseModel):
+    workflowId: str = Field(description="Workflow execution ID to rollback")
+    projectId: Optional[str] = Field(default=None, description="GCP Project ID override")
+    location: Optional[str] = Field(default=None, description="GCP Location override")
+    workflowName: str = Field(default="spanner-ingestion-workflow", description="Workflow name to retrigger")
+    dryRun: bool = Field(default=False, description="Dry run mode")
+
+class RollbackResponse(BaseResponse):
+    revertedImports: Dict[str, str] = Field(default_factory=dict, description="Map of import names to their reverted versions")
+    dryRun: bool = Field(default=False, description="Whether this was executed in dry run mode")
```
gmechali: Why do we need the dry run in the response?

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -456,6 +455,199 @@ def check_failed_imports(self) -> bool:
...
+
+            previous_version = None
```
gmechali: I kind of disagree with the architecture of this. I think the rollback shouldn't be triggered by the ingestion helper. I think the ingestion helper should be called to give the worfklow the information it needs to do the rollback.

I think we can use a subworkflow to package everything from updating the ingesiton_history, to dataflow, to aggregations.

Then the worfklow, on failure, can aks the ingestion Helper for the graphPath of the last successful ingestions for these imports. Then the worfklow can re-update the ingestionHistory/Importversionhistory statuses, re-laucnh dataflow with new params, relaunch postprocessing etc...

Otherwise, with this implementation, we're just complete obfuscating the process for the rollback.

pipeline/workflow/ingestion-helper/utils/imports.py
```diff
@@ -177,3 +178,117 @@ def get_ingestion_metrics(project_id, location, job_id):
...
+    Returns:
+        The name of the new workflow execution, or None if dry run or error.
+    """
+    if dry_run:
+        logging.info(f"[DRY RUN] Would retrigger workflow '{workflow_name}' in project '{project_id}' location '{location}' with argument: {{'is_retry': true}}")
+        return None
+
+    logging.info(f"Retriggering workflow '{workflow_name}'...")
+    try:
+        service = build('workflowexecutions', 'v1', cache_discovery=False)
```
gmechali: I dont think we should make this a separate workflow, but a subworkflow.

The first workflow acquires the lock, I believe it should NOT be allowed to release the lock until the database is back in a good state post rollback. That's why I think it should also be a subworkflow.

gmechali (commented):
hey Vishal, overall I think the major concern I have is that I think the ingesiton helper is being used for too much. It shouldn't contain such major implementation logic, I think all that should live within the workflow, and the ingestion helper should just give the workflow the information it needs (like the last successful graphpath).

Happy to talk this through if you want. lmk what you think

PR #607, opened 2026-07-02
feat: Support configurable aggregations setup

SandeepTuniki:
This PR adds support to specify config files that will be used to run the aggregations.

This PR supersedes https://github.com/datacommonsorg/import/pull/582 as per the offline discussion & feedback I received on the previous PR's design approach.

PR #614, opened 2026-07-06
refactor: remove legacy aggregation routes and code  from ingestion-helper

SandeepTuniki:
This is a follow-up of PR https://github.com/datacommonsorg/import/pull/607.

All the aggregations related code is moved into `pipeline/workflow/aggregation-helper` folder.

So we don't need to keep the same code in `import/pipeline/workflow/ingestion-helper` and can be removed.

Note: I will send this PR to review once we confirm that no other code (especially DCP related) is referencing this.


SandeepTuniki:
Fixed in another PR https://github.com/datacommonsorg/import/pull/676

PR #618, opened 2026-07-08
add embedding geneator to ingest embedding with BQ

shixiao-coder:
This PR added the embedding generation logic into BQ federation. It will process a read from Spanner first to extract the related nodes and trigger other filtering logic and link with BQ model to generate embeddings, at last it will export the generated embeddings back into the spanner


pipeline/workflow/aggregation-helper/aggregation/aggregation_test.py
```diff
@@ -24,6 +24,9 @@
 from aggregation import LinkedEdgeGenerator
 from aggregation import ProvenanceSummaryGenerator
 from aggregation import PlaceAggregationGenerator
+from aggregation import EmbeddingGenerator
+from aggregation.embedding_generator import EmbeddingSpec
+from aggregation.sql_utils import _escape_sql_literal
```
SandeepTuniki: I renamed `sql_utils.py` to `common.py` in a PR earlier in the day, so this import is non-existent. You can remove this line.

pipeline/workflow/aggregation-helper/aggregation/embedding_generator.py
```diff
@@ -0,0 +1,209 @@
...
+def _extract_nl_stat_var() -> List[str]:
+    path = _NL_STAT_VAR_FILE
+    content = ""
+    if path.startswith("gs://"):
+        try:
+            import urllib.request
+            url = "https://storage.googleapis.com/" + path[5:]
+            with urllib.request.urlopen(url) as resp:
+                content = resp.read().decode("utf-8")
+        except Exception as e:
```
SandeepTuniki: Is it possible to avoid using `Exception` and use a more specific sub-class so that other non-HTTP related errors aren't consumed here? Alternatively, you could re-raise the exception after handling the error.

pipeline/workflow/aggregation-helper/aggregation/embedding_generator.py
```diff
@@ -0,0 +1,209 @@
...
+    def __init__(self,
+                 executor: BigQueryExecutor,
+                 is_base_dc: bool = True) -> None:
+        """Initializes the EmbeddingGenerator with the executor."""
+        self.executor = executor
+        self.is_base_dc = is_base_dc
+
+    def run_all(self, specs: Optional[List[Any]] = None, embedding_table: str = "NodeEmbedding") -> List[bigquery.job.QueryJob]:
+        """Runs all embedding generations asynchronously and returns their jobs."""
+        enable_embeddings = os.environ.get('ENABLE_EMBEDDINGS', 'false').lower() == 'true'
```
SandeepTuniki: Is it possible to avoid processing env vars in the generator? Perhaps we could extract it from `main.py`/`orchestrator.py` and pass them here, so that these classes purely depend on inputs passed. That makes writing tests for them a lot easier.

SandeepTuniki: And speaking of tests, could you add some e2e tests in the below folder for these embeddings? :) 

https://github.com/datacommonsorg/import/tree/master/pipeline/workflow/aggregation-helper/aggregation/e2e_tests

In my experience, BQ federation over spanner has many moving parts, so it breaks too often and any changes are hard to verify without some automated e2e tests.

pipeline/workflow/aggregation-helper/aggregation/embedding_generator.py
```diff
@@ -0,0 +1,209 @@
...
+    def run_embedding_spec(self, spec: Any, embedding_table: str = "NodeEmbedding") -> Optional[bigquery.job.QueryJob]:
+        """Runs the embedding generation query for a single spec."""
+        if isinstance(spec, dict):
+            spec = EmbeddingSpec(**spec)
+
+        dest = self.executor.get_spanner_destination_uri()
+        conn_id = self.executor.connection_id
+        embedding_conn_id = os.environ.get('BQ_MODEL_CONNECTION')
+        project_id = self.executor.project_id
+        bq_dataset_id = os.environ.get('BQ_DATASET_ID', 'datacommons')
```
SandeepTuniki: Same comment about env vars as previously noted.

pipeline/workflow/aggregation-helper/aggregation/orchestrator.py
```diff
@@ -211,6 +215,30 @@ def run(self, active_imports: List[str], dry_run: bool = True, skip_deletions: b
...
+                step_type = calc.get("type")
+                if dry_run:
+                    logging.info(f"[DRY RUN] Would execute global step: {calc.get('name', step_type)}")
+                else:
+                    logging.info(f"Triggering global step: '{step_type}'...")
+                    step_jobs = self._dispatch_stage_steps(calc)
+                    if step_jobs:
+                        job_ids = [job.job_id for job in step_jobs if hasattr(job, "job_id")]
+                        logging.info(f"Submitted {len(job_ids)} global job(s): {job_ids}")
+                        self._wait_for_jobs(
```
SandeepTuniki: Once the job finishes, could you add the result to `run_result` like the previous loop does? This will be useful in debugging and status reporting.

SandeepTuniki (approved):

pipeline/workflow/aggregation-helper/aggregation/aggregation_test.py
shixiao-coder: Sure, removed this import

pipeline/workflow/aggregation-helper/aggregation/embedding_generator.py
shixiao-coder: change this to more specific exceptions

pipeline/workflow/aggregation-helper/aggregation/embedding_generator.py
shixiao-coder: Updated the os env get into base.py and move all related params into executor to be called through orchestrator

also added e2e tests

pipeline/workflow/aggregation-helper/aggregation/embedding_generator.py
shixiao-coder: updated like above

pipeline/workflow/aggregation-helper/aggregation/orchestrator.py
shixiao-coder: Sure, added the result to run_result

PR #627, opened 2026-07-09
ci(pipeline): automatically tag ingestion pipeline image as latest

yiyche:
last `latest` tag was updated jun 18. seems unintended?

### DIFF
--- a/pipeline/ingestion/cloudbuild.yaml
+++ b/pipeline/ingestion/cloudbuild.yaml
@@ -40,6 +40,18 @@ steps:
       - '--env'
       - 'FLEX_TEMPLATE_JAVA_MAIN_CLASS=org.datacommons.ingestion.pipeline.GraphIngestionPipeline'
     id: 'build-flex-template'
+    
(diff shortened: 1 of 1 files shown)


dwnoble (approved):

pipeline/ingestion/cloudbuild.yaml
```diff
@@ -40,6 +40,18 @@ steps:
       - '--env'
       - 'FLEX_TEMPLATE_JAVA_MAIN_CLASS=org.datacommons.ingestion.pipeline.GraphIngestionPipeline'
     id: 'build-flex-template'
+
```
vish-cs, line 43: Just tagging the image as 'latest' does not tag the dataflow template. This is where we copy the template file to generate the stable template

https://github.com/datacommonsorg/import/blob/master/pipeline/workflow/update-version.yaml

yiyche:
Thanks Vishal for sharing the pointer! I misunderstood how all of this works

PR #632, opened 2026-07-13
Keep JSON-LD subjects intact across shard boundaries

rohitkumarbhagat:
## Summary

  - Prevent JSON-LD node sharding from splitting triples for the boundary subject.
  - Allow shards to exceed the target size when needed to keep a contiguous subject block intact.
  - Preserve progressive memory cleanup without sorting or duplicating triples.
  - Add regression coverage for a subject crossing the nominal chunk boundary.

### DIFF
--- a/simple/stats/jsonld_stream_db.py
+++ b/simple/stats/jsonld_stream_db.py
@@ -379,7 +379,7 @@ class JsonLdStreamDb(Db):
       self._obs_records[import_name].clear()
 
   def _generate_node_chunks(self, import_name: str, import_temp_dir: str):
-    """Generates node chunks of size _CHUNK_SIZE, cleaning memory progressively."""
+    """Generates node chunks targeting _CHUNK_SIZE without splitting subjects,  cleaning memory progressively."""
     triples = self._triples.get(import_name, [])
     shard_index = 0
     while triples:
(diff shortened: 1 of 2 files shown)


simple/stats/jsonld_stream_db.py
```diff
@@ -379,14 +379,21 @@ def _generate_observation_chunks(self, import_name: str,
...
     shard_index = 0
+    # Keep subjects intact when chunking. A chunk may exceed the target size if
+    # a subject has triples on both sides of the boundary.
     while triples:
       chunk = []
       # Pop from the end to avoid O(N) list shifting overhead
       for _ in range(min(_CHUNK_SIZE, len(triples))):
         chunk.append(triples.pop())
+
+      boundary_subject = chunk[-1].subject_id
```
gmechali, line 393: Can you explain how you've tested this. Did it work the CS data that previously failed.

I thought the issue was more due to memory constraints, not downstream dataflow failures. 
Generally agree with not splitting subjects at the boundary though, so approving this PR.

gmechali (approved):

simple/stats/jsonld_stream_db.py
rohitkumarbhagat: OOM is a separate issue. Merging this for now. Thanks Gabe.

PR #633, opened 2026-07-13
[Ingestion ] Fixes the backwards compatibility of update_import_version calls

gmechali:
The previous PR made a mistake on backwards compatibility, one of the changes required a new column.

This PR ensures this is fully backwards compatible. Later this week, we will update the Schema, turn on ENABLE_UNIQUE_INGESTION_RUNS and get rid of the else branches in this code + in mixer.


clincoln8 (approved):
Thanks!

PR #636, opened 2026-07-14
fix(simple): preserve case of provenance ID in import name

clincoln8:
## Description
Preserve the casing of the Provenance ID in the import name. 

Currently, `import_name()` applies `.lower()` to the provenance ID. This causes a mismatch during the post-processing aggregation phase:
1. `LinkedEdgeGenerator` queries `Edge` (expects normalized case).
2. `ProvenanceSummaryGenerator` queries `TimeSeries` (expects case-preserved ID, e.g. `provenance/FrogCensus`).

By preserving the casing, both generators can match their respective rows and successfully generate aggregations.

Co-requisite: https://github.com/datacommonsorg/datacommons/pull/166


gmechali (approved):
Thanks Christie!

PR #639, opened 2026-07-14
Rename observationProperties to statVarProperties in SVG generation

n-h-diaz:


vish-cs (approved):
Need to update it in the schema repo as well.

PR #645, opened 2026-07-15
add embedding label index

shixiao-coder:
The current vector search contains a lot of conditions during the search and thus causing the vector search to do hard join from search on just index with actual embedding, content, type data and causing a large latency when doing to search. The new index stores the fields and avoid the hard join to boost the speed


pipeline/workflow/ingestion-helper/clients/schema.sql
```diff
@@ -188,6 +188,9 @@ OPTIONS (
   distance_type = 'COSINE'
 );
 
+CREATE INDEX {{ embedding_label_index }}
+ON {{ embedding_table }}(embedding_label) STORING (embedding_content, embeddings, node_types);
```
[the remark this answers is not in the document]
shixiao-coder: the embeddings are needed

pipeline/workflow/ingestion-helper/clients/schema.sql
```diff
@@ -211,6 +211,9 @@ OPTIONS (
   distance_type = 'COSINE'
 );
 
+CREATE INDEX {{ embedding_label_index }}
```
keyurva, line 214: Do we need to hint using this index when querying spanner from mixer?

keyurva (approved):
Thanks!

Can you add a PR description for future reference?

pipeline/workflow/ingestion-helper/clients/schema.sql
shixiao-coder: from the past test, no. But I think it might be good to force using this index to make the search stable (at least for now)

PR #656, opened 2026-07-17
Refactor Spanner embedding generation to run ML.PREDICT outside write transaction

dwnoble:
### Summary
This change refactors the Spanner embedding ingestion helper to separate remote machine learning inference (`ML.PREDICT`) from Spanner read-write transactions.

### Problem
Previously, the embedding generator executed `INSERT OR UPDATE ... SELECT FROM ML.PREDICT(...)` within a Spanner read-write transaction. Because `ML.PREDICT` makes external network calls to Vertex AI, the write transaction remained open for a prolonged period (~85 seconds in a typical bulk ingestion batch of 1,000 nodes). 

During bulk imports (such as the `undata` dataset), Spanner frequently runs high-priority background range splits and directory migrations (`_MoveRange`). The long-held write locks from our transaction repeatedly conflicted with these range moves, causing Spanner to abort the user transaction with a 409 conflict:
`Transaction was aborted. It was wounded by a higher priority transaction with tag "_MoveRange" due to conflict on keys with prefix [<internal>], column state in table split.`

### Solution
1. **Uncouple Prediction from Transaction**: Run the `ML.PREDICT` query as a read-only query outside the read-write transaction using a database snapshot. This fetches the generated embeddings into memory without holding Spanner locks.
2. **Fast DML Write**: Write the retrieved embeddings to the database in a short, local read-write transaction using batch parameter arrays. The transaction duration is reduced from ~85 seconds to <50 milliseconds, eliminating range split conflicts.
3. **DDL and Test Cleanups**:
   - Removed unused secondary index `NodeEmbeddingLabelIndex` on `embedding_label`.
   - Updated integration and unit tests in `app_test.py` and `utils/embeddings_test.py` to match the two-stage execution model."


gmechali (approved):
Thank you Dan! (pls wait for Xiao's review too)

pipeline/workflow/ingestion-helper/utils/embeddings.py
```diff
@@ -185,25 +197,59 @@ def chunked(iterable, n):
...
             }
             param_types = {
                 "nodes": Array(struct_type),
-                "embedding_label": STRING,
                 "task_type": STRING
             }
 
+            # 1. Run ML.PREDICT as a read-only query outside read-write transaction
+            predictions = []
+            try:
```
shixiao-coder, line 294: have this been tested with 1k+ Nodes to be inserted? I'm a bit concerned about the memory usage by keeping a large number of SVs in memory, maybe consider change this to yield

shixiao-coder: The PR lgtm except this potential memory usage risk

dwnoble: I tested with the full UN dataset which is a few thousand nodes and worked as intended

PR #659, opened 2026-07-17
Temporarily rewrite custom namespaces to dcid: in JSON-LD output

dwnoble:
Temporarily rewrite custom namespaces (like `undata:`) to the `dcid:` namespace in JSON-LD preprocessor output. 

This is a temporary workaround to bypass Spanner schema issues during JSON-LD parsing/normalization until the Java ingestion pipeline supports custom namespaces in the `@context`.

### Example JSON-LD Output Transformation:

**Before (dropped by Java parser due to missing namespace context):**
```json
{
    "@context": {
        "dcid": "https://datacommons.org/browser/"
    },
    "@graph": [
        {
            "@id": "undata:provenance/WHOGlobalHealthObservatory",
            "@type": "dcid:Provenance",
            "dcid:name": "WHO Global Health Observatory",
            "dcid:source": {
                "@id": "undata:source/WorldHealthOrganization"
            }
        }
    ]
}
```

**After (successfully parsed as standard DCID nodes):**
```json
{
    "@context": {
        "dcid": "https://datacommons.org/browser/"
    },
    "@graph": [
        {
            "@id": "dcid:provenance/WHOGlobalHealthObservatory",
            "@type": "dcid:Provenance",
            "dcid:name": "WHO Global Health Observatory",
            "dcid:source": {
                "@id": "dcid:source/WorldHealthOrganization"
            }
        }
    ]
}
```

Includes unit tests verifying the rewriting logic.


gmechali:
Can you make sure to re-pull down from master? It looks like there's some conflict around some of the work I did to make the job fully offload mem

simple/stats/jsonld_stream_db.py
gmechali, line 388: I think I already implemented this? 
Is this just a merge conflict issue. It looks like you may have rebuilt some of what I addressed in the past 2 days?

gmechali (commented):
Can you explain why we're rewriting from custom namespaces to dcid? 
Is this just so that we can tell them to generate their MCF / CSVs using custom namespaces, we dont respect them (for now), but then in a few weeks, when they reingest, we will start respecting the custom namespaces?

dwnoble:
> Can you make sure to re-pull down from master? It looks like there's some conflict around some of the work I did to make the job fully offload mem

Fixed! Can you take another look?

simple/stats/jsonld_stream_db.py
dwnoble: whoops, fixed

dwnoble:
> Can you explain why we're rewriting from custom namespaces to dcid? Is this just so that we can tell them to generate their MCF / CSVs using custom namespaces, we dont respect them (for now), but then in a few weeks, when they reingest, we will start respecting the custom namespaces?

that is exactly right. i added an inline comment that reflects this

simple/stats/jsonld_stream_db.py
```diff
@@ -53,6 +53,21 @@
 _EXPORT_PROCESSES_MAX = 8
 
 
+def _rewrite_custom_ns_to_dcid(val: str) -> str:
+  """Rewrites custom namespaces to 'dcid:' as a temporary workaround until the ingestion loader natively supports custom namespaces.
+
+  This allows users to author schemas with their custom namespaces today, which will be fully preserved when we re-ingest after backend support is added.
+  """
+  if not isinstance(val, str) or not val:
```
gmechali, line 66: What do you think about using a real regex to ensure it's
[a-z] for like min 3 chars max 9 chars (or whatever rule we want)
followed by ":"
and followed by [a-z|_] for the actual property/sv etc.?

I can imagine some bad data like undata:who:SV which then breaks something here? 

We could throw a new error for invalid namespace if it doesn't match the namespace you're expecting?

simple/stats/jsonld_stream_db.py
```diff
@@ -109,20 +124,20 @@ def _write_observation_shard(args):
...
       if provenance in prov_urls and prov_urls[provenance]:
         obs_obj["dcid:provenanceUrl"] = prov_urls[provenance]
     if unit:
-      obs_obj["dcid:unit"] = _uri_ref(unit)
+      obs_obj["dcid:unit"] = _uri_ref(_rewrite_custom_ns_to_dcid(unit))
     if scaling_factor:
       obs_obj["dcid:scalingFactor"] = _parse_numeric(scaling_factor)
     if mmethod:
-      obs_obj["dcid:measurementMethod"] = _uri_ref(mmethod)
+      obs_obj["dcid:measurementMethod"] = _uri_ref(_rewrite_custom_ns_to_dcid(mmethod))
```
gmechali: once we have the namespace in the first funciton, I wonder if we can store that and use it for a simpler string replace everywhere else to aviod the repeated call with the string check etc.

gmechali (approved):
Approving, but I do think we might want to make the namespace identification a bit more robust + add error throwing if it fails

simple/stats/jsonld_stream_db.py
dwnoble: I think we should do validation, but i think the validation should be when we're reading the MCF rather than here. let's leave that to a follow-up PR

simple/stats/jsonld_stream_db.py
dwnoble: added a lookup cache to the _rewrite_custom_ns_to_dcid function

PR #664, opened 2026-07-17
[DCP Ingestion] Ingestion Robustness improvements - Float precision + Obs ID Hash Collisions logging

gmechali:
## Overview

- **Fix Float Precision Loss**: Read CSVs as strings (`dtype=str`) in `VariablePerRowImporter` to preserve exact user numeric formats (`2` vs `2.0`, large ints) and removed obsolete conversion helpers.
- **Observation `@id` Collision Tracking**: Added a low-memory 64-bit int hash tracker in `JsonLdStreamDb` to log warnings and sample `@id`s when duplicate metadata keys produce colliding `@id`s.
- **Code Cleanups**: Replaced tuple unpacking in `_write_observation_shard` with explicit typed parameters.

## Testing
- Verified precision and collision logging using `precision_test` and `collision_test` sample datasets on Cloud Build.
- All 163 unit tests pass.
- Confirmed the JSONLD output in gs://gmechali-june22-dc-artifacts-datcom-website-dev/ingestion/internal/datacommons/jsonld/precision_test_20260718_131029_650454/provenance has preserved the right value types.


simple/stats/jsonld_stream_db.py
```diff
@@ -292,6 +313,48 @@ def __init__(self, output_dir, import_names, nodes) -> None:
     self._processed_imports = set()
     self._node_streaming_started = set()
 
+    # 64-bit Integer Hash Tracker for Observation @ID Collisions.
```
dwnoble, line 316: My jetski says:

```

2. Upgrade to 128-bit hash truncation to prevent false positive warnings on huge datasets
The code truncates the SHA-256 hash to 64 bits (16 hex characters) to save memory:

python
hash_int = int(obs_hash[:16], 16)
Under the Birthday Paradox, a 64-bit space starts to see random hash collisions when tracking larger datasets (for example, there is a $\approx 2.7$% chance of a random collision at 10 million rows, even when the metadata keys are totally different). This would result in false warning logs.

Since Python integers are arbitrary-precision, a 128-bit integer uses only $\approx 4$ additional bytes of memory per entry compared to a 64-bit integer.

Recommendation: Truncate to 128 bits (32 hex characters) instead. This virtually eliminates any chance of a false collision warning up to trillions of records for negligible extra memory:

python
# Truncate SHA-256 to 32 hex chars (128 bits) for absolute collision safety
hash_int = int(obs_hash[:32], 16)
```

But if your jetski disagrees or thinks the issue is minimal im fine with it as is

dwnoble (approved):

simple/stats/jsonld_stream_db.py
gmechali: My Gemini says your Gemini made a classic order of magnitude hallucination. It's 2.7% for 1 billion rows. Or  0.00027% at 10 million rows. 

This just logs a warning, it doesn't throw so will submit as is.

PR #675, opened 2026-07-21
[DCP Ingestion] Fully multi threaded ingestion preprocessor

gmechali:
### Summary of Changes
- **Parallel CSV Ingestion:** Leverages `ProcessPoolExecutor` in `DCP_BRIDGE` mode to parallelize CSV imports across worker processes.
- **Unified JSON-LD Directory Streaming:** Workers share the master timestamped `jsonld_dir`, outputting all shards to a single GCS path for clean wildcard handshakes.
- **Multi-Type Support:** Introduced `_create_importer_for_file` factory to seamlessly handle `OBSERVATIONS`, `EVENTS`, and `ENTITIES` in parallel.
- **Diagnostics & Node State Merging:** Collects collision metrics, sample collision strings, and resolved place/entity nodes from worker processes into parent state.
- **Pickle-Safe Exception Handling:** Updated `EntityResolutionError` for safe `multiprocessing` pickling and diagnostic error reporting.

22M nodes, processed in 15 minutes on Cloud Run Job 8-core (down from 55 minutes), or 5mins 8 seconds on cloud batch 32core.


simple/stats/runner.py
```diff
@@ -80,6 +82,97 @@ class RunMode(StrEnum):
...
+  resolved_entities = {
+      e.entity_dcid: e.entity_type for e in nodes.entities.values()
+  }
+  return (
+      file_rel_path,
+      db.obs_collision_count,
+      dict(db.file_collision_counts),
+      dict(db.file_sample_collisions),
+      resolved_entities,
+  )
```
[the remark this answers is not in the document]
gmechali: Done.

simple/stats/events_importer.py
```diff
@@ -198,9 +198,17 @@ def _write_event_triples(self) -> None:
 
     triples: list[Triple] = []
     for index, row in self.df.iterrows():
-      # If id column is configured, use it as the event dcid else generate based on row index.
-      dcid = row[
-          id_column_name] if id_column_name else f"{self.event_type}_{index}"
+      if self.id_column:
+        if id_column_name and id_column_name in row:
```
gmechali, line 202: FYI this will change in follow up for full column mappings support.

simple/stats/jsonld_stream_db.py
```diff
@@ -51,7 +54,7 @@
 
 # Configuration Constants
 _CHUNK_SIZE = 10000
-_UPLOAD_CONCURRENCY = 32
+_UPLOAD_CONCURRENCY = 4
```
gmechali, line 57: This causes GCS issues if we try to go too fast where we would clobber GCS. Note this is also per worker. If we run on 8 processes each worker can run 4 GCS uploads concurrently. 

With the previousl imit it was too high for GCS.

simple/stats/runner.py
```diff
@@ -80,6 +82,141 @@ class RunMode(StrEnum):
...
+        nodes=nodes,
+    )
+
+  import_type = config.import_type(input_file)
+
+  match import_type:
+
+    case ImportType.OBSERVATIONS:
+      input_file_format = config.format(input_file)
+      if input_file_format == InputFileFormat.VARIABLE_PER_ROW:
```
gmechali, line 116: A bunch of this stuff is already guaranteed bc DCP only, and should be cleaned up further.

simple/stats/runner.py
```diff
@@ -80,6 +85,141 @@ class RunMode(StrEnum):
...
+def _create_importer_for_file(
+    config: Config,
+    input_file: File,
+    process_dir: Dir,
+    db: Db,
+    reporter: FileImportReporter,
+    nodes: Nodes,
+    mode: Optional[RunMode] = None,
+) -> Importer:
+  if input_file.path.endswith(".mcf"):
```
dwnoble: should this be a case insensitive check?

simple/stats/runner.py
```diff
@@ -80,6 +85,141 @@ class RunMode(StrEnum):
...
+        e.entity_dcid: e.entity_type for e in nodes.entities.values()
+    }
+    event_types = dict(nodes.event_types)
+    entity_types = dict(nodes.entity_types)
+    variables = dict(nodes.variables)
+    sources = dict(nodes.sources)
+    provenances = dict(nodes.provenances)
+    groups = dict(nodes.groups)
+    properties = dict(nodes.properties)
+    return (
```
dwnoble: Return a dataclass here instead of the large tuple

simple/stats/runner.py
```diff
@@ -80,6 +85,141 @@ class RunMode(StrEnum):
...
+          reporter=reporter,
+          nodes=nodes,
+      )
+
+    case _:
+      raise ValueError(
+          f"Unsupported import type: {import_type} ({input_file.full_path()})")
+
+
+def _run_single_csv_import_proc(args: tuple):
```
dwnoble: as an alternative to making the arg a tuple, can you give named arguments like
```
    file_rel_path: str,
    input_dir_path: str,
    output_dir_path: str,
    process_dir_path: str,
    import_names: list[str],
    config_json_str: str,
    jsonld_dir_name: str,
```

and you can call the function like

```
                executor.submit(_run_single_csv_import_proc, *arg)
```

simple/stats/jsonld_stream_db.py
```diff
@@ -99,35 +102,32 @@ def _parse_numeric(val):
...
-    track_hash_fn: Optional callback function to track 64-bit @id hashes.
-  """
+def _write_observation_shard(chunk_or_args,
+                             shard_index: Optional[int] = None,
+                             jsonld_dir_path: Optional[str] = None,
+                             ns_map: Optional[dict[str, str]] = None,
+                             prov_urls: Optional[dict[str, str]] = None,
+                             track_hash_fn: Optional[Callable] = None,
+                             file_name: str = ""):
+  if isinstance(chunk_or_args, tuple):
```
dwnoble, line 112: Is the chunk_or_args as-a-tuple path used anywhere in the code? can we remove this branch?

simple/stats/runner.py
```diff
@@ -835,37 +975,85 @@ def _run_all_data_imports(self):
...
-        with concurrent.futures.ThreadPoolExecutor(
-            max_workers=num_mcf_threads) as executor:
-          futures = [
-              executor.submit(self._run_single_mcf_import, file)
-              for file in mcf_files
-          ]
-          for future in concurrent.futures.as_completed(futures):
-            future.result()
+        # TODO(gmechali): Remove thread fallback mock check once unit tests mock dc_client inside process pool.
+        is_mocked = isinstance(
```
dwnoble: is this checking if we're in a test or not? Instead, what do you think about passing into the runner a flag for running running in multiprocessor mode or not (true for live code, false for tests?)

dwnoble (approved):

simple/stats/runner.py
gmechali: done

simple/stats/runner.py
gmechali: done.

simple/stats/jsonld_stream_db.py
gmechali: obsolete.

simple/stats/runner.py
gmechali: ah yeah better. thanks!

PR #676, opened 2026-07-22
Remove obsolete aggregation code from ingestion-helper

SandeepTuniki:


vish-cs (approved):
Shall we also remove BQ env vars from
https://github.com/datacommonsorg/import/blob/master/pipeline/workflow/deploy-services.yaml#L39

SandeepTuniki:
> Shall we also remove BQ env vars from https://github.com/datacommonsorg/import/blob/master/pipeline/workflow/deploy-services.yaml#L39

Done.

PR #679, opened 2026-07-23
Refactor aggregation orchestrator and entrypoint with dataclasses

SandeepTuniki:
A small cleanup PR to manage the growing list of params that the aggregation orchestrator has to deal with. Mostly its about wrapping individual params into dataclasses.

### DIFF
--- a/pipeline/workflow/aggregation-helper/aggregation/orchestrator.py
+++ b/pipeline/workflow/aggregation-helper/aggregation/orchestrator.py
@@ -18,6 +18,7 @@ import glob
 import logging
 import os
 import time
+from collections import deque
 from dataclasses import dataclass, field
 from enum import Enum
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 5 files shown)


pipeline/workflow/aggregation-helper/aggregation/orchestrator.py
```diff
@@ -229,50 +231,59 @@ def run(self, active_imports: Optional[List[str]] = None, dry_run: bool = True,
...
+                except Exception as e:
+                    logging.error(f"Global calculation step '{step_type}' failed: {e}")
+                    return ImportExecutionResult(
                         import_name="GLOBAL",
-                        success=True,
-                        stages_executed=[]
+                        success=False,
+                        stages_executed=[],
+                        error_message=str(e)
                     )
```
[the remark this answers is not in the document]
SandeepTuniki: Applied the suggested fix.

pipeline/workflow/aggregation-helper/aggregation/orchestrator.py
```diff
@@ -79,60 +80,61 @@ class CalculationType(str, Enum):
     EMBEDDING_GENERATION = "EMBEDDING_GENERATION"
 
 
+GLOBAL_CALCULATION_TYPES = {
+    CalculationType.EMBEDDING_GENERATION,
+}
+
+
+@dataclass
+class OrchestratorConfig:
```
vish-cs, line 89: Do we need separate classes OrchestratorConfig and CloudRunJobConfig? They seem fairly similar so can we reuse?

pipeline/workflow/aggregation-helper/main.py
```diff
@@ -19,8 +19,84 @@
...
 import os
 import sys
+from dataclasses import dataclass
+from typing import List, Optional
+
+from aggregation import AggregationOrchestrator, OrchestratorConfig
+
+
+@dataclass
+class CloudRunJobConfig:
```
vish-cs: Can we rename to AggregationJobConfig?

pipeline/workflow/aggregation-helper/main.py
vish-cs, line 1: It would be helpful to have a quick README explaining the job config params. Useful for agents. Can be a follow up.

pipeline/workflow/aggregation-helper/aggregation/orchestrator.py
SandeepTuniki: Good point. I see 8 out of 12 fields are same across the two classes. I removed `CloudRunJobConfig` and instead wrote a small helper function that directly creates an `OrchestratorConfig` object.

pipeline/workflow/aggregation-helper/main.py
SandeepTuniki: Removed this class as per the previous comment's observation of its similarity with `OrchestratorConfig`.

pipeline/workflow/aggregation-helper/main.py
SandeepTuniki: I'll do this in a follow-up PR. I intend to add a few more helpful default params to make testing and triggering easier. I'll do all of them at once in a different PR.

vish-cs (approved):

PR #686, opened 2026-07-24
[DCP Ingestion] Namespace support in preprocessor

gmechali:
### Standardize on `provenance_id` across graph nodes and enable per-provenance GCS handshakes

- **Per-Provenance GCS Handshakes**: Update `dcpbridge` mode to generate a per-provenance `{workflow_id}.json` handshake manifest where `importList` specifies each individual provenance folder (`provenance/<name>`) and its shard glob (`**/*.jsonld`).
- **Standardize Data Model**: Replace `provenance_dir(s)` with `provenance_id(s)` across `Entity`, `Property`, `Provenance`, and `Source` to match `StatVar`, `EventType`, and `EntityType`.
- **Remove Upfront Normalization**: Eliminate boilerplate `config.import_name()` and `provenance_dir` piping in importers (`Observations`, `VariablePerRow`, `Events`, `Mcf`) and runner IPC serialization.
- **Centralize Shard Bucketing**: Consolidate folder normalization (`strip_namespace(prov_id)`) to a single location at the JSON-LD export boundary in `Nodes.triples_by_provenance_dir()`.


simple/stats/events_importer.py
```diff
@@ -237,14 +240,27 @@ def _resolve_entities(self) -> None:
...
+    except Exception:
+      prov_dir = ""
+    for dcid in column.tolist():
+      clean_dcid = strip_namespace(dcid)
+      if self.nodes.has_entity(clean_dcid) and prov_dir:
+        self.nodes.entities[clean_dcid].provenance_dirs.add(prov_dir)
+      elif prov_dir:
+        self.nodes.entity_with_type(clean_dcid,
+                                    "Thing",
+                                    provenance_dir=prov_dir)
```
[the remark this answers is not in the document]
gmechali: this is fine, we dont support it

simple/stats/util.py
```diff
@@ -55,6 +55,33 @@ def base64_decode_and_gunzip_json(encoded_data: str) -> dict:
   return json.loads(base64_decode_and_gunzip(encoded_data))
 
 
+_STANDARD_PREFIXES = {"http", "https", "dcid", "schema", "dcs"}
```
clincoln8, line 58: ensure these are documented for partners

simple/stats/util.py
```diff
@@ -55,6 +55,33 @@ def base64_decode_and_gunzip_json(encoded_data: str) -> dict:
...
+
+def is_custom_namespace_prefix(prefix: str) -> bool:
+  """Returns True if the prefix represents a custom instance namespace (not standard 'dcid', 'http', 'schema')."""
+  if not prefix:
+    return False
+  p_lower = prefix.lower()
+  return p_lower not in _STANDARD_PREFIXES
+
+
 def is_uri_or_namespace(val: str) -> bool:
```
clincoln8, line 85: Consider consolidating namespace validation across helper functions. Currently, there is divergence between has_namespace_prefix, get_namespace_prefix_and_suffix, and is_uri_or_namespace:

* has_namespace_prefix allows underscores and hyphens in prefixes via _NAMESPACE_PREFIX_RE (^[a-zA-Z0-9_\-]+:).
* is_uri_or_namespace checks prefix.isalnum(), which returns False for valid prefixes containing underscores or hyphens (e.g., custom_ns: or my-org:). It also checks " " not in val, which rejects strings that have spaces in the value portion (e.g., custom:hello world).
* get_namespace_prefix_and_suffix checks " " not in val.split(":", 1)[0] without validating prefix characters, allowing non-alphanumeric prefixes (e.g., foo/bar:baz).

Could we refactor both is_uri_or_namespace and get_namespace_prefix_and_suffix to use _NAMESPACE_PREFIX_RE (or a shared helper function) so that namespace detection, character rules, and splitting are 100% consistent across the module?

simple/stats/data.py
clincoln8, line 589: Do we need to update strip_namespace to check for full URLs (e.g., if v.startswith(("http://", "https://")): or using get_namespace_prefix_and_suffix from util.py) before stripping the namespace prefix? Or are we guranteed to be using namespaces, not urls, by the time we get to this step?

simple/util/filesystem.py
```diff
@@ -204,7 +204,10 @@ def __init__(self, store: "Store", path: str, create_if_missing: bool):
         # Make parent dir if needed.
         parent_dir_path = fspath.dirname(path)
         if not self.fs().isdir(parent_dir_path):
-          self.fs().makedirs(parent_dir_path)
+          try:
+            self.fs().makedirs(parent_dir_path)
```
clincoln8: Swallowing all exceptions in File.__init__ avoids concurrent worker race conditions, but masks real errors (e.g., permission denials or quota limits), causing confusing downstream failures when touch() runs.

Could we use self.fs().makedirs(parent_dir_path, recreate=True), or re-check if not self.fs().isdir(parent_dir_path): raise in the except block to ensure genuine errors aren't swallowed?

simple/stats/nodes.py
```diff
@@ -420,6 +446,61 @@ def triples(self, triples_file: File | None = None) -> list[Triple]:
 
     return triples
 
+  @thread_safe
+  def triples_by_provenance_dir(self) -> dict[str, list[Triple]]:
```
clincoln8, line 450: uggest DRYing up node partitioning loops and guarding against duplicate triples in triples_by_provenance_dir(). In triples_by_provenance_dir(), the logic for partitioning node triples across self.variables, self.event_types, self.entity_types, self.properties, and self.entities is identical across 5 separate loops. This could be simplified by iterating over the collections in a single helper loop.

Additionally, because result is a defaultdict(list) and appends via .extend(), if a node has duplicate entries in provenance_ids (or if two different namespaced IDs strip down to the exact same folder name), duplicate triples will be emitted into the same shard. Could we deduplicate p_ids (e.g., using set(p_ids)) or deduplicate the exported triples before writing?

simple/stats/validation.py
```diff
@@ -102,15 +104,15 @@ def _collect_defined_nodes(self) -> tuple[set[str], dict[str, str]]:
 
     for triple in all_triples:
       sub = self._clean_dcid(triple.subject_id)
-      pred = triple.predicate
+      pred = strip_namespace(triple.predicate)
```
clincoln8, line 107: Minor cleanup and naming clarification for _clean_dcid(). In _clean_dcid(), checking if val.startswith(("http://", "https://")): is immediately followed by if has_namespace_prefix(val):. Since has_namespace_prefix matches any prefix followed by a colon via regex (^[a-zA-Z0-9_\-]+:), it already returns True for http: and https: URIs. The explicit startswith check could be removed for brevity.

Additionally, because _clean_dcid actually adds the default dcid: prefix when missing (acting as the exact inverse of strip_namespace), the name _clean_dcid can be slightly confusing to future readers who might assume "cleaning" means stripping. Could we rename it to something more descriptive like ensure_namespace_prefix() or normalize_dcid()?

simple/stats/jsonld_stream_db.py
```diff
@@ -465,38 +462,43 @@ def _write_triples_to_disk(self, triples: list[Triple], import_name: str):
...
   def commit_and_close(self):
-    # Add global triples to every processed import's triples
     global_triples = self._triples.pop("_global", [])
     if not self._processed_imports:
       self._processed_imports.add(self.import_name)
 
-    # Write global triples as node shards to the local temp directory for each import
-    for import_name in self._processed_imports:
-      import_temp_dir = os.path.join(self.temp_local_dir, import_name)
+    # Write any remaining untagged triples once to the primary import directory (no duplication across provenances)
```
clincoln8, line 491: This won't cause any bugs if they refactor their config structure and only import a single provenance?

simple/stats/observations_importer.py
```diff
@@ -147,6 +149,12 @@ def _add_entity_nodes(self) -> None:
         dcid for dcid in entity_dcids if not self.nodes.has_entity(dcid)
     ]
 
+    prov_id = getattr(self, "provenance", "")
```
clincoln8, line 152: is this intentional despite the class not having a provenance attribute?

clincoln8 (approved):
Thank you!!!!!!

Nonblocking questions / comments. Can likely be addressed in followups if needed.

simple/stats/util.py
gmechali: Fair - we ll have to do some pretty significant work on namespace to make the real namespace stuff work. So I will leave that up for whoever does this in the coming weeks! 

We wil ldefinitely call out the need for a deep review of how far we got on namespace and open issues.

simple/stats/jsonld_stream_db.py
gmechali: This was a bug i introduced last week, when we extracted all the nodes, like SVs into this global var, we then wrote them all in the josnld output under each provenance. Essentially duplicating nodes across provenances they had nothing to do with. Im pretty sure this is safe now.

simple/stats/observations_importer.py
gmechali: half, it kinda copied it from other importers that did this. im not gonna worry about observations_importer since it's essentially dead code. Will properly tear it up very very shortly. :)

simple/util/filesystem.py
gmechali: done, added the recreate=true

simple/stats/data.py
gmechali: we strip the namespaces today but we dont want to strip it in the future. So leaving this up to whoever is wokring on nampespace support in the coming weeks. 

Will need to rethink when to strip / parse the namespace, and what's the full supported set.

simple/stats/nodes.py
gmechali: yeah thanks for catching, fixed.

simple/stats/validation.py
gmechali: I think this is also gonna need to fall in the clean namespace work coming up. There's alrady a weird combination of unorganized helpers. I think this isnt going to affect how it works but it does need to be cleaned up.

gmechali:
Thanks Christie! Fixed like half the comments, the half regarding namespaces i think will need to be hanlded in the namespace work coming up, but agree fully on the cleanup needed :)

PR #693, opened 2026-07-28
Add bundle-local deduplication for time series and fix metric counter…

vish-cs:
- Implement bundle-local deduplication in ExtractTimeSeriesFn using @StartBundle and a bundle-scoped HashSet to reduce 95%+ of duplicate time series keys before global Dataflow shuffle.
- Refactor anonymous inner DoFn classes in GraphReader (CountTimeSeriesFn, ExtractSeriesFromOptimizedFn, ExtractObservationsFromOptimizedFn, GraphToNodesFn, GraphToEdgesFn) into static named DoFn classes with private final Counter fields to ensure reliable Beam metric reporting in Dataflow.
- Add unit test testExtractTimeSeriesFn_BundleDeduplication in GraphReaderTest.

### DIFF
--- a/pipeline/util/src/main/java/org/datacommons/ingestion/util/GraphReader.java
+++ b/pipeline/util/src/main/java/org/datacommons/ingestion/util/GraphReader.java
@@ -6,8 +6,10 @@ import com.google.common.base.Joiner;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.beam.sdk.metrics.Counter;
 import org.apache.beam.sdk.transforms.Distinct;
@@ -197,12 +199,19 @@ public class GraphReader implements Serializable {
 
   /** Extracts unique TimeSeries (metadata series keys) from observation nodes. */
   public static class ExtractTimeSeriesFn extends DoFn<McfGraph, TimeSeries> {
+    private static final int MAX_BUNDLE_CACHE_SIZE = 100_000;
     private final String importName;
     private final boolean isBaseDc;
+    private transient Set<String> bundleSeenKeys;
 
     public ExtractTimeSeriesFn(String importName, boolean isBaseDc) {
       this.importName = importName;
       this.isBaseDc = isBaseDc;
+    }
+
+    @StartBundle
+    public void startBundle() {
+      bundleSeenKeys = new HashSet<>();
     }
 
     @ProcessElement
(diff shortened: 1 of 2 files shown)


pipeline/util/src/main/java/org/datacommons/ingestion/util/GraphReader.java
```diff
@@ -194,20 +187,28 @@ public static PCollection<Observation> extractObservations(
   public static class ExtractTimeSeriesFn extends DoFn<McfGraph, TimeSeries> {
     private final String importName;
     private final boolean isBaseDc;
+    private transient Set<String> bundleSeenKeys;
```
rohitkumarbhagat, line 205: Bound the bundle-local deduplication cache

 `bundleSeenKeys` retains one string per unique series for the entire runner-defined bundle. A large, low-duplication import can therefore consume substantial worker heap and potentially OOM. `@FinishBundle` only releases memory afterward; it does not bound peak usage. Also, the existing `Distinct.withRepresentativeValueFn` expands to `Combine.perKey`, which already permits runner-side precombining before shuffle. Either remove this extra cache or cap it and periodically clear it—the downstream global `Distinct` preserves correctness.

pipeline/util/src/main/java/org/datacommons/ingestion/util/GraphReader.java
```diff
@@ -515,21 +481,108 @@ public static PCollection<KV<String, Mutation>> graphToEdges(
...
-                        e -> {
-                          receiver.output(e);
-                        });
-                edgeCounter.inc(mutations.size());
-              }
-            }));
+        ParDo.of(new GraphToEdgesFn(provenance, spannerClient, edgeCounter)));
+  }
+
+  public static class CountTimeSeriesFn extends DoFn<TimeSeries, TimeSeries> {
```
rohitkumarbhagat: The named `DoFn` refactor does not fix counter serialization.

The named `DoFn` classes improve readability, but they should not change metric serialization. The previous anonymous classes were created inside static methods and captured the same serializable fields—such as `Counter` and `SpannerClient`—that the new classes store explicitly. Therefore, this refactor alone is unlikely to fix the missing Dataflow metrics.

rohitkumarbhagat (approved):

pipeline/util/src/main/java/org/datacommons/ingestion/util/GraphReader.java
vish-cs: That's right. Reverted

pipeline/util/src/main/java/org/datacommons/ingestion/util/GraphReader.java
vish-cs: Using a bounded bundle of size 100K

pipeline/util/src/main/java/org/datacommons/ingestion/util/GraphReader.java
```diff
@@ -194,20 +187,28 @@ public static PCollection<Observation> extractObservations(
...
 
     public ExtractTimeSeriesFn(String importName, boolean isBaseDc) {
       this.importName = importName;
       this.isBaseDc = isBaseDc;
     }
 
+    @StartBundle
+    public void startBundle() {
+      bundleSeenKeys = new HashSet<>();
+    }
```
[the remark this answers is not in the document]
vish-cs: Done

PR #700, opened 2026-07-31
Disable all non-core aggregations in aggregation-helper configs

SandeepTuniki:
Everything except the ones in `configs/common.yaml` are marked as disabled for now. We'll gradually enable them and bring them into auto-refresh.


vish-cs (approved):
So it seems we have to enable/disable these per-import

PR #711, opened 2026-08-04
Update workflow to support import list overrides

vish-cs:
Currently, ingestion workflow takes a list of import names as an argument. It then fetches the latest version information from the spanner ImportStatus table. This PR allows passing an optional latest version path as an input argument for easy backfilling. For this, we make the import list argument support a JSON body (importName and latestVersion). This input path eliminates the need for a spanner look up via an override. As part of this, code is cleaned up to merge graphPath and latestVersion params into a single latestVersion parameter.

### DIFF
--- a/pipeline/workflow/ingestion-helper/clients/spanner.py
+++ b/pipeline/workflow/ingestion-helper/clients/spanner.py
@@ -228,49 +228,74 @@ class SpannerClient:
             logging.error(f'Error releasing lock for {workflow_id}: {e}')
             raise
 
-    def get_import_info(self, import_list: list) -> list:
+    def get_import_info(self, import_list: list | None) -> list:
         """Get the details of imports to ingest.
 
-        If import_list is empty, return info for ready imports (STAGING).
-        If import_list is not empty, return info for the imports in the list that are in 'STAGING' status.
+        If import_list is empty, return info for ready imports (STAGING) from Spanner DB.
+        Otherwise, import_list is a list of dicts (or ImportItem objects) containing 'importName'
+        and optional 'latestVersion' (which contains the full GCS graph path).
 
         Args:
-            import_list: A list of import names to fetch details for.
+            import_list: A list of dicts or objects containing 'importName' and optional 'latestVersion'.
 
         Returns:
-            A list of dictionaries, where each dictionary contains 'importName', 'latestVersion', and 'graphPath'.
+            A list of dictionaries, where each dictionary contains 'importName' and 'latestVersion' (full GCS graph path).
         """
-        pending_imports = []
-        logging.info(f"Fetching imports from import list {import_list}.")
+        if not import_list:
+            import_list = []
 
+        pending_imports = []
+        imports_to_fetch = []
+
+        for item in import_list:
+            if isinstance(item, dict):
+                import_name = item.get("importName")
+                provided_version = item.get("latestVersion")
+            elif hasattr(item, "importName"):
+                import_name = getattr(item, "importName", None)
+                provided_version = getattr(item, "latestVersion", None)
+            else:
+                continue
+
+            if not import_name:
+                continue
+
+            if provided_version:
+                pending_imports.append({
+                    "importName": import_name,
+                    "latestVersion": provided_version,
+                })
+            else:
+                imports_to_fetch.append(import_name)
+
+        logging.info(f"Fetching import details. Provided directly: {len(pending_imports)}, Needing DB fetch: {len(imports_to_fetch)}, Empty list: {not import_list}")
+
+        sql = None
         params = {}
         param_types = {}
-        if import_list:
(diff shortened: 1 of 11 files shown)


pipeline/workflow/ingestion-helper/clients/spanner.py
vish-cs, line 1: /gemini review

pipeline/workflow/aggregation-helper/main.py
```diff
@@ -34,8 +34,18 @@ def parse_import_list(import_list_str: Optional[str]) -> List[str]:
...
-    return parsed
+    
+    res = []
+    for item in parsed:
+        if isinstance(item, dict) and "importName" in item:
+            res.append(item["importName"])
+        elif isinstance(item, str):
+            res.append(item)
+        else:
+            res.append(str(item))
```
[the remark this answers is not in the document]
vish-cs: This change is reverted out of this PR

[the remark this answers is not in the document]
vish-cs: This change is reverted out of this PR

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -228,49 +228,97 @@ def _release(transaction: Transaction) -> None:
...
+                            "importName": name,
+                            "latestVersion": latest_ver,
+                            "graphPath": graph_p,
+                        })
+            except Exception as e:
+                logging.error(f'Error getting import list: {e}')
+                raise
+
+        logging.info(f"Found {len(pending_imports)} import jobs.")
+        return pending_imports
```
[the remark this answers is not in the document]
vish-cs: This is outdated. GraphPath and LatestVersion are combined into a single param now.

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -228,49 +228,81 @@ def _release(transaction: Transaction) -> None:
...
+                with self.database.snapshot() as snapshot:
+                    results = snapshot.execute_sql(sql, params=params, param_types=param_types)
+                    for row in results:
+                        pending_imports.append({
+                            "importName": row[0],
+                            "latestVersion": f"{row[1].rstrip('/')}/{row[2].lstrip('/')}",
+                        })
+            except Exception as e:
+                logging.error(f'Error getting import list: {e}')
+                raise
```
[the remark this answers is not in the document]
vish-cs: cleaned up the code to use common query logic

pipeline/workflow/ingestion-helper/routes/imports.py
```diff
@@ -40,7 +40,7 @@ class ImportItem(BaseModel):
 
 
 class ImportInfoRequest(BaseModel):
-    importList: Optional[List[str]] = Field(default_factory=list)
+    importList: Optional[List[ImportItem]] = Field(default_factory=list)
```
rohitkumarbhagat, line 43: This breaks the existing runIngestion path. ImportInfoRequest now requires each importList entry to be an ImportItem, but pipeline/workflow/import-automation-workflow.yaml still constructs importList as a list of strings:

  importList:
    - ${text.split(importName, ":")[1]}

pipeline/workflow/ingestion-helper/clients/spanner.py
```diff
@@ -228,49 +228,74 @@ def _release(transaction: Transaction) -> None:
...
-            logging.error(f'Error getting import list: {e}')
-            raise
+        if sql:
+            try:
+                with self.database.snapshot() as snapshot:
+                    results = snapshot.execute_sql(sql, params=params, param_types=param_types)
+                    for row in results:
+                        pending_imports.append({
+                            "importName": row[0],
+                            "latestVersion": f"{row[1].rstrip('/')}/{row[2].lstrip('/')}",
```
rohitkumarbhagat, line 291: This changes latestVersion from a version identifier to the full graph path, but the same value is later written unchanged to ImportVersionHistory.Version by update_import_version_history.

  Rollback treats Version as a version name and joins it to the current import directory. For example, storing gs://bucket/import/v2/*/*.mcf produces a malformed rollback path such as gs://bucket/import/gs://bucket/import/v2/*/*.mcf.

  Please preserve the version identifier separately for ImportVersionHistory, or normalize the value before writing the history row.

rohitkumarbhagat (approved):

pipeline/workflow/ingestion-helper/routes/imports.py
vish-cs: Good catch. Updated

pipeline/workflow/ingestion-helper/clients/spanner.py
vish-cs: This actually is a good side effect. We currently have a pending TODO to store full version path in the history table so we can use it for rollbacks. This automatically takes care of it. I have updated rollback handling logic for the same.

